PR #9, opened 2014-11-01
Exaile no longer crashes on yosemite or mavericks

virtuald:
Suspect I need a different fix on Mavericks, can someone test this?

Don't merge until I do some testing to see if this fixes https://github.com/exaile-dev/exaile/issues/5


virtuald:
FYI: this error came back in OSX 10.11 for 3.4.x. I suspect it will not be an issue in 4.x, but haven't tried...

PR #25, opened 2015-01-11
Log compilation detection only in debug mode

genodeftest:
I think it is not necessary to log compilation detection in normal mode since this is just informative for debugging purposes.

### DIFF
--- a/xl/collection.py
+++ b/xl/collection.py
@@ -695,7 +695,7 @@ class Library(object):
                 artist not in ccheck[basedir][album]:
             if not (basedir, album) in compilations:
                 compilations.append((basedir, album))
-                logger.info("Compilation %(album)r detected in %(dir)r" %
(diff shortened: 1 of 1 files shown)


virtuald:
What does it harm though? Most people don't actually see the log output anyways? I'm not totally opposed to this, but I don't see a reason to do it.

genodeftest:
When scanning/rescanning >10k music files it generates a lot of output. Since all that output goes to syslog it *spams* syslog. Since I want to keep syslog readable, I want every non-warning or non-error message out.
It's no critical thing so if you don't like it that's no problem for me.

virtuald:
Heh, it goes to syslog? Or is that just how your system has it setup.

genodeftest:
That's how my system has it set up by default (Fedora 20, 21). Every output of every application by default goes to syslog.

PR #28, opened 2015-01-11
Use unambiguous kiB instead of kB

genodeftest:
See https://en.wikipedia.org/wiki/Kibibyte for details.
Backdraw: the kiB unit is less known by users.

### DIFF
--- a/xlgui/panel/files.py
+++ b/xlgui/panel/files.py
@@ -398,13 +398,13 @@ class FilesPanel(panel.Panel):
             for sortname, name, f in subdirs:
                 model.append((f, self.directory, name, ''))
             for sortname, name, f in subfiles:
-                size = f.query_info('standard::size').get_size() // 1000
+                size = f.query_info('standard::size').get_size() // 1024
                 
                 # locale.format_string does not support unicode objects
                 # correctly, so we call it with an str and convert the 
                 # locale-dependent output to unicode.
                 size = locale.format_string('%d', size, True)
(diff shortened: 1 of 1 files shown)


virtuald:
@sjohannes you're familiar with gio, do you know what size is returned by query_info? I'm confused why it would have a comment of 1000 next to it if it were really 1024.

A cursory test on my computer shows that it does indeed return the actual file size, and doesn't mess with the byte count.

sjohannes:
size = f.query_info('standard::size').get_size() // 1000

get_size() returns file size in bytes. I divided it by 1000 (arbitrarily rounding down) to get the size in kilobytes. So the current unit (kB) is correct. The merge request is also correct if we want to divide by 1024 and use the binary prefix (except it needs to be written KiB instead of kiB).

So the question is which one to use.

* OS X Finder reportedly uses the decimal unit (kB, dividing by 1000) for file sizes.
* Ubuntu [requires](https://wiki.ubuntu.com/UnitsPolicy) at least the decimal unit (kB, 1000) to be used by default for file sizes.
* Windows Explorer writes KB and divides by 1024, which is all sorts of wrong.
* I think Nautilus uses the binary unit (KiB, 1024), but I haven't checked.

I opted to follow the Ubuntu policy because I believe it makes the most sense, considering that harddisk sizes are usually specified with the decimal unit.

virtuald:
Oh wow. I saw the // and read it as a comment. I'm not getting enough sleep.

I agree with @sjohannes , will close this.

PR #29, opened 2015-01-11
Update german translation

genodeftest:


virtuald:
Thanks for the contribution! In the future, if you could submit translations via https://hosted.weblate.org/projects/exaile/master/, that would be preferable.

I don't have time at the moment to review all your issues/patches, but I'll try to get to it tonight.

PR #33, opened 2015-01-11
Ignore git error message

genodeftest:
Currently if no git or git repository is found, an error message is printed to stderr of exaile. This is unnecessary and will be fixed by this commit.

### DIFF
--- a/xl/version.py
+++ b/xl/version.py
@@ -41,7 +41,7 @@ def get_current_revision(directory):
     try:
         return subprocess.check_output([
             'git', 'rev-parse', '--short=7', 'HEAD'
-        ]).strip()
+        ], stderr=-1).strip()
     except subprocess.CalledProcessError:
         return None
 


xl/version.py
```diff
@@ -41,7 +41,7 @@ def get_current_revision(directory):
     try:
         return subprocess.check_output([
             'git', 'rev-parse', '--short=7', 'HEAD'
-        ]).strip()
+        ], stderr=-1).strip()
```
virtuald: I'm curious, the documentation doesn't state what happens when you set stderr to an invalid file descriptor...

virtuald: Ah, this is subprocess.PIPE. The python documentation states "Do not use stderr=PIPE with this function as that can deadlock based on the child process error volume". And while the output *should* be small, I'd rather not risk it.

I don't mind suppressing the output, but there has to be a better way.

sjohannes: Python 3 has `subprocess.DEVNULL`, but it's not in 2.x. I think`file(os.devnull, 'w')` works.

genodeftest:
~~~
        import os
        with open(os.devnull, 'w') as devnull:
            return subprocess.check_output([
                'git', 'rev-parse', '--short=7', 'HEAD'
            ], stderr=devnull).strip()
~~~
works fine for me too.
I choose stderr=-1 since this is used in plugins/moodpar/__init__.py:536 and plugins/streamripper/__init__.py:148 but the version with devnull is more readable.

virtuald:
I added the devnull fix that you suggested in e1fb177, thanks.

Issue #48, opened 2015-02-20
Apply group tag to more that one song at once 


flberger:
*Migrated from [https://bugs.launchpad.net/exaile/+bug/1341448](https://bugs.launchpad.net/exaile/+bug/1341448)*

In exaile 3.4.0-dev+bzr4524 and the group tagger plugin 0.3, when I select a couple of songs in a playlist and mark a group tag in the GroupTagger panel, only the very first of the selected songs is bein labelled.

This makes it very cumbersome to apply a group tag to more that a handful songs at once.

I propose the following feature:

- When selecting several songs and checking the group tag, a dialog appears "Are you sure you want to apply tag X to more that one song?"

- Confirming "OK", the tag is being applied to all selected songs.

- The dialog could also feature a "Remember selection and don't ask me again" checkbox.

Thanks for considering.

virtuald:
Fixed, albeit slightly differently.

PR #49, opened 2015-02-20
Allowing group tags to be applied to more than one selected song

flberger:
Proposed fix for  #48 "Apply group tag to more that one song at once".

Has been working nicely for me so far.


virtuald:
Clearly this fell off my radar -- my apologies for that. 

My objection to your patch (I haven't ran it yet) is that you're making changes to the currently selected track(s), which means that the track that is shown in the title isn't actually the one that's being edited.

.. though, it seems like if one did select new track(s), then the title would be updated. Perhaps though you could change the selection mechanism to store tracks in self, instead of self.track, and display to the user [multiple tracks] instead of the single title?

flberger:
Which title display do you mean? On 3.4.3+dc10b58, no title is being displayed upon selection, the status simply reads "Not Playing".
exaile-title-fs8

virtuald:
On the grouptagger panel.

virtuald:
As I said previously, I think the mechanism is a bit confusing and isn't particularly clear on *what* you're actually editing at any given time. I've implemented the functionality as two menu items instead, to 'add tags to all' or 'remove tags from all'. The functionality is available in the GI branch, and will be released with Exaile 4.0.

PR #56, opened 2015-03-08
Set the locale path dependent on where we were installed

### LINKED ISSUE #54: Doesnt show the translation / exaile in english
Somehow exaile doest take the translated .po files.
I tried the Windows installation (german) and the text in exaile is still in english.
If i look in the po folder, the german translation looks ok.
Same i found on linux.

virtuald:
@sjohannes or @mathbr can you sanity check this for me? Fixes #54.

Before applying this fix, if you ran exaile on linux via `sudo make install` and `LANG="de_DE.utf8" exaile`, the locale files wouldn't get loaded properly. However, after applying this fix it seems to work for me.

### DIFF
--- a/xl/nls.py
+++ b/xl/nls.py
@@ -31,7 +31,7 @@
 
 import sys
 import locale
-import os
+import os.path
 
 try:
     # Set to user default, gracefully fallback on C otherwise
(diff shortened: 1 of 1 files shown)


sjohannes:
I can confirm the fix. (How) did the current code ever work? It seems that it wasn't running bindtextdomain at all when Exaile is installed.

virtuald:
I suspect it didn't.

PR #68, opened 2015-03-29
Makes the moodbar plugin work on the Preview Device

bitterjug:
I like the info the moodbar gives me but I find I want it more from the preview device than the main player (because I'm previewing stuff live to decide what to play). This PR refactors the moodbar somewhat to abstract out the dependencies on the player and its gui, and then applies another instance to the preview device, when that is activated. To make this work  I have added a couple of events to the preview device, and made it keep a reference to its progress bar which the moodbar plugin can see. 
I think I'm at a point where I'm ready to invite you to review and comment. 
* I've also added a couple of logger.debugs which you might prefer to have removed?
* Should I also change the minor versions of the two plugins to reflect the changes?
* I've had to play with the possible enable/disable sequences of the two plugins to make sure they work and don't break. I don't understand how the gui stuff actually works, so some of this was trial-and-error. You might know better. 
* Any other suggestions?
Thanks -- Mark

### DIFF
--- a/plugins/moodbar/__init__.py
+++ b/plugins/moodbar/__init__.py
@@ -588,7 +588,7 @@ def _enable_preview_moodbar(event, preview_plugin, nothing):
 
 
 def _disable_preview_moodbar(event, preview_plugin, nothing):
-    global ExaileModbar
+    global PreviewMoodbar
     logger.info("Disabling preview moodbar")
     PreviewMoodbar.changeModToBar()
     PreviewMoodbar.remove_callbacks()
--- a/plugins/moodbar/__init__.py
+++ b/plugins/moodbar/__init__.py
@@ -554,7 +554,7 @@ class ExModbar(object):
 
 def _enable_main_moodbar(exaile):
     global ExaileModbar
-    logger.info("Enabling main moodbar")
+    logger.debug("Enabling main moodbar")
     ExaileModbar = ExModbar(
         player=player.PLAYER,
         progress_bar=exaile.gui.main.progress_bar
@@ -567,7 +567,7 @@ def _enable_main_moodbar(exaile):
 
 def _disable_main_moodbar():
     global ExaileModbar
-    logger.info("Disabling main moodbar")
+    logger.debug("Disabling main moodbar")
     ExaileModbar.changeModToBar()
     ExaileModbar.remove_callbacks()
     ExaileModbar.destroy()
@@ -576,7 +576,7 @@ def _disable_main_moodbar():
 
 def _enable_preview_moodbar(event, preview_plugin, nothing):
     global PreviewMoodbar
-    logger.info("Enabling preview moodbar")
+    logger.debug("Enabling preview moodbar")
     PreviewMoodbar = ExModbar(
         player=preview_plugin.player,
         progress_bar=preview_plugin.progress_bar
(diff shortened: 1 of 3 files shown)


virtuald:
Thanks for the PR, I'll try to look at it this weekend unless something comes up.

virtuald:
Sorry for the delay -- but this is still on my list of things to do!

plugins/moodbar/__init__.py
```diff
@@ -543,20 +608,41 @@ def enable(exaile):
...
-    ExaileModbar.setupUi()
-    event.add_callback(ExaileModbar.play_start, 'playback_track_start', player.PLAYER)
-    event.add_callback(ExaileModbar.play_end, 'playback_player_end', player.PLAYER)
+    _enable_main_moodbar(exaile)
+
+    event.add_callback(_enable_preview_moodbar, 'preview_device_enabled')
+    event.add_callback(_disable_preview_moodbar, 'preview_device_disabling')
+
+    try:
+        import previewdevice
```
virtuald: I don't like this import. I think there's a way to get the plugin object differently... well, I guess not.

I think the better option would be to add a call to the PluginManager like `get_enabled_plugin()`. All currently enabled plugins are in the enabled_plugins dictionary. The objects in there are the plugin modules.

plugins/previewdevice/__init__.py
```diff
@@ -42,80 +39,95 @@
...
+
         # preserve state
         if settings.get_option('plugin/previewdevice/shown', True):
             self._init_gui_hooks()
-    
-    
+
+        # We're ready; teall anyone who might be interested
+        logger.debug('Preview Device Enabled')
+        # event.log_event('preview_device_enabled', self, None)
```
virtuald: Meant to keep this commented out?

virtuald:
I would have rather you put the lint changes in a separate PR or something... it makes it annoying to figure out which pieces you actually changed.

plugins/moodbar/__init__.py
bitterjug: The `enabled_plugins` dictionary (thanks for pointing that out) looks sufficient; how about:
``` python
def _get_preview_plugin_if_active(exaile):
    previewdevice = exaile.plugins.enabled_plugins.get('previewdevice', none)
    return getattr(previewdevice, 'preview_plugin', none)

def _enable(eventname, exaile, nothing):
    _enable_main_moodbar(exaile)

    event.add_callback(_enable_preview_moodbar, 'preview_device_enabled')
    event.add_callback(_disable_preview_moodbar, 'preview_device_disabling')

    preview_plugin = _get_preview_plugin_if_active(exaile)
    if getattr(preview_plugin, 'hooked', false):
        _enable_preview_moodbar('', preview_plugin, none)

```

bitterjug:
I had hoped it would suffice to put the lint changes into separate commits; I can see how it must be annoying to view the whole PR as a single diff. Since Im going to change how the preview plugin instance gets into moodbar I'll make a separate PR for the layout changes. In fact I've stuck moodbar.__init__.py through [yapf](https://github.com/google/yapf) to sort out its erratic indentation, and rebased the other changes on top of that (and reading those diffs sucked!).

plugins/moodbar/__init__.py
virtuald: I'd rather add a function to the plugin manager instead of accessing the dictionary directly, but this definitely seems better.

bitterjug: Been thinking about this.

virtuald: Well, if we have other plugins accessing the dictionary, then it's probably ok to keep it as is. The lines between public and private API is a bit fuzzy for Exaile at the moment... just submit the PR as is, and I rescind my request.

PR #78, opened 2015-05-03
Make moodbar plugin work with preview device

bitterjug:
I like the info the moodbar gives me but I find I want it more from the preview device than the main player (because I'm previewing stuff live to decide what to play). This PR refactors the moodbar somewhat to abstract out the dependencies on the player and its gui, and then applies another instance to the preview device, when that is activated. To make this work I have added a couple of events to the preview device, and made it keep a reference to its progress bar which the moodbar plugin can see.

*    I have not changed the minor version numbers of either of the plugins. I guess you will want to do that, if you accept this, to reflect the fact that they have been modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ doc/_build
 # Eclipse settings
 .project
 .pydevproject
+
+# Vim Pymode settings
+.ropeproject
+*.swp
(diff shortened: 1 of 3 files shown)


bitterjug:
Even now you've merged the indentation changes, this PR still shows all the changes from including the indentation changes. I've pulled the your last changes and I'll try rebase the previewmoodbar changes on that to get a better diff, and make it fast forwardable again.

virtuald:
You could have actually rebased and just pushed it to this branch, and used this PR instead of creating a new one. Github will automatically update it.

PR #81, opened 2015-05-04
Make moodbar plugin work with preview device

bitterjug:
I like the info the moodbar gives me but I find I want it more from the preview device than the main player (because I'm previewing stuff live to decide what to play). This PR refactors the moodbar somewhat to abstract out the dependencies on the player and its gui, and then applies another instance to the preview device, when that is activated. To make this work I have added a couple of events to the preview device, and made it keep a reference to its progress bar which the moodbar plugin can see.

*    I have not changed the minor version numbers of either of the plugins. I guess you will want to do that, if you accept this, to reflect the fact that they have been modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ doc/_build
 # Eclipse settings
 .project
 .pydevproject
+
+# Vim Pymode settings
+.ropeproject
+*.swp
--- a/plugins/moodbar/__init__.py
+++ b/plugins/moodbar/__init__.py
@@ -547,6 +547,7 @@ class ExModbar(object):
 
 
 def _enable_main_moodbar(exaile):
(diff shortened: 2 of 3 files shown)


virtuald:
Thanks, seems fine to me, sorry for the delay. I don't actually have the moodbar thing installed, but preview plugin still works so I'm ok with it. :)

virtuald:
We do need to do something about our plugin versions too... they're a bit out of whack at this point.

bitterjug:
You mean out of what with each other. It'd be good to have consistent numbering. What to other projects do? Is it really necessary to have separate versions for the ones shipped with Exaile? Perhaps treat them as part of the core and just change the Exaile version number when they change.

PR #86, opened 2015-05-14
Reduce loglevel in .desktop file

genodeftest:
When started from exaile.desktop file e.g. by opening a system menu or from autostart users tend to ignore exaile's output. Furthermore output (from logging) is put into syslog, exaile doesn't need to fill ("spam") syslog.
I won't change the general loglevel in main.py because when someone starts exaile from command line it should still print INFO messages.


virtuald:
I'm not opposed to this if nobody else is. @mathbr @sjohannes ?

sjohannes:
Looks good to me. I didn't know that stderr goes into syslog when it's run from .desktop.

genodeftest:
@sjohannes It does on Fedora. I don't know about other distributions.

PR #87, opened 2015-05-18
Trivial fixes

fidencio:
Two really simple patches:
1) Fix my own name on the "about" dialog.
2) Avoid to show an exception to the users, when they click in the "Track properties" with no track selected.


fidencio:
Self-NACKED.

Allowing to open the Track Properties dialog with no track selected can cause more problems than the exception showing up in the users face.
This situation makes me thing a bit:
Why do we have to have the "Track Properties" dialog duplicated? It increase the complexity of the code and is just redundant. Can I just remove it from the Tools menu?

PR #89, opened 2015-05-18
menu: Remove 'Track Properties' item

fidencio:
The main reason for removing the item is that the user has to be lucky
for clicking in the entry having a track selected.
Also, there is no other menu item depending on a track being selected.

The same functionality can be accessed by:
Right-Click on a track -> Properties


virtuald:
You could make it conditional so it would only show up when a track is already selected.

.. other than that, I don't think I care about this, so we can remove it.

PR #90, opened 2015-05-18
xlgui/properties: gtk3-ize the properties.py

fidencio:
A few changes were done if comparing to the orignal properties.
The most significative one being the format of the string indicating the
last time the file was modified, not it is presented in the preferred
representation for the current locale.


virtuald:
I hope you used [pygi-convert.sh](https://git.gnome.org/browse/pygobject/tree/pygi-convert.sh) to do most of that :)

PR #91, opened 2015-05-19
BREAK ALL THE THINGS

virtuald:
I'm not 100% convinced this is a good idea, but here it is anyways. The UI *starts* at least for me... I got this by running pygi-convert.sh and then fixing things manually until it worked.

Probably a better way of going about this is doing it a single file at a time instead of wholesale like so.

### DIFF
--- a/data/migrations/migration_200907100931/olddb.py
+++ b/data/migrations/migration_200907100931/olddb.py
@@ -34,7 +34,7 @@ except ImportError:
 
 #sqlite.enable_shared_cache(True)
 
-import glib
+from gi.repository import GLib
 
 class DBOperationalError(Exception):
     
@@ -168,7 +168,7 @@ class DBManager(object):
             Closes the connection
         """
         if self.timer_id:
-            glib.source_remove(self.timer_id)
+            GLib.source_remove(self.timer_id)
 
     def execute(self, query, args=None):
         """
@@ -228,7 +228,7 @@ class DBManager(object):
         if threading.currentThread().getName() == 'MainThread':
             self.db.commit()
         else:
-            glib.idle_add(self.db.commit)
+            GLib.idle_add(self.db.commit)
 
     def read_one(self, table, items, where, args):
         """
--- a/data/ui/widgets/track_info.ui
+++ b/data/ui/widgets/track_info.ui
@@ -62,7 +62,6 @@
               <object class="GtkBox" id="action_area">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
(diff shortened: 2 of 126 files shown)


virtuald:
Crap. All the pack_* that I 'fixed' I set the defaults to False... should be True to stay true to PyGTK.

virtuald:
I'm heading to bed, I think we should merge this once you're done reviewing it and just deal with the pain now instead of putting it off. Part of my rationale is that I'm *sure* at some point they'll get rid of `pygtkcompat`, and I'd rather not deal with that again.

The only reason I can think to not merge this is to make it easier for a dev to figure out what's been looked at and what hasn't. But.. that's what the wiki is for? And then I lost 4 hours of work.

virtuald:
I think I addressed most of your comments, with the exception of the get/set_data stuff. Heading out to work, feel free to address more of them.

sjohannes:
Remaining issues from my review:
* [ ] {get,set}_data - gone
* [ ] osd: pixmap - gone
* [ ] xlgui/preferences/widgets: enable_model_drag_{source,dest} - may need Gtk.TargetEntry.new
* [ ] xlgui/widgets/common: self.window.set_functions(Gdk.FUNC_RESIZE) - gone
* [ ] Search all occurrences of ".popup(" - order of params changed

New issues found from testing:
* [ ] Volume slider covers whole width
* [ ] Menu: stock items and mnemonics don't work
* [ ] Can't open prefs: `'PreferencesDialog' object has no attribute 'set_transient_for'` (xlgui/preferences/\__init__)
* [ ] Non-ASCII chars break in some places - see wiki. The error I've seen is `UnicodeDecodeError: 'ascii' codec can't decode ...`. List below if you find any.
  * [ ] Files panel (xlgui/panel/files)
  * [ ] Track info area (xlgui/widgets/info)
  * [ ] strxfrm used a few times in preferences/{\__init__,plugin} - probably easiest to test with non-latin locale

virtuald:
I've reviewed every python file and none report any lint import errors related to Gtk et al.

sjohannes:
Merged, remaining tasks moved to [main wiki page](https://github.com/exaile/exaile/wiki/PyGI-migration-tasklist-(GTK3-GStreamer-1.x)).

genodeftest:
Did you use a script to do that by the way? Or did you manually change all the references?

virtuald:
There was a script, pygi-convert... I think it's in the pygobject tree somewhere.

genodeftest:
Thanks!

PR #101, opened 2015-06-06
Port glade files to Gtk 3.10, remove many deprecations

genodeftest:
This updates all .ui files by exaile (not including plugins) to Gtk 3.10.
There are some minor changes to layout, most without effects to python code.

Some changes are just reorderings generated by glade without effects. If you
prefer to not commit those changes, just give a comment.

Major changes:
* Deprecations:
  * GtkHBox, GtkVBox replaced by GtkBox
  * GtkHButtonBox, GtkVButtonBox replaced by GtkButtonBox
  * GtkTable replaced by GtkGrid
  * GtkHSeparator, GtkVSeparator replaced by GtkSeparator
  * GtkMisc:xalign, GtkMisc:yalign, GtkMisc:ypad replaced by halign or removed where unnecessary
  * GtkAlignment removed
  * GtkStock: stock icons replaced by new freedesktop.org icon names
* Some simplifications of unnecessary stacked Boxes
* Preferences dialog: use GtkGrid where possible to make layout easier to read
* added icons to some buttons

still left TODO:
(I will add these into the wiki page or as separate bug reports once my pull request is finished. I'm posting them here because there is no discussion on the wiki page.)
* GtkTable is deprecated and should be replaced in those two files and I don't think I can do that, but this is low priority:
  * trackproperties_dialog.ui
  * widgets/tracklist_info.ui
* I get this error when opening the "About" dialog:
   ```
xlgui/widgets/playback.py:892: TypeError: Passing arguments to gi.types.Boxed.__init__() is deprecated. All arguments passed will be ignored.
  Gdk.Event(e)))
   ```
* I found a bug newly introduced in the "gi" branch which happens when trying to refresh the collection on 
  ```
ERROR   : Error reading tags for file:///path/to/file/sound.wav
  File "/path/to/exaile/xl/trax/track.py", line 347, in read_tags
    f = metadata.get_format(loc)
  File "/path/to/exaile/xl/metadata/__init__.py", line 122, in get_format
    return formatclass(loc)
  File "/path/to/exaile/xl/metadata/_base.py", line 78, in __init__
    self.load()
  File "/path/to/exaile/xl/metadata/wav.py", line 50, in load
    ext = os.path.splitext(loc)[1][1:].lower()
  File "/usr/lib64/python2.7/posixpath.py", line 98, in splitext
    return genericpath._splitext(p, sep, altsep, extsep)
  File "/usr/lib64/python2.7/genericpath.py", line 99, in _splitext
    sepIndex = p.rfind(sep)
AttributeError: 'NoneType' object has no attribute 'rfind'
  ```
* GtkImageMenuItem has been deprecated since GTK+ 3.10. Quote from documentation:
  If you want to display an icon in a menu item, you should use GtkMenuItem and pack a GtkBox with a GtkImage and a GtkLabel instead. You should also consider using GtkBuilder and the XML GMenu description for creating menus, by following the GMenu guide. You should consider using icons in menu items only sparingly, and for "objects" (or "nouns") elements only, like bookmarks, files, and links; "actions" (or "verbs") should not have icons.
  what do we do? get rid of menus with images or build a wrapper in xlgui/widgets/menu.py simple_menu_item ? Right now the main application does not use any GtkImageMenuItems, so I would prefer dropping support for images in menus completely.
* plugins. I think I'll do that later
* Device manager UI is not intuitive:
  * There are some buttons shown and active, but they have no effect.
  * As long as no device is connected there should be a message at least saying something like "No known devices connected" and list supported device properties (MTP?)
* Cover manager is still broken (was before, and I can't fix that)

Please read and comment before merging. I don't have much experience in Gtk, Glade and Python.

_EDIT_: This is part of #3 .


virtuald:
This sounds great. I won't be able to test it until Tuesday night, currently traveling.

sjohannes:
Thanks; I really like what you've done with the layout and alignments of the Prefs dialog.

Window type hints for (I think) all Exaile subwindows should be `dialog`, to be consistent with the [EWMH](http://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472629520). `utility` is meant for persistent windows like GIMP's toolbox; "[UTILITY is] distinct from type DIALOG because it isn't a transient dialog, the user will probably keep it open while they're working."

Another concern I have is on setting `enable_grid_lines=horizontal` on the prefs tree. I like it visually, but is there precedent for it? I'm genuinely clueless because I don't use too many GTK+ 3 apps. Should other listviews in Exaile have gridlines as well?

> I get this error when opening the "About" dialog: `xlgui/widgets/playback.py:892: TypeError: Passing arguments to gi.types.Boxed.__init__() is deprecated. All arguments passed will be ignored. Gdk.Event(e)))`

Yeah, that's part of the "Fix re-emitting events" in my todo list in the wiki. It happens when you have this sort of code:

    def some_gtk_signal_handler(self, event):
        self.foo.emit('some-gtk-signal', event)

In PyGI the `event` we get is some weird object that is a bit like GdkEvent but not really, causing the `emit` call to fail. In some previous version, wrapping it in the `Gdk.Event` constructor fixed this, so that's what I've done. At some point this broke again and I haven't found a proper fix.

> Cover manager is still broken (was before, and I can't fix that)

The error I get when trying to open the cover manager is `RuntimeError: Data access methods are unsupported. Use normal Python attributes instead`. This is the "{get,set}_data" point in the wiki.

genodeftest:
With `dialog` vs. `utility`: You are right. The Gdk documentation is not precise on that. I'll fix that later, changing everything (except the main window) to `dialog`. It is better to use `dialog` for user experience anyway since `utility` does not grab focus on opening.

I put `enable_grid_lines=horizontal` because I like it too ;)
It is not used consistently in GNOME: Nautilus (file manager), contacts, extended preferences use it. The file chooser doesn't, and evolution (mail client) partially does.
It is not used consistently in exaile either: The collection panel does, most other panels and the playlist don't. Maybe we should decide to either use it everywhere or nowhere. I vote for "everywhere" since it improves readability.

To the other errors: Thanks, I didn't notice that.

genodeftest:
Oh, and I should add that as a result of removing GtkStock there will be much more text to translate. I don't like that but the Gtk devs decided to deprecate GtkStock. And I think it is more important to not carry deprecated code. Feel free to comment if you think different.

sjohannes:
Ok, I'm fine with the gridlines in the prefs treeview for now.

For the main playlist treeview, we currently use [`rules-hint`](https://developer.gnome.org/gtk3/stable/GtkTreeView.html#GtkTreeView--rules-hint), which is supposed to alternate colours between rows, but that's also deprecated in 3.14 and doesn't seem to have an effect anymore.

Regarding GtkStock, it's not a huge priority but yeah, we can start replacing them.

virtuald:
Looks good to me, thanks for the fixes! Sorry for the delay, I'm busy until probably the end of the month or so...

virtuald:
FWIW, I don't really like the gridlines in the treeviews, it looks rather strange on Fedora 21 w/black adwaita theme. But I don't have a strong enough opinion to actually change it yet.

PR #106, opened 2015-06-15
[gi] Preferences→Plugins: reduce size of description, reduce UI elements

genodeftest:


sjohannes:
I'm generally ok with this patch. I don't really like the bottom part growing based on the number of lines in the description, but it's probably still better than the current situation of having it take up excessive space.

I quite like [how Pidgin does it](http://cdn5.howtogeek.com/wp-content/uploads/2012/01/psychic-mode-plugin.png). Each entry in the list has one row for the plugin title and another row for a short description. Other information (author, website, long description) are only visible on the tooltip or if you expand the Plugin Details expander. That goes beyond just a UI file change, though.

genodeftest:
Hmm, I could add a minimum size. Problem is that it can only be given in pixels, I'd prefer to set a minimum number of lines. How about 3 lines (approximately 60 pixels)?
Personally I don't like how pidgin does it because
a) it is limited to mouse users (have you ever seen – or used – tooltips on a touch device?)
b) it adds another click for the users

genodeftest:
I think I should be more precise on my previous comment.

What I like about the way how pidgin does it is that you see the description below the title. This way title, author, version and description always stay together.
What I don't like is the "plugin details" thing below where you have to do another click to see the detailed description.

I have some more ideas on how to redesign the preferences/plugins page, which still need a mockup. Since you don't really like it and nobody else has a (strong) opinion on it you may also close this pull request if you prefer to.

sjohannes:
Actually I'm ok with it. I asked Dustin yesterday to have a look and he also said it's fine, but the patch doesn't apply anymore. Could you update it for the latest revision of the `gi` branch? It may be easier to redo your changes rather than try to fix the conflict.

genodeftest:
Ok, I'll redo my changes on the current `gi` branch later. Will take a while.

PR #112, opened 2015-06-18
[gi] Fix main menu button warning

genodeftest:
Fix the warning `TypeError: get_menu_position() takes exactly 3 arguments (5 given)` when opening and closing main menu


sjohannes:
This would actually cause an error in GTK+ < 3.16. It was due to an API-breaking change made in GTK+'s introspection data since 3.16. See [pgi-docgen ticket](https://github.com/lazka/pgi-docgen/issues/92), [our workaround](https://github.com/exaile/exaile/commit/5fb96af38f712c08b65661b5ce7cdc75dc580dcb).

I've committed something that works in both as 9091e20.

genodeftest:
Ok, I didn't know that. Thanks!

Issue #104, opened 2015-06-15
Plugin "Equalizer": profile "Full Treble": value out of range


genodeftest:
I just noticed that in "Full Treble" the values for band6, band7, band8 and band9 are out of range. Limit is −24 dB … +12 dB and they are 16, 16, 16 and 16.8. See https://github.com/exaile/exaile/blob/master/plugins/equalizer/__init__.py#L346 for a place where to fix this.

sjohannes:
I wonder where these values come from in the first place; they've been there since the first commit at 967d36a. I guess we can add a 5 dB preamp and adjust accordingly.

sjohannes:
After testing @genodeftest's patch at #115, which implemented what I suggested above, it's apparent that it was a bad idea. It was overly loud and caused clippings with the track that I tried.

Meanwhile, I've discovered where the preset values came from: [XMMS](http://legacy.xmms2.org/), which itself took the values from WinAMP. I wonder if there was something wrong with that initial conversion.

Frankly the presets have never made sense to me. They can be useful for quickly tweaking the audio to your preferred style (e.g. bass boost, vocal boost), but the genre-based ones are arbitrary at best. @virtuald: what do you think about removing them or replacing with a better alternative (from where)?

virtuald:
Perhaps the limits in the other audio players were different, but whoever copied them just did so blindly? We could do a survey of presets from different audio players perhaps... 

I don't care too much what is done with the equalizer. It's a bit clunky to use in its current form anyways (namely, in a separate window), so I don't typically use it. Others have mentioned that it causes problems (#75), not sure if those problems would be resolved in GST 1.x or not.

sjohannes:
I've modified all the preset values to be within the -24..12 range, as well as to be closer to each other in terms of loudness.

PR #115, opened 2015-06-19
[gi] Easy fix for #104

genodeftest:
This is an easy fix for #104. If you don't like it, just close it, I won't mind.


sjohannes:
This is way too loud; I can't justify pushing this to users, even if it's what the preset creator intended.

I'll leave further notes in #104.

PR #116, opened 2015-06-19
[gi] Plugins: some minor fixes

genodeftest:
Improves plugin dependencies, makes Lyrics Viewer and Wikipedia run at the same time and fixes daapserver imports so daapserver can start

### DIFF
--- a/plugins/contextinfo/PLUGININFO
+++ b/plugins/contextinfo/PLUGININFO
@@ -3,4 +3,4 @@ Authors=['Guillaume Lecomte <guillaume86@gmail.com>']
 Name=_('Contextual Info')
 Description=_('Show various informations about the track currently playing.\nDepends: libwebkit >= 1.0.1, python-webkit >= 1.1.2, python-imaging (a.k.a. PIL)')
 Category=_('Information')
-RequiredModules=['PIL', 'webkit']
+RequiredModules=['PIL']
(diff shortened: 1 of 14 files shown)


plugins/contextinfo/PLUGININFO
```diff
@@ -3,4 +3,4 @@ Authors=['Guillaume Lecomte <guillaume86@gmail.com>']
 Name=_('Contextual Info')
 Description=_('Show various informations about the track currently playing.\nDepends: libwebkit >= 1.0.1, python-webkit >= 1.1.2, python-imaging (a.k.a. PIL)')
 Category=_('Information')
-RequiredModules=['PIL', 'webkit']
+RequiredModules=['PIL']
```
virtuald: This seems odd. Shouldn't there be *something* here?

sjohannes: It's currently not possible with the way we detect plugin compatibility, partly because `imp.find_module` doesn't support submodules, and partly because GI modules are imported in a super special way.

The only code that I find works to detect the presence of GI modules without loading it is this:

```py
import gi
has_webkit2 = bool(gi.importer.repository.enumerate_versions('WebKit2'))
```

So perhaps we can add a special `gi:WebKit2` syntax or something.

genodeftest: Did you mean extending the `RequiredModules` functionality to add gi:WebKit2? How about adding another field to PLUGININFO, e.g. `RequiredGiModules`? It would be useful to add another one too for dbus services (for plugins awn, currentsong, …), e.g. call it `RequiredDbusServices`.

virtuald: Moved discussion to https://github.com/exaile/exaile/issues/121

PR #123, opened 2015-07-20
reduce size of description, reduce UI elements (again)

genodeftest:
I accidentially closed #106, this is the change from there.

### DIFF
--- a/data/ui/preferences/plugin.ui
+++ b/data/ui/preferences/plugin.ui
@@ -149,7 +149,7 @@
                   </object>
                   <packing>
                     <property name="left_attach">0</property>
-                    <property name="top_attach">1</property>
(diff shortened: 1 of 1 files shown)


virtuald:
This behavior seems fine to me, with the exception that having the two checkboxes move over and over again is a bit annoying. 

I wonder if we should move them to somewhere they won't move (maybe top or bottom). I feel like we're [bikeshedding](http://bikeshed.com/) this a bit too much though...

virtuald:
@sjohannes ?

genodeftest:
You're right. I second moving them up or down. Any preference?

sjohannes:
No preference between top, middle, and bottom for me.

virtuald:
Thanks!

PR #127, opened 2015-08-01
Add year column

virtuald:
@SaiyanRiku try this out and let me know if it does what you want.


SaiyanRiku:
Ok thank you very much, i will try tonight to install it.

PR #128, opened 2015-08-04
Gi engine refactor

virtuald:
Ok, lots to say here, but this seems to be in fairly good shape. I've completely gutted and refactored the Exaile player object(s), replacing the normal and unified engines with a single gstreamer engine, and creating a player interface that is engine independent if we decide to get rid of gstreamer in the future or add support for other engines.

This single engine supports all of the things:

* Crossfading
* Gapless playback
* Dynamic audio device switching
* start/stop offsets for audio files

Known bugs:
* [x] Haven't added audio filters back in yet, need to fix the plugins too
* [x] Need to add the 'user fade' setting back in
* [x] For some reason, when I start a track pulseaudio sets the system volume to 69% -- doesn't do so when I comment out the volume setting code 

I've implemented fading with a separate object, so that makes it a bit easier to grok. The crossfading is implemented by sending the output of two playbins to separate audiosinks in two separate pipelines. It seems to work for me on Linux using pulse, haven't tried anything else yet.

@sjohannes or @mathbr if you could code review this, I'd greatly appreciate it. Anyone else who wants to use it and report bugs (@genodeftest), please do.


sjohannes:
1. Start Exaile with `--debug`.
2. Start playing a track, preferably a very short one.
3. Wait until the "Will fade out in..." message appears.
4. Start playing another track, say track number 10 of an album.
5. Stop playback after a second or so.
6. Wait out the timer from step 3.

Track number 11 will play by itself.

xlgui/preferences/widgets.py
```diff
@@ -918,7 +956,8 @@ def on_clicked(self, button):
 
 class ComboPreference(Preference):
     """
-        A combo box
+        A combo box. The value stored in the settings must be the
+        first row of the combo box model.
```
sjohannes: Isn't it first column?

tests/xl/player/gst/test_fader.py
```diff
@@ -0,0 +1,178 @@
...
+
+def check_fader(test):
+
+    stream = FakeStream()
+    fader = TrackFader(stream, stream.on_fade_out, 'test')
+    
+    for data in test:
+        print data
+        now = data[0]
+        stream.position = int(now*1000000000)
```
sjohannes: Is this our own unit? We should probably have a constant similar to `Gst.SECOND` to make it easier to read.

xl/common.py
```diff
@@ -856,4 +856,40 @@ def write(self, s):
...
+        
+        :param section: Settings section
+        :param options: Dictionary of key: option name, value: attribute on
+                        'self' to set when the setting has been updated. The
+                        attribute must already have a default value in it
+        :param self:    Object to set attribute values on
+        
+        :returns: A function that can be called to unsubscribe
+        
+        .. versionadded:: 3.5.0
```
sjohannes: Opinions on making the next version 4.0?

virtuald: I'm open to the idea. It has been a major revision effort.

virtuald:
Figured out what was annoying about the Pulse thing... not quite a bug, but apparently in 'flat-volume' mode (default on most Linux distros now) when Exaile changes the pulsesink's volume via playbin, it has the potential to change the *system* volume too. However, this behavior is much better now that I've updated the engine to change the volume slider when the volume changes.

There are various opinions on whether the flat-volume thing is a good thing or not, but seeing that it is what it is, I think this is the best we can do under the current circumstances.

Found some recommendations about how to deal with some of this: http://cgit.freedesktop.org/pulseaudio/pulseaudio/commit/?id=3535fd7a076228fdeae3755cf8ac6fdcfa28741d ... hadn't thought of the possible OSD notifications, so we might want to add an additional volume control element and change that instead of changing it at the sink level. Going to put that off until someone complains about it, I think.

virtuald:
Alright, @sjohannes I think this thing is ready for a final look, and then we can merge it in and move on to fixing the rest of the bugs in this branch. It does everything I want it to, seems cleaner, and I haven't had any problems in my limited testing.

One remaining problem is upgrading our custom Windows sink, but I'm going to put that off for now, and try to get a minimal change included in gstreamer mainline.

I feel like you mentioned that there was a slight 'noise' when resuming playback? Might be because I removed the set_volume stuff from the queue... though I don't notice it on my machine. I had forgotten that I had removed it until just now... though, if I drop the volume there, probably want to drop the volume for implementing __start_offset too?

sjohannes:
It occurs on startup if the "Resume playback on start" and "Resume playback in paused state" options are enabled. A very short segment from the beginning of the track is played, the seek happens (to the resume point), then another segment from the new position is played.

Tested with and without crossfading on Windows; haven't tested on Linux.

eri-trabiccolo:
If interested, on Linux I'm unable to reproduce this.(audiosink: ALSA)

virtuald:
@sjohannes I've pushed/rebased again -- more documentation, and I think this should remove the noise you were complaining about, though I don't have a machine to test it on.

virtuald:
Of course, the rebase makes it harder to see what I changed. Links to the changes: [queue change](https://github.com/virtuald/exaile/commit/b168b66a95a608d8e41ca5fba9414f197e28b160) and [engine change](https://github.com/virtuald/exaile/commit/575f0c9d2ad0808da86ab4e0ea6371db5d490741)

virtuald:
Alright, sounds like initial feedback is good. One more rebase, and it's getting merged so I can move on.

PR #129, opened 2015-08-10
Fix: small fixes for the lyricsviewer plugin:

eri-trabiccolo:
- use event coords as mouse pointer coords on the source url textview
- avoid warning when trying to remove a glib source already dismissed
- expand the scrolling window container

### DIFF
--- a/plugins/lyricsviewer/__init__.py
+++ b/plugins/lyricsviewer/__init__.py
@@ -193,8 +193,8 @@ class LyricsViewer(object):
         cursor_type = window.get_cursor().get_cursor_type().value_name
 
         if self.source_url != "":
-            x, y, mod = window.get_pointer()
-            x, y = textview.window_to_buffer_coords(Gtk.TextWindowType.TEXT, x, y)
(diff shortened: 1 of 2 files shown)


plugins/lyricsviewer/__init__.py
```diff
@@ -248,8 +248,13 @@ def do_update(refresh):
                 self.set_top_box_widgets(False, True)
             return False
 
-        if self._lyrics_id != 0:
+        g_thread = GLib.main_context_get_thread_default() or \
+                        GLib.main_context_default()
+
+        if ( self._lyrics_id != 0 and 
+                g_thread.find_source_by_id(self._lyrics_id) ):
```
virtuald: This seems suspect, instead at the end of `do_update` try setting self._lyrics_id to 0.

eri-trabiccolo: Why suspected? You mean g_thread or the find_source_by_id ?
Anyway resetting _lyrics_id should make sense yeah.. thanks

virtuald:
Fixed this elsewhere, omitted the strange thread stuff. If there's more to be fixed, please do so in a separate PR. Thanks!

PR #133, opened 2015-09-07
Update some PLUGININFO files

genodeftest:
Update RequiredModules, update Description

Note that I removed a dependency by moodbar plugin on moodbar since the moodbar module is part of the moodbar plugin. Please check that this is right before merging.

### DIFF
--- a/plugins/contextinfo/PLUGININFO
+++ b/plugins/contextinfo/PLUGININFO
@@ -3,4 +3,4 @@ Authors=['Guillaume Lecomte <guillaume86@gmail.com>']
 Name=_('Contextual Info')
 Description=_('Show various informations about the track currently playing.\nDepends: libwebkit >= 1.0.1, python-webkit >= 1.1.2, python-imaging (a.k.a. PIL)')
 Category=_('Information')
-RequiredModules=['PIL']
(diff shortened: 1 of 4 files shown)


sjohannes:
Exaile already requires the `cairo` module, so it should be omitted. (The convention is to omit anything that Exaile already requires to run, e.g. gi:Gtk, gi:Gst, mutagen.) I'm not sure about Atk (it's a GTK+ dependency that we don't use directly) but I'm ok with it being added.

The *moodbar* dependency refers to the `moodbar` app that the plugin needs, so don't remove it from the description.

genodeftest:
Oh, you are right, Atk is a hard dependency of Gtk.
And I'll undo the `moodbar` change.

Regarding cairo: You are right about that too. How about including the DEPS file (or a part of it) in the doc/dev/plugin.rst file so that a plugin developer knows which modules he/she can rely on?

sjohannes:
Yeah, I think it would be useful. We don't import too many modules; from the DEPS file I see dbus, gi:Gst, gi:Gtk, and mutagen. And then the only extra import I can think of is cairo, which I'm pretty sure is required by gi:Gtk anyway.

PR #135, opened 2015-09-12
Add a 'loading' indicator when loading large playlists

virtuald:
@sjohannes can you take a look at this? Seems to work great for those pesky really large playlists, not sure if I missed an important race condition anywhere...


sjohannes:
Looks good to me. Exaile still locks up for a bit if I open multiple large playlists at the same time, but that may be the GtkListView drawing itself.

PR #140, opened 2015-09-17
Install and import fixes

sarnold:
Two "minor" fixes to get plugins working and banish import errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,11 +78,11 @@ install: install-target install-locale
 install_no_locale: install-target
 
 install-target: make-install-dirs
-	install -m 644 exaile.py ${EXAILELIBDIR}	
+	install -m 644 exaile.py ${EXAILELIBDIR}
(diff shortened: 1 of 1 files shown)


xl/settings.py
```diff
@@ -37,7 +37,7 @@
 import logging
 import os
 import sys
-
+import dbus
```
virtuald: Hm? Settings shouldn't depend on dbus.

sarnold:
It coughs a missing dbus warning starting up in a console; but I couldn't really find a better place.

virtuald:
There is no dbus module in windows, so importing it in settings is probably a bad idea.

sarnold:
I think it was a side-effect of the missing pylast, seems like the install fix took care of it (since I can't reproduce it now after commenting that out in my test package).  Nevermind...

sarnold:
Okay, bye-bye dbus commit...

PR #145, opened 2015-10-06
Collection Panel: use GtkStack, center message

genodeftest:
Use a GtkStack instead of adding and removing widgets
Center the "empty collection" message

TODO: change wiki entry for "empty collection" message after this PR is merged.

### DIFF
--- a/data/ui/panel/collection.ui
+++ b/data/ui/panel/collection.ui
@@ -17,15 +17,14 @@
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">Collection</property>
     <child>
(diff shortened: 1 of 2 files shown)


virtuald:
I merged this in without checking to see if it works, and it looks like it doesn't actually work. I like the idea of using GtkStack though.

Anyways, I had pushed a fix for the vertical space before noticing that you had done this PR, so I pushed that in instead.

sjohannes:
Looks like the commit was reverted/removed; that's why the PR gets automatically closed but the commit isn't in the current `gi` history. Could you rebase (or might have to redo) the GtkStack changes?

Issue #70, opened 2015-04-02
Add MANPREFIX Make option


sjohannes:
According to #69, manpages in FreeBSD are located in `$PREFIX/man`, while in Linux it's `$PREFIX/share/man`.

sarnold:
Since I've been having so much fun with make and autotools lately, I went ahead and added this to master (commit is on top of pull request #184) but it took a little cleanup and a compatibility fix to the launcher script to make it work correctly.  (see commit msg, it was a weird bug...)

mexicarne:
That's great news, thank you!

sjohannes:
My comments to 8fb10cc ended up cluttering #148, which is really about a different issue, so I'll move them here. I've also added an extra reply to the `.NOTPARALLEL` stuff.

```diff
-PYTHON2_CMD	?= `command -v python2`
+PYTHON2_CMD	= `command -v python2`
 
-PREFIX 		?= /usr/local
-EPREFIX		?= ${PREFIX}
-LIBINSTALLDIR 	?= /lib
-XDGCONFDIR 	?= /etc/xdg
+PREFIX		= /usr/local
+EPREFIX		= $(PREFIX)
+LIBINSTALLDIR	= $(EPREFIX)/lib
+DATADIR		= $(PREFIX)/share
+XDGCONFDIR	= /etc/xdg
+MANPREFIX	= $(PREFIX)/share
```

I tried replacing all the `=`s here to `?=` and running `make PREFIX=foo` and it seem to still work. As far as I know, the only difference between `=` and `?=` is that `?=` allows reading values from the environment, i.e. `PREFIX=foo make` instead of `make PREFIX=foo`, so I don't understand how you're seeing the problem you mentioned in the commit log.

```diff
+.NOTPARALLEL:	install-target install-locale make-install-dirs
```

>> The [GNU Make manual](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#index-_002eNOTPARALLEL) suggests that just the presence of `.NOTPARALLEL:` causes the whole makefile to be run sequentially (and that any prerequisites are ignored). I've never had problems with `make -j`; what is `.NOTPARALLEL` fixing here?

> It probably went in early in debugging the missing "/usr" path, but in this case it should be safer than not (or relying on upstream packagers like me to add MAKEOPTS=-j1 as a workaround). Otherwise your two different make targets have a potential race condition between creating the target dirs and installing to the same dirs

[...]

PR #148, opened 2015-10-08
Fixes for Makefile

sarnold:
for amazoncovers plugin to work.  I forgot about this one, and I promise to start testing gi branch now...  really...

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -32,16 +32,30 @@ translations.
 Environment Variables
 ^^^^^^^^^^^^^^^^^^^^^
 
+Note: normally it should be enough to set PREFIX, and potentially LIBINSTALLDIR
+on a 64-bit system.  The additional variables can provide further installation
+customization.
+
 Environment variables that affect "make install":
 
   LIBINSTALLDIR
-    The lib directory to be appended to PREFIX. 
-    Default: /lib
+    The full path to the lib directory.
+    Default: EPREFIX/lib
+  DATADIR
+    The full path to non-arcitecture dependent data files.
+    Default: PREFIX/share
+  MANPREFIX
+    The full path to the parent of the man page installation directory (same
+    as system DATADIR)
+    Default: PREFIX/share
+  EPREFIX
+    The executable installation prefix.
+    Default: PREFIX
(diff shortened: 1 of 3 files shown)


virtuald:
Unless I'm reading it incorrectly, nothing in that file uses lxml, why does the import need to happen there?

sarnold:
AFAICT amazon cover lookup is broken without it; the imports are a little bit different between 0.3x and 3.x/HEAD but it pretty much needs to go somewhere (_ecs maybe?).  It also needs valid AWS credentials, but Amazon lookup works in both old and new with the patch.

sarnold:
Top commit fixes issue #70 (see commit msg for interesting details).

sarnold:
It probably went in early in debugging the missing "/usr" path, but in this case it should be safer than not (or relying on upstream packagers like me to add MAKEOPTS=-j1 as a workaround).  Otherwise your two different make targets have a potential race condition between creating the target dirs and installing to the same dirs (and in this case you lose nothing since you're not really compiling anything that's worth doing in parallel in the first place, plus it's not worth the maintenance headache for a static makefile; better to switch to autotools at that point).

sarnold:
Oh, and the rest of the makefile updates are to make it more consistent and flexible for OS/package maintainers (as well as developers) as it fixes an inconsistency in how EPREFIX and EXAILE* dirs were being used.  The makefile cleanup ended up getting a /usr/usr/blah path in the launcher script, as it changes the args to be more self-consistent/correct.  Did I mention static makefiles were a maintenance headache?  Autotools *does* do python, or distutils/setuptools are another option.

sjohannes:
I've moved my #70-related comments there; I didn't expect them to appear in this PR.

sarnold:
Turns out it does work in the latest version (without the import).  Thought I tested it in that version, but apparently I just migrated the patch.

virtuald:
@sarnold if you could rebase this down to a single commit, that would be great and we can merge it in to master (unless @sjohannes has additional objections).

sarnold:
Check.

virtuald:
Thanks!

PR #150, opened 2015-10-11
Fix: tags not read for opus files

Zarokka:
Tags of opus files could not be read.
This small changes fix it (at least my manual tests where successful, but my python knowledge is limited, so please review thoroughly)

An example file could be found here:
http://xiph.org/~giles/2012/opus/detodos.opus

thx


sjohannes:
Thanks; could you move the Vim modeline (the line that starts with `# vim: `) to the end of the file, or simply remove it?

Zarokka:
sure, done, thx

Issue #58, opened 2015-03-10
Is grouping by album artist supported?


cpitclaudel:
I've found a few references to this on Ubuntu forums, including a post suggesting that it is supported, but I can't find the relevant option in the latest master. There also seems to be an old patch on launchpad; I didn't test it.

Is there a way to group the left pane by Album Artist?

virtuald:
I don't see the option either. If you link to the old patch, perhaps we can bring it in.

cpitclaudel:
I was thinking of this 2010 bug discussion on exaile's [launchpad tracker](https://bugs.launchpad.net/exaile/+bug/135618)

The [patch](https://launchpadlibrarian.net/57296546/collection-album-artist.patch) itself is in [this post](https://bugs.launchpad.net/exaile/+bug/135618/comments/12).

sbrubes:
Related: https://github.com/exaile/exaile/issues/414

virtuald:
Fixed in #576, please try out Exaile 4 and let us know if this has been addressed for you.

cpitclaudel:
Thanks!

PR #159, opened 2015-10-20
Change to xlgui/panel/collection.py adds "Album Artist" sort (Issue #58)

tfree87:
Adding these lines to the python code for the panel addressed this issue for me. As long as your album artist tags are correct, this seems to allow for sorting by album artist as opposed to just the track artist. Works on my Fedora 22 system.


virtuald:
Heh, is it really that easy?

tfree87:
Ran Exaile again with the new code. It seems as long as all of the tracks have an "albumartistsort" tag, that the fix works. If "album artist" is used instead, however, it does not seem to display all the tracks. I am not entirely sure if all users will have their files tagged with "albumartistsort". I use the beets command line program to tag all of my files with Musicbrainz and it seemed to add this tag, however, other users may have to add it manually. I will try to see if I can figure out why just using "albumartist" produces weird output.

virtuald:
Any word on this?

tfree87:
I haven't been able to make any progress on this. What I have found is that in a compilation where the Album Artist is the same but the track artist varies the subtrees in the collection do not display the tracks where the artist does not match the album artist. The top tree shows the correct counts for the number of tracks, but these do not display in the subtrees. For example say there is an album where the album artist is 'Foo' and there are three tracks. The tracks are by 'Foo', 'Foo' and 'Bar', and 'Foo featuring Bar'. When albumartist is used the top level (albumartist) shows three tracks ('Foo (3)') but the subtrees album and track only display the track by 'Foo'.

This does not happen when albumartistsort is used. I cannot figure out why this behavior  is exhibited. The track search is obviously finding all the tracks as seen by the top level of the collection tree but when subtrees are opened they fail to display all the tracks.

tfree87:
Here is an example, when the an album is used with the first sorting tag as 'albumartistsort' you get the following which is what you would expect:
albumartistsort
However, the same album when 'albumartist' is used produces the following output:
albumartist
There does not appear to by any obvious reason why this is happening from what I can see. It works in one case and not the other.

Issue #83, opened 2015-05-11
Lots of thread related issues


virtuald:
There are lots of places in Exaile's source code where we don't deal with multithreaded code in an appropriate/safe way. I'm a culprit just as much as anyone else when it comes to these things, but we really should deal with them at some point.

In particular, we should never interact with GTK widgets unless we're on the main thread. There's lots of places where this accidentally can happen, because our eventing system issues events from arbitrary threads -- for example, from a GST thread, or from a collection scan thread, or whatever.

virtuald:
One thing I've been contemplating is creating a debugging tool that attaches itself to all the GTK widgets and checks what thread it's in when methods on them get called. Could be useful for tracking some of these down.

virtuald:
After looking into this, apparently at one time the events were executed on the idle thread. Then in 9976a20e0ef9a1c82cd93b6fa13042f6788d4f98, e2c24b625718011b93481e8f32d5c58b3b3abe21 it looks like events were made to be synchronous, which resulted in the situation we have now.

At one time, the gtkrun decorator existed in guiutil, but was deprecated in 02e63dadaf9433c1cd56246cd507eac2d1e2b5a4, which was before events were made synchronous. There's a comment there about not needing this for events anymore -- but that's no longer true now that events are synchronous.

I think the right way to fix events (which will in turn then fix most of our thread related issues I suspect) is to keep them synchronous as they are now, except allow handlers to mark themselves as 'UI' via a parameter to event.add_callback, in which case they will be called in a UI-safe way. If you use threads_enter/threads_exit, the problem of course is that the GTK lock is not reentrant on Linux. So.. probably call it via idle_add instead of using a gtkrun-type of operation?

virtuald:
For future reference: https://wiki.gnome.org/Projects/PyGObject/Threading

genodeftest:
I just ran into a non-reproducible crasher (SIGSEGV) when opening and closing cover manager window repeatedly. I guess it is a threading issue too, but I can't tell.

genodeftest:
Adding many files to playlist (e.g. from files or library panel) is freezing exaile. Looks like the adding (and possibly reading metadata from files) work is done in main thread.

genodeftest:
Some crashers are caused by xlgui/cover.py which is not correctly threaded. I'll be working on that.

genodeftest:
I am trying to fix xlgui/cover.py, but I can't get `event.log_event()` working when send to a different class/object.
[...]

PR #161, opened 2015-10-21
event: add first-class support for UI callbacks

virtuald:
* Add add_ui_callback function to event to signify that you only want to receive events on the UI thread
* No more idle_add required, easier to audit

This is a first step towards fixing #83. I'd love feedback for making this more efficient, this was the first implementation that popped into my head. In particular, I wonder if there's a way that this could be implemented without using three separate dictionaries... I didn't want to keep checking to see if a callback was 'ui' or not, though, perhaps it's not that big of a deal. 

Fixed some correctness stuff in the weakref usage also.

@sjohannes , @genodeftest , thoughts?

Once this gets in, then I think the next step is to convert most of the UI classes to use add_ui_callback instead of add_callback... maybe plugins too. Then will need to audit each usage of common.threaded to ensure that they aren't touching GTK widgets.

### DIFF
--- a/xl/event.py
+++ b/xl/event.py
@@ -48,10 +48,7 @@ import time
 import weakref
 from gi.repository import GLib
 
-from xl import common
-from xl.nls import gettext as _
-
-# define this here so the interperter doesn't complain
+# define this here so the interpreter doesn't complain
 EVENT_MANAGER = None
 
 logger = logging.getLogger(__name__)
@@ -60,6 +57,9 @@ class Nothing(object):
     pass
 
 _NONE = Nothing() # used by event for a safe None replacement
+
+# Assumes that this module was imported on main thread 
+_UiThread = threading.current_thread()
 
 def log_event(type, obj, data):
     """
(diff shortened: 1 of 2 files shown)


xl/event.py
```diff
@@ -220,7 +248,10 @@ class EventManager(object):
         Manages all Events
     """
     def __init__(self, use_logger=False, logger_filter=None, verbose=False):
+        # sacrifice space for simplicity
```
sjohannes: This really sacrifices space for a bit of `emit()` speed, as you can always call `_emit` twice (once with `callbacks`, once with `ui_callbacks`).

xl/event.py
```diff
@@ -297,18 +351,21 @@ def add_callback(self, function, type, obj, args, kwargs):
...
-            if obj is None:
-                obj = _NONE
-            try:
-                callbacks = self.callbacks[type][obj]
-            except KeyError:
-                callbacks = self.callbacks[type][obj] = []
-
-            # add the actual callback
-            callbacks.append(Callback(function, time.time(), args, kwargs))
+            for cbs in [self.callbacks, self.all_callbacks, self.ui_callbacks]:
```
sjohannes: This adds `cb` to all three variables; you just want it on `all_callbacks` and one of the others.

genodeftest:
Sorry, I can't help with this topic. I don't have the expertise.

xl/event.py
virtuald: Right, ha.

sjohannes:
Looks good to me.

virtuald:
Ok, one more look at this. I've modified the code so that callbacks aren't called while holding the lock, which should prevent non-UI callbacks from accidentally blocking UI callbacks... though really, callbacks shouldn't take an excessive amount of time anyways.

Also, I did a small examination, and it appears most events are called from the UI thread anyways. But still, this is better I think.

PR #169, opened 2015-10-26
Main window: move ProgressManager into footer

genodeftest:
Move the ProgressManager into footer ("status bar") beside GtkStatusBar.

ProgressManager: Replace 2-line-design (grid) by one-line-design (box) to fit into footer

main window GtkPaned: force disable shrink of right and left child. Even after removing and re-adding left child. Note: this change effectively forces a minimum window size.

Playlists panel: force minimum size to match changes to main window GtkPaned.

### DIFF
--- a/data/ui/main.ui
+++ b/data/ui/main.ui
@@ -10,10 +10,9 @@
     <signal name="delete-event" handler="on_delete_event" after="yes" swapped="no"/>
     <signal name="window-state-event" handler="on_window_state_event" swapped="no"/>
     <child>
-      <object class="GtkBox" id="vbox1">
+      <object class="GtkGrid" id="grid1">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="orientation">vertical</property>
         <child>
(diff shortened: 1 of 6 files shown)


virtuald:
You can do a 1-line design with grid too. Isn't Box deprecated?

genodeftest:
According to https://developer.gnome.org/gtk3/unstable/GtkBox.html#gtk-box-new GtkBox is not deprecated. According to https://developer.gnome.org/gtk3/unstable/gtk-migrating-GtkGrid.html it might replace GtkBox once. But since most of Gtk's composite widgets use GtkBox, I doubt that will be soon. I used GtkBox because in GtkGrid you don't have a simple `pack_start()` used to insert new widgets (usually a "refresh" image as busy icon).
Should I try to use GtkGrid anyway?

virtuald:
Oh, I thought it was deprecated... I see! I didn't realize that Gtk.Box was different from Gtk.VBox/Gtk.HBox. This is fine then, I'll test tonight and merge.

virtuald:
I like it! Thanks!

PR #171, opened 2015-10-27
Gi

mexicarne:
These changes only adding python3 compatibility (not yet finished) and are no-op for python2.

### DIFF
--- a/xl/common.py
+++ b/xl/common.py
@@ -294,9 +294,9 @@ def profileit(func):
         stats = pstats.Stats(prof)
         stats.strip_dirs()
         stats.sort_stats('time', 'calls')
-        print ">>>---- Begin profiling print"
+        print(">>>---- Begin profiling print")
         stats.print_stats()
-        print ">>>---- End profiling print"
+        print(">>>---- End profiling print")
         prof = stats = None
         return res
     return wrapper
@@ -569,7 +569,7 @@ class MetadataList(object):
         self.__list = list(iterable)
         meta = list(metadata)
         if meta and len(meta) != len(self.__list):
-            raise ValueError, "Length of metadata must match length of items."
+            raise ValueError("Length of metadata must match length of items.")
         if not meta:
             meta = [None] * len(self.__list)
         self.metadata = meta
@@ -664,7 +664,7 @@ class MetadataList(object):
 
     def del_meta_key(self, index, key):
         if not self.metadata[index]:
-            raise KeyError, key
+            raise KeyError(key)
         del self.metadata[index][key]
         if not self.metadata[index]:
             self.metadata[index] = None
--- a/xl/common.py
+++ b/xl/common.py
@@ -41,11 +41,21 @@ import string
 import subprocess
 import sys
 import threading
-import urllib2
-import urlparse
 from functools import wraps, partial
 from collections import deque
 from UserDict import DictMixin
+
+try:
+    # python3 have not urllib2
+    import urllib.request as urllib2
+except ImportError:
+    import urllib2
+
+try:
+    # nor urlparse
+    import urllib.parse as urlparse
+except:
+    import urlparse
 
 logger = logging.getLogger(__name__)
 
--- a/xl/common.py
+++ b/xl/common.py
@@ -43,7 +43,11 @@ import sys
 import threading
 from functools import wraps, partial
 from collections import deque
-from UserDict import DictMixin
+
+try:
+    from UserDict import DictMixin
+except ImportError:
+    from collections import MutableMapping as DictMixin
 
 try:
     # python3 have not urllib2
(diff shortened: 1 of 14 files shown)


virtuald:
That's awesome! Just out of curiosity, why not use something like the six project to do this? Though, most of the changes seem to be pretty trivial.

mexicarne:
I'm not sure. Just wanna to avoid extra dependency.

virtuald:
That's fine, just curious. I'll try to test this out tonight.

xl/xdg.py
```diff
@@ -60,6 +60,8 @@
 if config_dirs is None:
     if sys.platform == 'win32':
         config_dirs = [exaile_dir]
+    elif 'freebsd' in sys.platform:
+        config_dirs = [data_dirs]
```
sjohannes: This doesn't look right; `data_dirs` is already a list. Also, data dir is in `$prefix/share`, whereas we want `$prefix/etc`. The latest version&mdash;which I've just pushed&mdash;should handle the FreeBSD case correctly (it searches `/usr/local/etc` in addition to `/etc`), so this change shouldn't be necessary.

xl/common.py
```diff
@@ -294,9 +308,9 @@ def wrapper(*args, **kwargs):
         stats = pstats.Stats(prof)
         stats.strip_dirs()
         stats.sort_stats('time', 'calls')
-        print ">>>---- Begin profiling print"
+        print(">>>---- Begin profiling print")
```
virtuald: For python2 compat, need `from __future__ import print_function`?

xl/main.py
```diff
@@ -582,7 +582,7 @@ def format_usage(self, usage):
 
     def version(self):
         from xl.version import __version__
-        print "Exaile", __version__
+        print("Exaile {}".format( __version__))
```
virtuald: `from __future__ import print_function`

xl/metadata/_matroska.py
```diff
@@ -355,7 +355,7 @@ def dump_tags(location):
     except KeyError:
         timecodescale = 1000000
     length = info['Duration'][0] * timecodescale / 1e9
-    print "Length = %s seconds" % length
+    print("Length = %s seconds" % length)
```
virtuald: `from __future__ import print_function`

xl/xldbus.py
```diff
@@ -111,9 +111,9 @@ def run_commands(options, iface):
         if getattr(options, command):
             value = iface.GetTrackAttr(attr)
             if value is None:
-                print _('Not playing.')
+                print(_('Not playing.'))
```
virtuald: `from __future__ import print_function`

virtuald:
Thanks for taking the initiative to do this, it's definitely welcomed, and certainly a good start. I think I would prefer taking the [six.py](https://bitbucket.org/gutworth/six/src/1991f8b5b654f077e773f05695a08e0506b7367f/six.py) library and drop it in `xl.externals`, where we keep our other small dependencies that are just easier to have in tree.

I suspect that this will make it easier to continue the porting effort, particularly for dealing with things like iteritems/items, range/xrange, etc.

xl/common.py
sjohannes: Not actually needed unless you are printing multiple things, e.g. `print(1, 2)` will print `(1, 2)` otherwise.

But in general I try to code with `from __future__ import division, print_function, unicode_literals`.

xl/settings.py
```diff
@@ -53,9 +62,11 @@
     'B': bool,
     'L': list,
     'D': dict,
-    'U': unicode
 }
 
+if PY2:
+    TYPE_MAPPING['U'] = unicode
```
virtuald: So... what happens if a user has a U setting, and migrates from py2 to py3? I suspect it will break?

xl/xdg.py
mexicarne: You are right, it should be 
```
config_dirs = data_dirs
```
but it works as is. FreeBSD port doesn't install anything to /usr/local/etc, everything is installed into /usr/local/share/exaile.

xl/common.py
mexicarne: It is only needed for pythons < 2.6 if I understand correctly. 
And yes, it is required to use .format() to print multiple values the same way in python3 and python2.

xl/xdg.py
sjohannes: I think it should be moved to `/usr/local/etc/xdg/exaile`, but if FreeBSD doesn't like that path then I guess we have no choice.

virtuald:
I'd like to get this merged, @mexicarne are you willing to fix the merge conflicts?

mexicarne:
Sure, will do at weekend. But doesn't the problems you mentioned are not the case anymore?

PR #172, opened 2015-10-28
Use gtktemplate

genodeftest:
Add GtkTemplate to About dialog, Collection manager and Device manager.

### DIFF
--- a/xlgui/collection.py
+++ b/xlgui/collection.py
@@ -35,58 +35,38 @@ from xl import (
     xdg
 )
 from xlgui.widgets import dialogs
+from xlgui.guiutil import GtkTemplate
 
 logger = logging.getLogger(__name__)
 
-class CollectionManagerDialog(object):
+@GtkTemplate('ui', 'collection_manager.ui')
+class CollectionManagerDialog(Gtk.Dialog):
     """
         Allows you to choose which directories are in your library
     """
+
+    __gtype_name__ = 'CollectionManager'
+
+    view, model, remove_button, content_area = GtkTemplate.Child.widgets(4)
+
     def __init__(self, parent, collection):
         """
             Initializes the dialog
         """
+        Gtk.Dialog.__init__(self)
+        self.init_template()
+
         self.parent = parent
         self.collection = collection
-        builder = Gtk.Builder()
-        builder.add_from_file(xdg.get_data_path(
-            'ui', 'collection_manager.ui'))
-        self.dialog = builder.get_object('CollectionManager')
-        self.dialog.set_transient_for(self.parent)
-        self.view = builder.get_object('view')
-        self.model = builder.get_object('model')
-        self.remove_button = builder.get_object('remove_button')
+
+        self.set_transient_for(self.parent)
         self.message = dialogs.MessageBar(
-            parent=builder.get_object('content_area'),
+            parent=self.content_area,
             buttons=Gtk.ButtonsType.CLOSE
         )
 
-        builder.connect_signals(self)
-
-        selection = self.view.get_selection()
-        selection.connect('changed', self.on_selection_changed)
-
         for location, library in collection.libraries.iteritems():
             self.model.append([location, library.monitored, library.startup_scan])
-
-    def run(self):
-        """
-            Runs the dialog, waiting for a response before any other gui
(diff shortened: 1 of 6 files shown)


xlgui/devices.py
```diff
@@ -33,62 +33,54 @@
...
+        self.btn_edit.destroy()
+        self.btn_remove.destroy()
 
         # object should really be devices.Device, but it doesnt work :/
         self.model = Gtk.ListStore(object, GdkPixbuf.Pixbuf, str, str)
-        self.tree = self.builder.get_object('tree_devices')
-        self.tree.set_model(self.model)
+        self.tree_devices.set_model(self.model)
 
         render = Gtk.CellRendererPixbuf()
```
virtuald: You could add all these cell renderers to the glade file.

virtuald:
I like the net loss of python code. If you could add the CellRenderers to the glade file, that'd be excellent.

xlgui/collection.py
```diff
@@ -35,33 +35,40 @@
...
+        
+        self.set_transient_for(self.parent)
         self.message = dialogs.MessageBar(
-            parent=builder.get_object('content_area'),
+            parent=self.content_area,
             buttons=Gtk.ButtonsType.CLOSE
         )
 
-        builder.connect_signals(self)
+        #builder.connect_signals(self)
```
virtuald: Could just remove this.

xlgui/collection.py
```diff
@@ -186,7 +181,7 @@ def on_force_rescan_button_clicked(self, widget):
         from xlgui import main
         main.mainwindow().controller.on_rescan_collection_forced()
         
-    
+    @GtkTemplate.Callback
```
virtuald: `GtkTemplateWarning: Signal 'on_selection_changed' was declared with @GtkTemplate.Callback but was not present in template`

genodeftest: Oh, sorry. I should just start exaile with `--debug`, otherwise I won't see these issues.

xlgui/devices.py
genodeftest: Without [stackoverflow](https://stackoverflow.com/questions/4307786/glade-treeview-liststore-data-not-visible-cant-imagine-why) I didn't realize how to add columns and renderers to a GtkListView.
It looks like I don't own any devices for testing this functionality. I do own some music player hardware but they don't seem to be talked to over MTP. (How) Can I add dummy devices to test?

virtuald: That would be a great testing tool.... though I don't actually like having the device manager, really, as Exaile has such poor device support I think I'd almost rather get rid of it totally. I'm not opinionated enough about it yet to actually do that, however.

However, it looks like enabling the CD plugin and inserting an audio disc will put something in the device manager. Most people tend to have CD's somewhere...

genodeftest: Ok, CD seems to work but my changes fail with it. Do you know which type I have to specify for "object" from `Gtk.ListStore(object, GdkPixbuf.Pixbuf, str, str)` in the glade file's GtkListStore? It could look like this:
```
<object class="GtkListStore" id="model">
    <columns>
      <!-- column-name column_device_ptr -->
      <column type="GObject"/>
      <!-- column-name column_device_icon -->
      <column type="GdkPixbuf"/>
      <!-- column-name column_device_name -->
      <column type="gchararray"/>
      <!-- column-name column_driver_name -->
      <column type="gchararray"/>
    </columns>
  </object>
```
I tried:
* GObject. Result: `Expected GObject.Object, but got cd.CDDevice` in `File "/usr/lib64/python2.7/site-packages/gi/overrides/GObject.py", line 287, in set_value`
* Device. Result: `Must be gobject.GType, not gobject.GType` in same file
* gpointer. Result: `ValueError: Pointer arguments are restricted to integers, capsules, and None. See: https://bugzilla.gnome.org/show_bug.cgi?id=683599`

virtuald: You want [PyObject](https://github.com/exaile/exaile/blob/gi/data/ui/coverchooser.ui#L8)

genodeftest:
Ok, I hope I got all your tips.

virtuald:
Awesome, thanks!

PR #175, opened 2015-10-30
Some minor fixes to the gi branch, mostly Gtk2 to Gtk3 API breaks

genodeftest:
Some minor fixes to the gi branch, mostly related to Gtk API breaks from Gtk2 to Gtk3. Fixes some plugins and warnings.


genodeftest:
Please don't merge yet, I've found some more things to change.

PR #179, opened 2015-11-09
Fix notify plugin

genodeftest:
Fix an issue preventing notify plugin to display notifications due to an API change


sjohannes:
Merged, thanks.

PR #184, opened 2015-11-27
Cleanup code to comply to PEP8

genodeftest:
I ran autopep8 on the whole exaile tree. Most changes are whitespace fixes, some imports are reordered and long lines broken.

I read through some of these changes and found no breaks. I ran exaile for some time and it still works fine.

Please note that this change is so big that it probably will break mergeability of every other branch out there, including #171 and virtuald/exaile/tree/progress-gen. This change can easily reproduced by running `autopep --recursive --in-place` so you may do it later.

I'd suggest running autopep8 on exaile's code on a regular basis or use an editor/IDE that does this. It might prevent some errors and is good for preparing for python 3 which has more strict syntax parsing.

### DIFF
--- a/data/migrations/migration_200907100931/__init__.py
+++ b/data/migrations/migration_200907100931/__init__.py
@@ -4,7 +4,8 @@ from xl import settings, common
 from xl.playlist import PlaylistManager, Playlist
 from ConfigParser import SafeConfigParser
 import urlparse
-import oldexailelib, olddb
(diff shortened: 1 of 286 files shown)


virtuald:
I'm not terribly opposed to this, but if we're going to do this let's do it directly before the beta release. I still have a few things I need to take care of, but I spent too much time working on the OSX installer so those things are still not done.

PR #185, opened 2015-11-27
retire bundled pylast

### LINKED ISSUE #183: Please update or unbundle pylast
Currently exaile is shipping an outdated copy of pylast in xl/externals/pylast.py. I don't know whether or not it works and am unable to test. Anyhow there are some minor changes and fixes in newer releases of pylast.
Some pitfalls (diff `xl/externals/pylast.py` vs. `__init__.py` from 1.4.2):
* pylast now depends on `six`
* `Artist.get_images()` is gone (wasn't used by exaile anyway)

other API changes:
* new functions:
  * `Artist.get_band_members()`
  * `Artist.get_correction()`
  * `Track.get_correction()`
  * `User.get_tagged_albums()`
  * `User.get_tagged_artists()`
  * `User.get_tagged_tracks()`
  * `cleanup_nodes()`
  * `_extract_artists()`
  * `_extract_albums()`
  * `_extract_tracks()`
* Functions with new optional parameters:
  * `User.get_recent_tracks()`
  * `_extract_element_tree()`

I guess this change should happen before publishing exaile 4.0 since shipping a 2 years old pylast.py is probably a bad idea.

I'd prefer removing pylast.py from xl.externals since it is available on most linux distributions as package anyway, it will be up-to-date without any work of exaile developers needed. And distro packagers usually don't like applications shipping 3rd party code due to their packaging strategy. I don't know the situation for Windows and Mac OS X though.

genodeftest:
Removes bundled pylast, leaving it to the user/distro to provide it.
Consult your distros package manager or run pip to install pylast.

Fixes #183


virtuald:
If @mathbr  and @sjohannes  don't object, this looks fine to me.

sjohannes:
Looks good. I don't know why we bundled pylast; maybe it wasn't commonly packaged at one point, or maybe we needed a specific version that OSes didn't have (yet), but I think it should be stable and popular enough right now.

eri-trabiccolo:
I think 'cause wasn't that stable in the past.
Also the first project was basically dismissed since 2011.
See:
https://bugs.launchpad.net/exaile/+bug/1320491

sjohannes:
I see, thanks. Looks like most distributions have switched to the new project; PyPI has v1.4.2, Debian has v1.0.0. Hopefully the API hasn't changed from 0.6. @genodeftest did you test that the plugin still works?

genodeftest:
I read the source code. The diff is quite small and doesn't contain much changes. Didn't test it though.

mbrodala:
It seems like `Track.unlove` was [added to pylast](https://github.com/pylast/pylast/commit/50bb692383cca794b28cbc78f6a271237a4056cf) in the meantime so I don't mind. :-)

PR #187, opened 2015-11-27
Disable tray icon on wayland

### LINKED ISSUE #181: [gi branch] Detect wayland and don't run a status icon ("Tray icon")
In Gtk+ 3, `GtkStatusIcon` is deprecated, since gtk devs think status icons should not be used any more.

On Wayland this leads to an issue when running exaile with gtk3's wayland backend (the default on Fedora 23 in Gnome+Wayland sessions): In case exaile is configured to minimize or close to its status icon, the window is gone and cannot be easily opened. Running `exaile --toggle-visible` doesn't help. Only killing exaile via SIGTERM is getting me out of this. There is an upstream bug report for that [at gnome](https://bugzilla.gnome.org/show_bug.cgi?id=758677).

As a workaround I suggest detecting wayland.
On x11, `Gdk.Display.get_default().get_name()` gives something like `:0`, or more generally, `:number`.
On wayland it gives `Wayland`.

3 places where this change should be done:
* If exaile has detected to be running on wayland it should disable the status icon. The `gui/use_tray` option should be ignored and assumed as `False`. Add some logging.
* If exaile is started with `--start-minimized` it should just open a window, but not request focus.
* In preferences/appearance the "use tray" option should be disabled (not only unchecked) and have a tooltip telling telling the user "Status icons are not available on Wayland.".

genodeftest:
This change does:
* prevent exaile from starting minimized to tray on Wayland
* disable the `gui/use_tray` preference when started on Wayland
* not interfere with exaile running on anything but Wayland except for the point above
* fix #181

What I noticed is that with and without this change applied the `--start-minimized` option is ignored and exaile will never start to tray. Can someone confirm that? Is this probably specific to gnome sessions? If yes, can someone please test that my change doesn't break the `--start-minimized` option on some other desktop?

### DIFF
--- a/xlgui/__init__.py
+++ b/xlgui/__init__.py
@@ -147,7 +147,11 @@ class Main(object):
             .child_set_property(self.panel_notebook, 'shrink', False)
 
         if settings.get_option('gui/use_tray', False):
-            self.tray_icon = tray.TrayIcon(self.main)
+            if tray.is_supported():
+                self.tray_icon = tray.TrayIcon(self.main)
+            else:
+                settings.set_option('gui/use_tray', False)
+                logger.warn("Tray icons are not supported on your platform. Disabling tray icon.")
             
         from xl import event
         event.add_ui_callback(self.add_device_panel, 'device_connected')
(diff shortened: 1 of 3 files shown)


virtuald:
On Fedora 22, `--start-minimized` seems to work just fine for me.

virtuald:
I closed the `gi` branch, but merged this branch into master.

Issue #156, opened 2015-10-19
[gi] [library] popup menu broken when multiple items selected


genodeftest:
A recent commit (not yet bisected) on the gi branch broke the popup menu when multiple items are selected in the library.

Now if you
1. select multiple items in library
2. right-click or press the context menu button on your keyboard
you will get a context menu for the panel tabs, not for the library panel content.

About 2 weeks ago the default popup menu still worked on selecting multiple items in library, but only if you used the context menu button on your keyboard, not with right-click on your mouse.

What I would like to see:
No matter whether you select one or more items in library panel you should always see the same popup menu.

virtuald:
I don't have this problem when right clicking, the appropriate menu always shows up. Which panel are you doing this in?

However, when using the menu key on the keyboard, it always shows the panel tabs context menu, which is incorrect.

genodeftest:
I am using the library panel.
Now I'm seeing the correct popup menu too if using the right mouse button.

genodeftest:
The change d74419d I commited fixes part of this issue: It fixes the right-click popup menu. It does not fix the menu key (keyboard key) though. And this code is too complex so I probably won't be able to find this issue.
A long-term fix of this code should remove all `event.button == <number>` code and replace it with constants from Gtk. Furthermore it should include the context menu key. To make this key work you need to connect to `popup` instead of `button-pressed` signal, which will break Shift+Click.

sjohannes:
I've reverted the patch, so this is broken again until we can find a better fix.

sbrubes:
This bugs started on **v3.2 to v3.3**, probably.
I have to check a code that I don't have access right now.

It affects me on Files Panel, and at that time I made a solution. Maybe not exactly this one and more tested.

By now it looks fine if you add the handler to _button_press_ and makes it stop to propagate, adding this method to _FilesPanel_. https://github.com/exaile/exaile/blob/master/xlgui/panel/files.py#L64

```
    def button_press(self, button, event):
        return True
```

Still needing to look if it works for _CollectionPanel_.

sbrubes:
https://github.com/exaile/exaile/pull/474 fix it.

genodeftest:
@sbrubes wrote:
> #474 fix it.

No, it does not. The steps to reproduce in https://github.com/exaile/exaile/issues/156#issuecomment-149591165 still apply. Also, the code change proposed in https://github.com/exaile/exaile/issues/156#issuecomment-329036339 does not help.

PR #189, opened 2015-11-27
Fix panel popup menu

genodeftest:
This is reported as bug #156:
You couldn't right-click open a popup menu on multiple files or items. Exaile always deselected all but one elements.

Cause: handlers for signal button-press-event must return true in some cases. See [1] for this tip.
The change in DragTreeView reflects the fact that users might move their mouse a bit between pressing and releasing their right mouse button. In this case selection should not break either.

This issue would not have happened if python were enforcing explicit returns instead of assuming that returning nothing is returning False. Blame you, python!

[1] http://mailman.daa.com.au/cgi-bin/pipermail/pygtk/2005-June/010465.html

### DIFF
--- a/xlgui/panel/collection.py
+++ b/xlgui/panel/collection.py
@@ -455,6 +455,7 @@ class CollectionPanel(panel.Panel):
                 return True
             else:
                 return False
+        return True
 
     def on_expanded(self, tree, iter, path):
         """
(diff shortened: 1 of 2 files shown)


sjohannes:
Unfortunately this breaks the selection behaviour while right-clicking on a non-selected row.

To reproduce: Activate the Files panel. Right-click on a row that is not selected.

Expected behaviour: The right-clicked row is selected. A popup menu is shown. Any action taken from the popup affects only the newly selected row.

Actual behaviour: The selection does not change. A popup menu is shown. Any action taken from the popup affects the old selected rows.

I think this is a more severe bug, so I'm reverting the patch.

PR #194, opened 2015-12-11
Require GIRepository

genodeftest:
Prevent warning when opening preferences page "Plugins" for the first time. Since all the other essential `gi.require_version()` calls are here, I put it here too, which might be wrong.


virtuald:
Seems fine to me.

PR #202, opened 2015-12-19
Minor fixes

genodeftest:
Just some minor fixes. Mostly to get less warnings.

### DIFF
--- a/plugins/currentsong/PLUGININFO
+++ b/plugins/currentsong/PLUGININFO
@@ -1,6 +1,6 @@
 Version='4.0.0'
 Authors=['Aly Hirani <alyhirani@gmail.com>', 'Jorge Villaseñor <salinasv@gmail.com']
-Name=_('Current Song')
+Name=_('Current Song in Pidgin')
 Description=_('Sets the currently playing status in Pidgin. Check the Pidgin FAQ to find out the suppported services.')
 Category=_('Notifications')
 RequiredModules=['dbus']
--- a/plugins/desktopcover/desktopcover_preferences.py
+++ b/plugins/desktopcover/desktopcover_preferences.py
@@ -21,7 +21,7 @@ from xl.nls import gettext as _
 name = _('Desktop Cover')
 basedir = os.path.dirname(os.path.realpath(__file__))
 ui = os.path.join(basedir, "desktopcover_preferences.ui")
-icon = 'desktop'
+icon = 'image-x-generic'
 
 class AnchorPreference(widgets.ComboPreference):
     default = 'topleft'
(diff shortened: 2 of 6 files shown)


xl/player/player.py
```diff
@@ -397,7 +397,8 @@ def is_playing(self):
             :returns: whether the player is currently playing
             :rtype: bool
         """
-        return self._engine.get_state() == 'playing'
+        return self._engine.get_state() is not None \
```
sjohannes: This doesn't look right to me. [`ExaileGstEngine.get_state`](https://github.com/exaile/exaile/blob/master/xl/player/gst/engine.py#L201) can only return `playing`, `paused`, or `stopped`.

genodeftest: Hm, yes, doesn't look right to me either. There is one case though where it returns `None` that I can reproduce:
1. start exaile with both "Preview Device" and "BPM" disabled
2. open preferences
3. enable "Preview Device"
4. disable "Preview Device"
5. enable "BPM"

Without this change you get a traceback:
```
11:27:25,996:DEBUG   : Provider BPM registered for service mainwindow-info-area-widget with target None (xl.providers)
11:27:25,998:ERROR   : Event callback exception caught! (xl.event)
  File "/path/to/exaile/xl/event.py", line 345, in _emit
    event.data, *cb.args, **cb.kwargs)
  File "/path/to/exaile/xl/providers.py", line 211, in _add_callback
    self.on_provider_added(provider)
  File "/path/to/exaile/xlgui/main.py", line 1129, in on_provider_added
    widget = provider.create_widget(self)
  File "/path/to/exaile/plugins/bpm/__init__.py", line 90, in create_widget
    return BPMWidget(info_area.get_player(), self)
  File "/path/to/exaile/plugins/bpm/__init__.py", line 169, in __init__
    if self.player.is_playing():
  File "/path/to/exaile/xl/player/player.py", line 400, in is_playing
    return self._engine.get_state() == 'playing'
AttributeError: 'NoneType' object has no attribute 'get_state'


Locals at innermost frame:

{ 'self': <xl.player.player.ExailePlayer object at 0x7fc8e2e5eb50>}


11:27:26,  0:DEBUG   : Provider _bpm registered for service track-panel-menu with target None (xl.providers)
11:27:26,  1:DEBUG   : Provider _bpm registered for service playlist-context-menu with target None (xl.providers)
11:27:26,  1:DEBUG   : Loaded plugin bpm (xl.plugins)
```
With this change the traceback is **NOT** gone (and it can't be). I'll revert it.

sjohannes: > `'NoneType' object has no attribute 'get_state'`

Oh, that means it's `self._engine` that is None at that point. You can check that instead.

genodeftest: > Oh, that means it's `self._engine` that is None at that point.

Yes.

I'm not quite sure whether it is allowed that `self._engine` is `None`.  This might also be a bug in PreviewDevice plugin. What do you think?

sjohannes: Hm, I'm not sure where the source of the bug is, but there does seem to be problems in the interaction between those two plugins. The `self` there is the Preview Device's player object (you can check by printing `self._name`)&mdash;which should have disappeared at that point.

PR #207, opened 2015-12-19
Simplify progress thread usage and make it threadsafe

virtuald:
This is an idea for simplfying the progress monitor stuff. It's not particularly well tested yet, but I think it's sound. I had started migrating the collection progress monitor to it, but got distracted so it's not in here. Would be good to integrate this with my gio library, perhaps?


sjohannes:
I haven't tested this, but the code looks ok other than the few clarity recommendations.

PR #227, opened 2016-02-20
added plugin: shufflecollection

rain0r:
Added a plugin which shuffles through the whole collection. It add a song from the collection dynamically to the current playlist. Code style might be ugly but it's working.

### DIFF
--- a/plugins/shufflecollection/__init__.py
+++ b/plugins/shufflecollection/__init__.py
@@ -1,0 +1,163 @@
+#!/usr/bin/python
+
+# Copyright (C) 2012 Rainer Hihn ( rainer@hihn.org )
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


plugins/shufflecollection/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+    settings,
+    trax
+)
+from xl.player.adapters import (
+    PlaybackAdapter,
+    QueueAdapter
+)
+from xlgui import main
+from xlgui.widgets import menu
+import glib, gtk, logging, os, random, time, thread
```
virtuald: Have you tested against master? We're using GTK 3.x and PyGObject 3.x now.

rain0r:
Yeah, unfortunately it doesn't work with 4.0.0-dev but with 3.4.5 though.

virtuald:
We won't be able to include this in master if it doesn't work in master.

virtuald:
Closing this PR. If you port to Exaile 4.0, feel free to resubmit it!

PR #230, opened 2016-02-22
SeekProgressBar: enable seek on scroll

genodeftest:
This change adds a feature also known from e.g. VLC media player:
When scrolling on SeekProgressBar, progress is being changed adding/removing some seconds.

There is no fixed reason why I chose 5% and 0.5%, you may change that.
I am unable to test the "smooth scroll" part, but the rest is working fine. No problems when not playing songs. Press your scroll wheel left/right to "scroll" left/right.


virtuald:
Sorry for taking so long to get to this, works for me.

Issue #212, opened 2015-12-27
Please store size and position of the 'Track Properties' dialog


strahlc:
Exaile has great possibilities to edit metadata, but changing the window size and position each time the properties dialog shows up is annoying.

PR #233, opened 2016-03-05
Fixes #209 and #212

### LINKED ISSUE #209: The 'Properties' dialog doesn't show up when the ID3 tag has a string in the BPM field (Exaile 3.4.5)
If a mp3 File has a string value (like 'abc') in the ID3 BPM field, the properties dialog cannot be opened.

strahlc:
…has a string in the BPM field)


virtuald:
This seems fine to me, I haven't tested it yet.

strahlc:
split pull request into #235 and #236

PR #235, opened 2016-03-06
Fixes #209 (The 'Properties' dialog doesn't show up when the ID3 tag …

### LINKED ISSUE #209: The 'Properties' dialog doesn't show up when the ID3 tag has a string in the BPM field (Exaile 3.4.5)

strahlc:
…has a string in the BPM field)

### DIFF
--- a/xlgui/properties.py
+++ b/xlgui/properties.py
@@ -30,6 +30,7 @@ import datetime
 import io
 import os
 import string
+import re
 
 from gi.repository import Gdk
 from gi.repository import GdkPixbuf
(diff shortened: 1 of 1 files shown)


strahlc:
You cant test this PR by setting a string to the BPM field:

id3v2 --TBPM "123.45 BPM" "05. Artist - Title 5.mp3"

without this PR you get a traceback

virtuald:
I'll try it out tonight, thanks!

Issue #226, opened 2016-02-07
'Date added' is set to 'Never' for new tracks when copying new tracks in 'Monitored' collection folder


strahlc:
I have a lot of tracks where the "Date added" column is set to "Never".
Where does this come from and how can I avoid that.
(I usually add Tracks by moving them in the montored folders and rescan the collection.)

genodeftest:
For a reason why this happens, have a look at xl/formatter.py from line 685 on (or search for `DateTagFormatter`). This class `DateTagFormatter` was intended to be used for the "Last played" date (column) but has also been applied to the "Date added" date (column). In your case ("Date added" shows "Never"), the timestamp for the file is `None`. This happens when a file is not yet added to your collection.

Steps to reproduce the behavior you described:
1. add a file which is not in exaile's collection to a playlist
What happens:
Column "Date added" shows "Never"

Expected behavior (1):
1. add a file which is not in exaile's collection to a playlist
2. add folder containing this file to collection
What happens:
Column "Date added" changes to "Today"

Expected behavior (2):
1. add a folder to exaile's collection
2. add any file from exaile's collection to a playlist
What happens:
Column "Date added" shows "Today" or any previous date.

What could be done is improving column header text, e.g. from "Date added" to "Added to library". I'll file a pull request for that.

So there is nothing wrong with how the date is shown. Can you please make sure that these files really are in your collection? Use the collection panel and search for these files. If this file is not in your collection, this is not a bug. Maybe something went wrong adding these files to exaile's collection?

strahlc:
Only about 25% of the file **in my collection** has the 'Date added' set to a date. All other files **in my collection** are set to 'Never'.
For example: I have an album in my collection where some tracks has the date added set correct while other tracks of the same album are set to never.

Expected behavior:
1. rescan collection
2. all files found by exaile in 'Monitored' folders should get a time stamp with the current time for date added, if they do not already have one.

genodeftest:
Is there anything about these files in your exaile log files? Probably something went wrong monitoring/rescanning your collection. I still cannot reproduce it. Is anything special about these files like are they on a removable disk that might have been unavailable at least once when exaile started or was running?

Have you tried deleting the collection from exaile and adding it again?

strahlc:
When I copy the album to a folder outside my collection and add add this folder in 'Collection Manager', the album appears twice and the newer one has all tracks with 'Date added' set to 'Today'.

genodeftest:
[...]

PR #238, opened 2016-03-10
Clarify meaning of `__date_added` field

genodeftest:
When a track has not been added to exaile's collection, it will show "Never" in the "Date added" column. It is unclear where the track has (never) been added to.

See https://github.com/exaile/exaile/issues/226#issuecomment-195025945 for more details.

### DIFF
--- a/xl/metadata/tags.py
+++ b/xl/metadata/tags.py
@@ -70,7 +70,7 @@ tag_data = {
     
     '__bitrate':        _TD(N_('Bitrate'),      'bitrate', editable=False),
     '__basedir':        None,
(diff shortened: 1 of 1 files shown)


virtuald:
I like the idea, but "Added to library" sounds like a True/False to me, as opposed to a timestamp.

genodeftest:
I guess your English is probably better than mine ;) In German this just makes perfect sense…
Any suggestions? "Date added to library" is probably too long.

virtuald:
Let me contemplate it, I agree there's not a good alternative yet.

genodeftest:
Ok, so this change is no good use.

PR #239, opened 2016-03-10
Don't print "Unknown" in "Grouping" column if no group is set

### LINKED ISSUE #225: Please leave the "Grouping" column blank if no grouping is set
Right now exaile displays "Unknown" in the "Grouping" column if no grouping is set.
This does not look good because if a grouping is present, the text is usually shorter.

genodeftest:
Fixes #225

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -633,6 +633,8 @@ class Track(object):
             if tag == 'title':
                 basename = self.get_basename_display()
                 value = u"%s (%s)" % (value, basename)
+            elif tag == 'grouping':
+                value = ''
 
         # Convert value to unicode or List[unicode]
         if isinstance(value, list):


strahlc:
With this PR the Grouping column is always empty, even the track has a grouping set.

genodeftest:
Yeah, wrong time to do this. Let's try again tomorrow.

strahlc:
Put it only a few line lower and it solves the problem :grin:

```
diff --git a/xl/trax/track.py b/xl/trax/track.py
index 408d5ce..6be2bd9 100644
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -633,6 +633,8 @@ class Track(object):
             if tag == 'title':
                 basename = self.get_basename_display()
                 value = u"%s (%s)" % (value, basename)
+            elif tag == 'grouping':
+                value = ''
 
         # Convert value to unicode or List[unicode]
         if isinstance(value, list):
```

genodeftest:
Thanks!

sjohannes:
I don't really know how Grouping is normally used for so I'll leave this one to Dustin.

As a general comment, though, I wonder what others think about all the "Unknown" labels for unset/empty tags. They've been around forever, but personally I think I prefer them blank for reduced clutter.

strahlc:
I prefer them blank too.

virtuald:
I'm fine with changing all unused tags to be blank. Change this PR to do that?

strahlc:
My proposal:
```
diff --git a/xl/trax/track.py b/xl/trax/track.py
index 408d5ce..31cf04e 100644
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -629,10 +629,10 @@ class Track(object):
             value = self.__tags.get(tag)
 
         if value is None:
-            value = _UNKNOWNSTR
+            value = ''
             if tag == 'title':
                 basename = self.get_basename_display()
-                value = u"%s (%s)" % (value, basename)
+                value = u"%s (%s)" % (_UNKNOWNSTR, basename)
 
         # Convert value to unicode or List[unicode]
         if isinstance(value, list):
```

What do you think?
Or should we use basename only, when the title tag is missing?

virtuald:
@strahlc I like that proposal

genodeftest:
Closing in favor of #250

PR #240, opened 2016-03-11
Fix traceback when saving cover to metadata

### LINKED ISSUE #217: Traceback when trying to add cover art to mp3 files
If I try to add a JPEG file as album art to a MP3 file (in the Properties dialog), I got the following error

  File "/lib64/exaile/xlgui/properties.py", line 1080, in on_button_clicked
    self.set_pixbuf(pixbuf, info['mime_types'][0])
TypeError: 'PixbufFormat' object has no attribute '__getitem__'

and the file chooser dialog stays open until I press ESC.

Affects current master (8655a17)

genodeftest:
see https://github.com/mypaint/mypaint/issues/236
and https://github.com/mypaint/mypaint/commit/d12f42585d04758d90163789553b05f2866a75e7

This should fix #217, fix confirmed working by @strahlc. Please someone have a look over it anyway, especially on a non-Fedora/non-Gtk+ 3.18 system.

### DIFF
--- a/xlgui/properties.py
+++ b/xlgui/properties.py
@@ -1047,7 +1047,22 @@ class TagImageField(Gtk.HBox):
         mime = self.mime_info[self.info.mime]
         # Retrieve proper image data
         writer = io.BytesIO()
-        self.pixbuf.save_to_callback(writer.write, mime['type'], mime['options'])
+
+        def gdk_pixbuf_save_func(buf, count, user_data):
+            if writer.write(buf) == count:
+                return True
+            return False
+
+        # workaround for some undocumented changes in GdkPixbuf API
+        # see https://bugzilla.gnome.org/show_bug.cgi?id=670372
+        # see https://github.com/mypaint/mypaint/issues/236
+        try:
+            save_to_callback_function = self.pixbuf.save_to_callbackv
+        except AttributeError:
+            save_to_callback_function = self.pixbuf.save_to_callback
+        save_to_callback_function(gdk_pixbuf_save_func, None, mime['type'], \
+                mime['options'].keys(), mime['options'].values())
+
         # Move to the beginning of the buffer to allow read operations
         writer.seek(0)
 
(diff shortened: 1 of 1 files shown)


sjohannes:
If I understand correctly from the GTK+ [bug report](https://bugzilla.gnome.org/show_bug.cgi?id=670372) and [changeset](https://git.gnome.org/browse/gdk-pixbuf/commit/?id=112eab4), the various `save` methods should be identical to their `v` counterparts. That is, both should be used with the same args, e.g.

```py
save_to_callback = getattr(self.pixbuf, 'save_to_callbackv', None) or self.pixbuf.save_to_callback
save_to_callback(save_func, None, ...)
```

(checking for `save_to_callbackv` first because presumably that's the one that will be available going forward).

genodeftest:
Sadly that's not true in my case. `save_to_callback` is broken. I'm getting `AttributeError: 'Pixbuf' object has no attribute 'save_to_callback'`.

I don't understand your code or what you want to tell me with that.

sjohannes:
I meant that (assuming I read the GTK+ bug report correctly) `save_to_callback`, before it was removed, used to be the same as `save_to_callbackv`, so

```py
try:
    self.pixbuf.save_to_callback(writer.write, mime['type'], mime['options'])
except AttributeError:
    self.pixbuf.save_to_callbackv(gdk_pixbuf_save_func, None, mime['type'], \
        mime['options'].keys(), mime['options'].values())
```

isn't right; both of them should be called with the second set of args (func, data, type, keys, values). The original set of args (func, type, options_dict) came from pygtk but probably never worked in pygi.

So my recommendation is to replace that block with

```py
try:
    save_to_callback = self.pixbuf.save_to_callbackv
except AttributeError:
    save_to_callback = self.pixbuf.save_to_callback

save_to_callback(gdk_pixbuf_save_func, None, mime['type'],
    mime['options'].keys(), mime['options'].values())
```

(this is equivalent to the suggestion in my previous comment but probably more readable).

genodeftest:
Better?

sjohannes:
Yup, thanks.

PR #241, opened 2016-03-11
Some more Gtk+ 3.x work

genodeftest:
Replace Gtk{H,V}{Box,ButtonBox,Paned} with Gtk{Box,ButtonBox,Paned}.
Port track info dialog from GtkTable to GtkGrid.
Port some classes to GtkTemplate.

No major regressions expected.

### DIFF
--- a/xlgui/properties.py
+++ b/xlgui/properties.py
@@ -51,6 +51,7 @@ from xl import (
 )
 
 from xlgui.widgets import dialogs
+from xlgui.guiutil import GtkTemplate
 from xl.metadata.tags import tag_data, get_default_tagdata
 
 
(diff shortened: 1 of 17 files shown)


xlgui/properties.py
```diff
@@ -80,10 +81,10 @@ def __init__(self, parent, tracks, current_position=0, with_extras=False):
...
         self.__changed_attributes = Pango.AttrList()
 
-        self.message = dialogs.MessageBar(
-            parent=self.builder.get_object('main_container'),
-            buttons=Gtk.ButtonsType.CLOSE
-        )
+        main_container = self.builder.get_object('main_container')
+        self.message = dialogs.MessageBar(buttons=Gtk.ButtonsType.CLOSE)
+        main_container.insert_row(0)
+        main_container.attach(self.message, 0, 0, 6, 1)
```
sjohannes: Could you keep the old layout, with `main_container` being a VBox? That way if we add new buttons we don't need to update this.

genodeftest: Are you sure about that? Gtk+ people have deprecated Gtk.VBox (so you probably meant a Gtk.Box with Gtk.Orientation.VERTICAL). But they also told to prefer GtkGrid over GtkBox.

But yes, there is a point. Having code easy for upcoming changes would probably be preferred…

sjohannes: Yeah, I meant a vertical GtkBox.

Regarding Box vs. Grid: We can of course use single-row/single-column Grids (and the docs say they "provide exactly the same functionality"), but I think the Box API is still better for those cases, though I don't know if it's going to be deprecated any time.

genodeftest: Yeah, pack_start() and pack_end() is something that's definitely missing in GtkGrid.

PR #242, opened 2016-03-11
[Plugin: podcasts] Unbundle outdated feedparser module

genodeftest:
Feedparser has seen some updates but ist 99% API stable.[1]
It is available with most Linux Distributions (debian, arch, fedora, ubuntu) and via pypi.

Version 4.1 shipped with exaile is at least 5 years old.
Therefore I suggest unbundling feedparser. If you don't want to do that, please update feedparser.

[1] https://pypi.python.org/pypi/feedparser

I run into some non-critical tracebacks with the bundled version of feedparser. Some of these tracebacks are gone with this change. See also plugins wiki page. This update doesn't fix all of those, but it doesn't get worse at least.

The podcasts plugin references feedparser only twice so there is not much risk to break stuff on both unbundling and updating it.


genodeftest:
rebased after #243

PR #243, opened 2016-03-12
Make "print", "except" and "raise" compatible with Python3

nsuke:


sjohannes:
Thanks.

By the way, there are also a few changes at [my branch](https://github.com/exaile/exaile/compare/master...sjohannes:py3) that makes Exaile run on Python 3, but the patch is dirty (has a bunch of hacks and unrelated changes). I'm not really keen on merging most of those until we fully commit to Python 3, as they would clutter the code too much with compatibility hacks.

But the changes that you made here are fine, as they are just modernizing the code.

PR #245, opened 2016-03-12
Initial Python 3 support

nsuke:
@sjohannes hi, merging your branch, it's now somewhat working on Python 3 for me.
Skipping to next track does not work for some reason on Py3, but hopefully it shouldn't break anything on Py2.
Test cases work fine for both 2 and 3.

---

This change is [Reviewable](https://reviewable.io/reviews/exaile/exaile/245)

### DIFF
--- a/plugins/cd/__init__.py
+++ b/plugins/cd/__init__.py
@@ -171,9 +171,8 @@ class CDPlaylist(playlist.Playlist):
             songs[song.get_loc_for_io()] = song
 
         # FIXME: this can probably be cleaner
-        sort_tups = [ (int(s.get_tag_raw('tracknumber')[0]),s) \
-                for s in songs.values() ]
-        sort_tups.sort()
+        sort_tups = sorted([ (int(s.get_tag_raw('tracknumber')[0]),s) \
+                for s in songs.values() ])
         sorted = [ s[1] for s in sort_tups ]
 
         self.extend(sorted)
--- a/plugins/cd/__init__.py
+++ b/plugins/cd/__init__.py
@@ -25,6 +25,7 @@
 # from your version.
 
 
+from six.moves import range
 from xl.nls import gettext as _
 from xl import providers, event
 from xl.hal import Handler, UDisksProvider
--- a/plugins/cd/__init__.py
+++ b/plugins/cd/__init__.py
@@ -30,21 +30,25 @@ from xl import providers, event
 from xl.hal import Handler, UDisksProvider
 from xl.devices import Device, KeyedDevice
 import logging
-logger = logging.getLogger(__name__)
 
-import dbus, threading, os, struct
+import dbus
+import os
+import struct
 from fcntl import ioctl
 from xl import playlist, trax, common
-from xl import settings
 import os.path
 
 try:
-    import DiscID, CDDB
-    CDDB_AVAIL=True
+    import DiscID
+    import CDDB
+    CDDB_AVAIL = True
 except:
-    CDDB_AVAIL=False
+    CDDB_AVAIL = False
 
 import cdprefs
+
+logger = logging.getLogger(__name__)
+
 
 def get_preferences_pane():
     return cdprefs
@@ -57,6 +61,7 @@ CDROMREADTOCENTRY = 0x5306
 CDROM_LEADOUT = 0xAA
 CDROM_MSF = 0x02
 CDROM_DATA_TRACK = 0x04
+
 
 class CdPlugin(object):
     
--- a/DEPS
+++ b/DEPS
@@ -20,6 +20,7 @@ Essential:
 * pygobject (>= 3.13.2)
 * python-cairo
 * GIR libraries for GTK+/GStreamer
+* six
 
 Translation:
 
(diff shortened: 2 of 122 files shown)


genodeftest:
You probably need to rebase due to my pull request #242.

I've set up a tracker issue at #246 including a list of compatibilities for dependencies of plugins on https://github.com/exaile/exaile/wiki/Python3-dependencies-migration-tasklist

plugins/cd/__init__.py
```diff
@@ -25,27 +25,32 @@
 # from your version.
 
 
+from six import range
```
genodeftest: Even with python3-six installed this breaks on python3 with a traceback:

```
ERROR   : Unable to enable plugin cd
  File "/path/to/exaile/xl/plugins.py", line 145, in enable_plugin
    plugin = self.load_plugin(pluginname)
  File "/path/to/exaile/xl/plugins.py", line 84, in load_plugin
    plugin = imp.load_source(pluginname, os.path.join(path,'__init__.py'))
  File "/usr/lib64/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "<frozen importlib._bootstrap>", line 1220, in load
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/path/to/exaile/plugins/cd/__init__.py", line 28, in <module>
    from six import range
ImportError: cannot import name 'range'
```

nsuke: It should have been `from six.moves`. I'll double check other six imports too.
Is there recommended way to cover changes in plugin dirs ?

genodeftest: > Is there recommended way to cover changes in plugin dirs ?

I don't understand what you mean. Can you please explain more detailled or give an example?

nsuke:
I mean, how can I verify my changes all over the plugin directories efficiently ?
Running nosetests on root dir (rather than ./tests) seems to report some import errors like this, though.

genodeftest:
Oh, I don't know any better way then just running exaile and enabling all the plugins. And even that won't catch all of these issues since some imports are done inside of functions.

I've read of two tools which could possibly be useful, but I haven't tested those:
1. python module `modulefinder`
2. https://pythonhosted.org/Sumatra/reference/dependency_finder.html

nsuke:
@genodeftest thanks, I'll check these out !

nsuke:
@genodeftest hi, I've rebased and fixed the import issue.
I tried quickly automating the plugins load test but it turned out that main `exaile` and GUI dependencies are a bit hard to workaround.
I've just enabled them once from GUI as you suggested.

virtuald:
@nsuke I haven't had time to test this, but at first glance seems good to me. As you've noticed, our test coverage is not very good at all. I'll try to run this tonight and see if it works for me.

Can you add a note to the dependencies that 'six' is required, as that wasn't a prior dependency?

nsuke:
@virtuald updated DEPS and rebased.
I also found a missing \__next__ method in the original patch.
Adding [this](https://github.com/exaile/exaile/pull/245/commits/b04f25b90c96d305aeceec0bad1f762cfcada684#diff-df43d474b6c176b121650c22cfae129fR142) seems to have fixed the next track problem for py3.

sjohannes:
Thanks for this. I've cherry-picked some of the safe ones that don't require compatibility hacks. If everything goes well, next time you rebase a few of the commits should disappear from this pull request.

I've also rebased my branch in order to clean it up a bit. Some of the changes in my branch are actually pretty horrible; I wanted to do a complete review of it at some point.

Regarding the rest of your changes, I haven't had time to carefully look at them all, but I wanted to comment on a few things:

> ```diff
> -    def next(self, autoplay=True, track=None):
> +    def __next__(self, autoplay=True, track=None):
> ```

(in xl/player/queue.py) This looks incorrect. "Next" here refers to "play the next song" (compare with `prev()`); it should stay as `next()`. Similarly, the `next(x)` calls that refer to it should stay as `x.next()`.

> `to_unicode`

I don't recommend using this too much; it's mostly a hack for when you just want a Unicode object and don't really know or care what the input is (it works with any object). In most cases it's probably better to limit the input to either just a `bytes` or just a `six.text_type` and use `encode`/`decode` accordingly.

nsuke:
@sjohannes thanks, I've rebased it and reverted the unnecessary \__next__ changes you mensioned.
This time I've cherry-picked only mandatory changes from your branch rather than merging.
N.B. I excluded encoding changes in playlist.py that does not work with existing data on Python2.

As to the `to_unicode()`, it's basically just plain `unicode()` except that it works even with missing/redundant `encoding` argument.
This kind of encoding handling is unavoidable to some extent when we want py2 and py3 at once.
So while it might be undesirable hack for py2 alone, in this py2/py3 context I think it is a reasonable compatibility layer.

That said, I've put utf8 related commits to the last so that we can easily leave that part for now.

virtuald:
Closing this as it's too old to be useful now I think. I think we do want to support Python 3 in the nearish future though, as Python 2 is nearing EOL.

PR #254, opened 2016-03-28
Issue #58, sort by album artist

tfree87:
It seems that these adjustments will allow for sorting by album artist and fix the issue of searching, sorting, and displaying tracks in the collection panel. I have observed the following:

- If there is an 'albumartistsort' tag, then that is used for the searching, sorting, and display in the collection panel.
- If no album artist is given, then either 'Various Artists' will be used when it is a compilation or 'Unknown' will be used.
- Tracks will be sorted by 'albumartistsort' (if given) or album artist followed by 'Various Artists' and 'Unknown' which are placed at the end of the list.


virtuald:
Looks like this fixes a lot more than issue #58? Perhaps try doing it on a separate branch?

virtuald:
And it looks like you do have some good bugfixes in there, I just don't want to bring them all in a single PR.

tfree87:
Oh, when I made the pull request I thought it would only include the commits up to 6c5e4f7. I should not have merged my 'python3' branch with my master branch.

virtuald:
Nope, it includes the whole history too. If you can isolate these different sets of changes that would be great.

You may be interested in #245 for python3 support, as I'd like to keep python2 support also for the moment and I don't believe 2to3 (as you've used here) accomodates that.

tfree87:
I executed an interactive git rebase in order to remove the extra commits that were unnecessary and only include those that were involved in the album artist sorting. These changes are in the master branch of my fork.

I will check out #245. I was having fun playing around with the code in python3 and was using that branch as my 'playground' but accidentally merged it with my master branch. Not exactly sure how that happened.

PR #256, opened 2016-03-31
Add support for sorting by Album Artist, Issue #58

tfree87:
Sorting by album artist works with proper sorting in the collection panel where 'Various Artists' and 'Unkown Artists' are pushed to the bottom of the tree.


xl/trax/track.py
```diff
@@ -660,12 +668,12 @@ def get_tag_search(self, tag, format=True, artist_compilations=False,
         """
         extraformat = ""
         if tag == "albumartist":
-            if artist_compilations and self.__tags.get('__compilation'):
-                value = self.__tags.get('albumartist', None)
-                tag = 'albumartist'
-                extraformat += " ! __compilation==__null__"
```
virtuald: What happened to the extraformat? No longer necessary?

xl/trax/track.py
```diff
@@ -660,12 +668,12 @@ def get_tag_search(self, tag, format=True, artist_compilations=False,
...
+            if self.__tags.get('albumartistsort'):
+                value = self.__tags.get('albumartistsort')
+            elif artist_compilations and self.__tags.get('__compilation'):
+                value = self.__tags.get('albumartist', _VARIOUSARTISTSSTR)
             else:
-                value = self.__tags.get('artist')
+                value = self.__tags.get('albumartist', _UNKNOWNSTR)
```
virtuald: So if you're asking to search by albumartist, and there is no tag, it uses unknown instead of using the artist?

tfree87: Precisely, any album that lacks and album artist becomes 'Unknown'.

tfree87: Unless it is a compliation, then it automatically assigns it to 'Various Artitsts'

xl/trax/track.py
tfree87: It appears to me that the line is unnecessary. Adding the extraformat to the Albumartist tag simply adds ' ! __compilation==__null__' to the end of the albumartist tag. This does not seem to affect sorting or display but seems like it was used as a way to replicate the function of self.__tags.get('__compilation')

virtuald:
@tfree87 I know we've neglected this for awhile, but I'd like to get this taken care of. Can you rebase this PR and I will review it and actually get it in this time. Thanks!

Issue #176, opened 2015-10-30
Ignore diacritics for search


SaiyanRiku:
Exaile should ignore diacritics and accent when searching music on the library.

For example:
- searching "helene" should return result containing "Hélène".
- searching "coeur" should return result containing "cœur"

To help, I already did a function for one of my program in C with gtk to handle this. I hope this can work be adapted with python.

    gchar*
    g_utf8_removediacritics(const gchar *str, gssize len)
    {
        gchar *szNormalizedString;
        GString* szStringBuilder;
        gchar *szRes = NULL;
        gunichar c;
        gchar *szPtr = NULL;

        if(str != NULL){
            szNormalizedString = g_utf8_normalize (str, len, G_NORMALIZE_NFD);

            szStringBuilder = g_string_new ("");

            szPtr = szNormalizedString;
            while(szPtr){
                c = g_utf8_get_char(szPtr);
                if(c != '\0'){
                    if (!g_unichar_ismark(c)){
                        g_string_append_unichar (szStringBuilder, c);
                    }
                    szPtr = g_utf8_next_char (szPtr);
                }else{
                    szPtr = NULL;
                }
            }

            szRes = g_string_free (szStringBuilder, FALSE);
            g_free(szNormalizedString);
        }

        return szRes;
    }

Thank you in advance

genodeftest:
Python provides a function for that: `unicodedata.normalize()`

genodeftest:
https://stackoverflow.com/questions/517923/what-is-the-best-way-to-remove-accents-in-a-python-unicode-string and https://stackoverflow.com/questions/9042515/normalizing-unicode-text-to-filenames-etc-in-python show some possible solutions to this problem.

PR #261, opened 2016-05-03
This fixes exaile/exaile#176: Ignore diacritics for search

tfree87:
I have used the unidecode library to add the ability to ignore diacritics while searching. If the unidecode library is present, then typing in something like "motley crue" or "mötley crüe" will both match "Mötley Crüe". This was done by converting both the `search_string` and the tag values into ascii text when a search was conducted. If the user does not have the unidecode library, then the default behavior is restored where diacritics are not ignored during search. I have tested this on my system and it works for searching when unicode and non-unicode strings are utilized for searching tracks.

### DIFF
--- a/xl/trax/search.py
+++ b/xl/trax/search.py
@@ -26,6 +26,16 @@
 
 import time
 import re
+import logging
+try:
+    from unidecode import unidecode
+except:
+    error('Could not import unidecode module, searches will not ignore diacritics')
+    UNIDECODE_IMPORTED = False
+else:
+    logging.info("Unidecode imported")
+    UNIDECODE_IMPORTED = True
+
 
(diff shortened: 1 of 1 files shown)


sjohannes:
Thanks; this looks safe for 4.0, especially as it's optional. However, have you tried the suggestions in #176 of using `g_utf8_normalize` or `unicodedata.normalize`? If we could use either of them we'd avoid the extra dependency.

tfree87:
If found this solution from j4mie/normalise.py. This solution used the unicodedata.normalize function from the Python standard library. Under my tests this worked when using a unicode string to search and when using a. This solution performs the same function as before where unicode search strings are converted to ascii if they are unicode and all content from music tags are converted to ascii as well when compared.

Issue #253, opened 2016-03-27
Exportet playlist conatins wrong path when the playlist name contains withespaces


strahlc:
When exporting a playlist with one or more whitespaces in the name, the playlist entries pointing to an invalid location.

Steps to reproduce:
1. create a playlist "my cool playlist"
2. add one or more titles
3. export the playlist to m3u
4. check the file contend -> result: The path of each entry in the exported playlist points to an invalid location
5. rename the playlist to "my_cool_plalist" and export it again
6. check the file contend -> result: The path of each entry in the exported playlist is correct

Example:
With white spaces in playlist name:
```
$ cat "my cool playlist.m3u" 
#EXTM3U
#PLAYLIST: my cool playlist
#EXTINF:2,Track 6 - Artist
../test_data/artist/album/06. Artist - Track 6.mp3
#EXTINF:2,Track 7 - Artist
../test_data/artist/album/07. Artist - Track 7.mp3
```
After rename and export:
```
$ cat "my_cool_playlist.m3u" 
#EXTM3U
#PLAYLIST: my_cool_playlist
#EXTINF:2,Track 6 - Artist
test_data/artist/album/06. Artist - Track 6.mp3
#EXTINF:2,Track 7 - Artist
test_data/artist/album/07. Artist - Track 7.mp3
```

tfree87:
It seems that there was an error in getting the basename of the exported playlist file. The `playlist_file.get_basename()` was seems to only get part of the filename if there is a space in the file name. Using `os.path.basename(playlist_path)` seems to fix this issue.

tfree87:
Not sure why `Gio.File.get_basename()` would be broken in this way. This will require more investigation...

tfree87:
I figured it out, `Gio.File.get_basename()` no longer returns strings with spaces replaced by `%20`. This means that when removing the basename from the end of the uri using `len()` will leave 3 characters on the end. Adding `.replace(' ', '%20') to the end of `Gio.File.get_basename` fixes this issue. Not sure if there is a more elegant way to do this. 

I have squashed this change with my previous commit. Please excuse all the updates as I was having difficulties with `git rebase` again.

PR #264, opened 2016-05-17
Fixes exaile/exaile#253

tfree87:
It seems that `Gio.File.get_basename` does not replace spaces with `%20`. This causes errors when exporting playlists that have spaces in the name of the exported file. This simple change addresses that issue.

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -294,7 +294,8 @@ class FormatConverter(object):
         if options is not None and options.relative:
             playlist_file = Gio.File.new_for_uri(playlist_path)
             # Strip playlist filename from export path
-            export_path = playlist_file.get_uri()[:-len(playlist_file.get_basename())]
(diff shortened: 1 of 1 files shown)


xl/playlist.py
```diff
@@ -294,7 +294,8 @@ def get_track_export_path(self, playlist_path, track_path, options):
         if options is not None and options.relative:
             playlist_file = Gio.File.new_for_uri(playlist_path)
             # Strip playlist filename from export path
-            export_path = playlist_file.get_uri()[:-len(playlist_file.get_basename())]
```
virtuald: Why not `playlist_file.get_parent().get_uri()`?

tfree87: That is a much more elegant way of addressing this issue. Testing this change shows that it also works. I will make a new commit for this.

PR #265, opened 2016-05-19
added plugin 'exaile-shuffle-collection'

rain0r:
Works with exaile 4.x (see: https://github.com/exaile/exaile/pull/227)


plugins/exaile-shuffle-collection/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+import glib
+import gtk
```
virtuald: These are incorrect, it should be:

```
from gi.repository import Gtk
```

virtuald: Actually, you can remove the 'import glib' and 'import gtk' lines, as you don't use either of those imports in your plugin.

plugins/exaile-shuffle-collection/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            Callback for when a track starts. The next track will be
+            added when a track starts, not when a track fades out.
+        '''
+        if not self.do_shuffle:
+            return
+        self.play()
+
+    def refresh_track_list(self):
+        self.tracks = list()
+        for track in self.exaile.collection.get_tracks():
```
virtuald: Why do you create your own list? What if the user has a really large collection? Can't you access the collection by index?

virtuald: Hm, apparently you can't, I stand corrected. I don't like the copying though... it seems like it would be better to add a 'get_random_track' to the collection instead.

rain0r:
Okay, thanks for the input!

What do you think of that?

Reworked the plugin: https://github.com/rain0r/exaile-shuffle-collection/blob/master/__init__.py
Added get_random_track() in TrackDB: https://gist.github.com/rain0r/4703db1768f58ef8aef26bb9b4385317

virtuald:
I think if you updated this PR with those changes, that would probably be ok.

rain0r:
Okay, I will create an updated PR.

virtuald:
For future reference, when you push new commits to a branch, those get added to the PR, so you don't need to create a new one. :)

PR #268, opened 2016-05-25
Remove deprecated Gtk.Table in plugins and replace with Gtk.Grid

tfree87:
I was looking to see if there were any remaining uses of the deprecated `Gtk.Table` and I found the following which I replaced with a `Gtk.Grid` instance. This should make any future transition to Gtk4 easier.

I had to rename some functions due to the name clash of 'Gtk.Grid.remove_rows()` with the some class methods.

### DIFF
--- a/xlgui/widgets/filter.py
+++ b/xlgui/widgets/filter.py
@@ -69,7 +69,7 @@ class FilterDialog(Gtk.Dialog):
         top.show_all()
 
         self.filter = f = FilterWidget(sorted(criteria, key=lambda k: _(k[0])))
-        f.add_row()
+        f.add_criteria_row()
         f.set_border_width(5)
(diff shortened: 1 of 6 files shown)


xlgui/widgets/filter.py
```diff
@@ -260,38 +259,21 @@ def add_row(self):
...
+        self.attach(criterion, 0, self.n, 1, 1)
+        self.attach(remove_btn, 1, self.n, 1, 1)
 
         self.rows.append((criterion, remove_btn, remove_btn_handler_id))
         self.n += 1
 
-    def remove_row(self, row):
+    def remove_criteria_row(self, row):
         """Remove a criteria row."""
-        rows = self.rows
```
virtuald: I like the huge amount of deletions here.

PR #269, opened 2016-06-01
Fix Gio.File.replace() error when using Playlist Analyzer

tfree87:
When saving an HTML file from the Playlist Analyzer the following traceback occurs:

`Gio.File.replace() takes exactly 5 arguments (3 given)`

Apparently in `Gio.File.replace()` all the arguments are now required. This pull request simply supplies those arguments and passes the file content as a string so that the file can be written by`fp.write()`


virtuald:
Thanks!

Issue #270, opened 2016-06-02
Non-Latin search always matches every track


sjohannes:
1. Edit a track, add `中` to the title.
1. Search for `中` in the Collections panel.

Expected: Only tracks containing the character match.

Actual: Every track matches.

This is a regression from e5ea6ba (#176 / #261).  The issue is that `unicodedata.normalize('NFKD', u'中').encode('ascii')` returns `?` instead of the original character. (For the record, `GLib.utf8_normalize(u'中', 1, GLib.NormalizeMode.NFKD)` also has the same issue.)

Note that unicode searching in the tracklist (right pane) is currently broken for a different reason, so this can only be tested in panels and maybe smart playlists.

tfree87:
A good solution for this might be to go through each character in the `search_string` and normalize each one. If the character can't be normalized to an ascii character (i.e. `unicodedata.normalize('NFKD', search_string).encode('ASCII','ignore')` returns `None`), then the original character can be returned.

virtuald:
This was fixed in #270, now it appears that zero tracks are matched when searching for that character. I'm adding a unit test to make sure that happens -- and if that's not what's supposed to happen, then we can fix it.

virtuald:
I stand corrected, it does the search correctly.

PR #275, opened 2016-06-02
Fix exaile/exaile #270 (Non-Latin search always matches every track)

tfree87:
This fix allows for normalization of strings while ignoring any unicode characters that cannot be converted into ascii.

### DIFF
--- a/xl/trax/search.py
+++ b/xl/trax/search.py
@@ -27,7 +27,7 @@
 import time
 import re
 import unicodedata
-
+import string
 
 __all__ = ['TracksMatcher', 'search_tracks']
 
@@ -68,9 +68,10 @@ class _Matcher(object):
             for item in vals:
                 if item != None:
                     try:
-                        item = self.lower(normalize_string(item))
+                        item = item.decode('ascii')
                     except:
-                        pass
+                        item = shave_marks(item)
+                    item = self.lower(item)
                     if self._matches(item):
                         return True
                     else:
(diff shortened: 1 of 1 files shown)


xl/trax/search.py
```diff
@@ -544,3 +541,25 @@ def match_track_from_string(track, search_string,
     matcher = TracksMatcher(search_string, case_sensitive=case_sensitive,
         keyword_tags=keyword_tags)
     return matcher.match(SearchResultTrack(track))
+
+def normalize_string(string):
```
virtuald: I haven't profiled it, but I have to imagine this is going to make searches really slow since it's running on every file. Can we find a more performant solution?

tfree87:
I believe this commit should make the searches more efficient. I have changed the code so that text is only normalized if it cannot be decoded into ascii by `[].decode('ascii')` this way we don't have to loop through all the characters of every possible search item. In addition, I borrowed a function from Fluent Python (http://bit.ly/1Q5lAAm) that will only remove diacritics off of non-latin characters instead of using the `normalize_string()` function that I created before which I think may be more efficient.

virtuald:
This seems better, though if your entire collection is non-ascii, then you're still going to pay overhead. Not sure how to get around that other than to store a normalized version in the database? But then it's a different form of overhead and easier to mess up...

virtuald:
FYI, this throws exceptions when tags aren't stored as a string.

PR #282, opened 2016-07-26
Cleanups

javiercantero:
Some cleanups I made while reviewing the code.

### DIFF
--- a/xl/formatter.py
+++ b/xl/formatter.py
@@ -410,7 +410,7 @@ class TrackFormatter(Formatter):
 
         return Formatter.format(self)
 
-class TagFormatter():
+class TagFormatter(object):
     """
         A formatter provider for a tag of a track
     """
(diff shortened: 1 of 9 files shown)


virtuald:
I'm on vacation until August 6, but if someone else doesn't review it I'll do it then.

javiercantero:
Ok, thanks @virtuald

xl/formatter.py
```diff
@@ -410,7 +410,8 @@ def format(self, track, markup_escape=False):
 
         return Formatter.format(self)
 
-class TagFormatter():
+# XXX: should inherite from class Formatter?
```
genodeftest: This should probably be clarified before merging. Otherwise the code looks good to me.

javiercantero: Well, the thing is since the other classes in the module all are subclasses of Formatter base class, I don't know if it's a mistake or deliberate. Only if it's deliberate, it should inherit from object.

virtuald:
This looks great, thanks!

PR #284, opened 2016-08-09
Fix searching when non-string tags are present

virtuald:
The problem this resolves is that mp4 returns BPM as an integer, not a string like the other formats. 

@sjohannes does this seem sane, or are there searches that I'm going to accidentally break by converting everything to a string? I'm not really sure if the `str(item)` conversion should be here... really, it seems like get_tag_search() should already be returning the correct thing? But if that's true, then we probably should move the unicode conversion over to get_tag_search also... 

Fixes a bug introduced by #261, #275

### DIFF
--- a/tests/xl/trax/test_search.py
+++ b/tests/xl/trax/test_search.py
@@ -411,3 +411,14 @@ class TestSearchTracks(object):
         assert gen.next().track == tracks[2]
         with pytest.raises(StopIteration):
             gen.next()
+            
+    def test_search_tracks_with_int_from_string(self):
+        # unlike mp3, mp4 will return integers for BPM.. make sure that works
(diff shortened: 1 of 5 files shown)


sjohannes:
> really, it seems like get_tag_search() should already be returning the correct thing? But if that's true, then we probably should move the unicode conversion over to get_tag_search also...

Yeah I think that's the better solution, for get_tag_search to normalise the tag values (unless it's also used for greater-than/less-than comparisons?).

virtuald:
It is used for gt/lt comparisons, but those already explicitly cast the value to a float. I've reorganized it to get rid of special cases... it seems to work for me, can you think of anything it'll break?

sjohannes:
Looks fine to me.

PR #289, opened 2016-08-15
Misc fixes for python 3

mexicarne:
Hi, I use this patches locally to make a FreeBSD package built against python3. Please consider to merge.

### DIFF
--- a/plugins/daapclient/__init__.py
+++ b/plugins/daapclient/__init__.py
@@ -307,12 +307,13 @@ class DaapManager:
             
         self.history = DaapHistory(5, menu=hmenu, callback=self.connect_share)
 
-    def connect_share(self, obj, (name, address, port, svc)):
+    def connect_share(self, obj, *args):
         '''
             This function is called when a user wants to connec to
         a DAAP share.  It creates a new panel for the share, and
         requests a track list.
         '''
+        name, address, port, svc = args
         conn = DaapConnection(name, address, port)
         
         conn.connect()
--- a/tools/generate-launcher
+++ b/tools/generate-launcher
@@ -20,7 +20,7 @@ PYTHON2_CMD=${5:-python2}
 #  exit 1
 #fi
 
-EPREFIX=`$PYTHON2_CMD -c 'import os,sys;print os.path.realpath(sys.argv[1])' $EPREFIX`
+EPREFIX=`$PYTHON2_CMD -c 'import os,sys;print(os.path.realpath(sys.argv[1]))' $EPREFIX`
 
 cd ${DESTDIR}${EPREFIX}/bin && \
 cat <<EOF > exaile
--- a/xlgui/panel/files.py
+++ b/xlgui/panel/files.py
@@ -270,7 +270,7 @@ class FilesPanel(panel.Panel):
             Refreshes the current view
         """
         cursorf = self.model[self.tree.get_cursor()[0]][0]
-        print cursorf.get_uri()
+        print(cursorf.get_uri())
         self.load_directory(self.current, False, cursor_file=cursorf)
 
     def entry_activate(self, widget, event=None):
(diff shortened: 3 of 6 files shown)


virtuald:
I would take f4ac7a16fb0dd6539f2110e3538803de2063b5db as a separate PR.

Did you actually try this with python3, I think there are a lot more errors than this (see #245)?

plugins/daapclient/__init__.py
```diff
@@ -307,12 +307,13 @@ def hmfactory(menu, parent, context):
             
         self.history = DaapHistory(5, menu=hmenu, callback=self.connect_share)
 
-    def connect_share(self, obj, (name, address, port, svc)):
+    def connect_share(self, obj, *args):
```
virtuald: This isn't the same, you should omit the *.

xlgui/panel/files.py
```diff
@@ -270,7 +270,7 @@ def refresh(self, widget):
             Refreshes the current view
         """
         cursorf = self.model[self.tree.get_cursor()[0]][0]
-        print cursorf.get_uri()
```
virtuald: This print statement can be removed, or changed to a logging statement.

tools/generate-launcher
```diff
@@ -20,7 +20,7 @@ PYTHON2_CMD=${5:-python2}
 #  exit 1
 #fi
 
-EPREFIX=`$PYTHON2_CMD -c 'import os,sys;print os.path.realpath(sys.argv[1])' $EPREFIX`
+EPREFIX=`$PYTHON2_CMD -c 'import os,sys;print(os.path.realpath(sys.argv[1]))' $EPREFIX`
```
virtuald: We aren't dropping python2 support, so maybe do `sys.stdout.write` instead to retain compatability for both?

plugins/daapclient/__init__.py
mexicarne: Ok, it seems you are right about this one.

I didn't yet actually tried with python3, because there is still some packaging issue with python3.5 (exaile's Makefile does not installing 3.5's __cache__  dirs with bytecode in them, only py[co] bytecode that is produced with pythons < 3.5), but with these changes it compiles to bytecode without any errors. So I believe it is the way to go. 

I done exaile-devel FreeBSD port (from git master) and it works flawlessly (I'm using it about month already). But I want it to use python3, so here is the changes.

I'll try to overcome __cache__ issue and make it a try with python3 today. Will let you know.

tools/generate-launcher
mexicarne: And btw, most of exaile codebase is using print as function, so why should we care about compatibility with prehistoric pythons in couple of leftover places?

virtuald: Because we still support python2.

And you're right, because it printed out a single parameter, it prints out correctly, I was wrong.

virtuald:
I've cherry-picked 3317eba1c4e648a71ec6fbff79bdde0de03709a6, 74e3eb2239674a257f242380893129ea5d093633, and ac3b221eb598f3fb4806918d1c4737641aa4ee28 into master. The other changes are fine if you address my comments.

mexicarne:
Hello, so I finally tested it with python3 yesterday and it doesn't work. And this is part of the problem why I have a troubles with `__cache__` files when making package. So I'll add development package to FreeBSD repository that is running against python2 and will retest the things when #245 would be merged. 

Anyway, what you expect me to do to get fd5bb19 and acd988d merged? First is fine as it is, the second just need asterisk removed. Should I just fix that asterisk and you'll merge?

virtuald:
Yes please.

virtuald:
I have #245 on my radar, but I'm trying to focus on fixing some of the more critical bugs first. Feel free to take a whack at it.

tools/generate-launcher
virtuald: Done.

Issue #299, opened 2016-10-11
Export Files playlist dialog is broken


javiercantero:
The dialog to export files from a playlist (launched with the 'Export Files' context menu option over such playlist tab) fails with this error:
```
Traceback (most recent call last):
  File ".../exaile/xlgui/widgets/dialogs.py", line 1614, in <lambda>
    dialog.connect( 'uris-selected', lambda widget, uris: _on_uri(uris[0]))
  File ".../exaile/xlgui/widgets/dialogs.py", line 1604, in _on_uri
    if hasattr(playlist, 'get_playlist'):
UnboundLocalError: local variable 'playlist' referenced before assignment
```
This is happening since the commit 1a8eb62 that allowed exporting of SmartPlaylists  (#216).

javiercantero:
Ok, I found it: in [line 1605](https://github.com/exaile/exaile/blob/master/xlgui/widgets/dialogs.py#L1605) of `xlgui/widgets/dialogs.py` the commit tries to modify the free variable `playlist` of the `_on_uri()` closure, but in Python the binding of a free variable in a closure is **read-only**. The solution is simple: don't assign `playlist`.

javiercantero:
Fixed with #300

Issue #198, opened 2015-12-14
copy_async no longer exists - playlist file export breaks


virtuald:

javiercantero:
The error message is:
```
Traceback (most recent call last):
  File ".../exaile/xlgui/widgets/dialogs.py", line 1616, in <lambda>
    dialog.connect( 'uris-selected', lambda widget, uris: _on_uri(uris[0]))
  File ".../exaile/xlgui/widgets/dialogs.py", line 1611, in _on_uri
    dialog.do_copy()
  File ".../exaile/xlgui/widgets/dialogs.py", line 1394, in do_copy
    self._start_next_copy()
  File ".../exaile/xlgui/widgets/dialogs.py", line 1462, in _start_next_copy
    self.source.copy_async( self.destination, self._finish_single_copy, flags=flags, cancellable=self.cancel ) 
AttributeError: 'GLocalFile' object has no attribute 'copy_async'
```
The problem is that the Gio function `g_file_copy_async()` is not introspectable in the current gobject-introspection package, hence it can be accessed using python gi. That's the situation since the commit GNOME/glib@89b558077f97c630c5b26b3527a6c7d0515f1f2c in which they disable it (the "(skip)" bit in the annotated function). Note that commit is from 2010.

The cause of such omission is harder to trace, but I think the GNOME bug [616044](https://bugzilla.gnome.org/show_bug.cgi?id=616044) explains it, plus its duplicate [708297](https://bugzilla.gnome.org/show_bug.cgi?id=708297) and the comments in [746111](https://bugzilla.gnome.org/show_bug.cgi?id=746111). In summary, they don't know how to handle functions with two callbacks and two different scopes for the closure variables. The Ubuntu bug [1033196](https://bugs.launchpad.net/ubuntu/+source/gobject-introspection/+bug/1033196) (from 2012) supports this too.

I see three possible solutions:

- Push for upstream to solve their bug and make `copy_async` available to everybody using g-i. That's the best-case scenario, but unfortunately I doubt this is going to happen soon (if happened) since 7 years have already been passed and nobody has done anything yet.
- Use the python machinery to write a non-blocking copy. For example, do the copy in a python thread should suffice. The problem is that adding more threads to the current situation (meaning #83) might not be such a good idea.
- Since 'Export Files' is not a essential feature, disable it for Exaile 4.0 and delay any solution to 4.0.1 or a later version.

virtuald:
[...]

PR #300, opened 2016-10-11
Fix Export Files playlist dialog (#299)

javiercantero:
The dialog to export files from a playlist (launched with the 'Export
Files' context menu option over such playlist tab) fails because it
tries to assign a value to a free variable in a closure (which are
read-only in Python).

This patch modifies the logic in the closure so the free variable is
only used as an R-value.


virtuald:
LGTM... does the export files dialog actually work? I thought there was a bug related to async copy not existing (reported as #198).

javiercantero:
@virtuald Since I was looking for #198, I had to fix this one first.

Issue #302, opened 2016-10-18
Exception caught when passing a song through the command line


javiercantero:
```
$ ./exaile a.mp3 
INFO    : Using PyGObject 3.22.0
INFO    : Loading Exaile 4.0.0dev+8eed81e on Python 2.7.12+...
INFO    : Loading settings...
INFO    : Using es_ES UTF-8 locale
INFO    : Using GTK+ 3.22.1
INFO    : Using GStreamer 1.8.3
INFO    : Reconfiguring crossfading
INFO    : Crossfade: disabled
INFO    : Loading plugins...
INFO    : Loading collection...
INFO    : Loading devices...
INFO    : Connected to udisks2
INFO    : Loading interface...
INFO    : Loading main window...
INFO    : Connecting main window events...
INFO    : Loading panels...
INFO    : Connecting panel events...
INFO    : Done loading main window...
ERROR   : Event callback exception caught!
Traceback (most recent call last):
  File ".../exaile/xl/event.py", line 343, in _emit
    event.data, *cb.args, **cb.kwargs)
  File ".../exaile/xlgui/main.py", line 654, in on_playlist_tracks_added
    self.statusbar.update_info()
  File ".../exaile/xlgui/widgets/info.py", line 749, in update_info
    self.info_label.set_label(self.formatter.format())
  File ".../exaile/xl/formatter.py", line 300, in format
    substitute = substitute(*args, **parameters)
  File ".../exaile/xlgui/widgets/info.py", line 594, in get_collection_count
    return _('%d in collection') % main.exaile().collection.get_count()
  File ".../exaile/xl/main.py", line 727, in exaile
    raise AttributeError(_("Exaile is not yet finished loading"
AttributeError: Exaile is not yet finished loading. Perhaps you should listen for the exaile_loaded signal?
INFO    : Playing file://.../exaile/a.mp3
```

[...]

PR #303, opened 2016-10-18
Don't start playing songs until the UI is loaded

javiercantero:
Postpone playing songs from either the command line or the restored
player state until the UI is loaded and operative. (Fixes: #302)


virtuald:
LGTM, thanks!

PR #309, opened 2016-12-14
Fix vorbis default quality

genodeftest:
According to the GStreamer documentation, module `vorbisenc`, `quality` ranges from −0.1 to +1.0.

Use setter functions to make sure the issue is not being made again.


sjohannes:
Merged, thanks.

PR #310, opened 2016-12-14
pylint: cleanup main, doc, tests, tools

genodeftest:
Use pylint to clean up the exaile{,_osx,_win}.py and all python files in doc/, tests/ and tools/

Those changes are just cosmetic, but I intend to continue cleaning up all of exaile's code to ease reading code and finding bugs.

I'm doing this pull request now asking you whether you are interested in such work. It'll probably take a day or two where other commits are not welcome at the same time because rebasing is ugly.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -11,7 +11,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
+import sys
+import os
 
 from os.path import abspath, join, dirname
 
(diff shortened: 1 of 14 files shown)


virtuald:
I'm not opposed to this... though, if you're going to go through all this effort, I would recommend adding something to the unit testing that happens on Travis-CI to ensure that people don't break it in the future, otherwise it'll just get ignored.

However, if we went that route, there are certain lint rules I'd rather not see enforced -- in particular, I dislike enforcing removal of whitespace between lines:

```
____indented_statement
____
____another_statement
```

I dislike it when people remove the center spaces. I realize it's unnecessary content in the file, but it makes it annoying to arrow through the file or add something to that line without hitting enter.

virtuald:
Also, I dislike enforcing an 80-character limit religiously. While it's often a good idea to keep yourself within that range, enforcing it is annoying.

genodeftest:
Thanks!
I'll run autopep8 with a few specific rules and leave room for discussion in a pull request where we can set these rules.

PR #311, opened 2016-12-17
remove deprecated xalign from builder XML files

genodeftest:
xalign property is deprecated since Gtk+ 3.14.
Removing it has no side-effects in recent versions of Gtk+.

No side effects to be expected. This stuff was just hold back by Glade always re-adding the field on save [(see this GNOME bug)](https://bugzilla.gnome.org/show_bug.cgi?id=750985)


sjohannes:
Merged, thanks.

PR #313, opened 2016-12-18
Replace deprecated Gtk.Alignment (fix warnings)

genodeftest:
Replace or remove all remaining usages of Gtk.Alignment, which has been deprecated in Gtk+ 3.14.

### DIFF
--- a/xlgui/cover.py
+++ b/xlgui/cover.py
@@ -1046,19 +1046,8 @@ class CoverChooser(GObject.GObject):
 
         self.cover_image_box = self.builder.get_object('cover_image_box')
 
-        self.loading_indicator = Gtk.Alignment()
-        self.loading_indicator.props.xalign = 0.5
-        self.loading_indicator.props.yalign = 0.5
-        self.loading_indicator.set_size_request(350, 350)
-        self.cover_image_box.pack_start(self.loading_indicator, True, True, 0)
-
-        try:
-            spinner = Gtk.Spinner()
-            spinner.set_size_request(100, 100)
-            spinner.start()
-            self.loading_indicator.add(spinner)
-        except AttributeError: # Older than GTK 2.20 and PyGTK 2.22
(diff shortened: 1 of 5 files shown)


sjohannes:
GitHub is really confused about what I've done. Here's a summary:

* I've merged the first two commits.
* For the third commit, I'd like the padding back.
* The fourth commit is not necessary anymore because I've removed the class.
* I haven't reviewed the fifth commit.

genodeftest:
Ok, that's weird. I don't really get what I need to do here, so just tell me what I should do.

xlgui/cover.py
```diff
@@ -1046,19 +1046,8 @@ def __init__(self, parent, track, search=None):
...
-
-        try:
-            spinner = Gtk.Spinner()
-            spinner.set_size_request(100, 100)
-            spinner.start()
-            self.loading_indicator.add(spinner)
-        except AttributeError: # Older than GTK 2.20 and PyGTK 2.22
-            self.loading_indicator.add(Gtk.Label(label=_('Loading...')))
+        self.__stack = self.builder.get_object('__stack')
+        self.__stack_ready = self.builder.get_object('__stack_ready')
```
sjohannes: Just use `stack` and `stack_ready` for the variable and GtkBuilder object names.

sjohannes:
Remaining things to address:

* Third commit: re-add padding.
* Fifth commit: change variable & builder object names.

Just add the changes on top of your current branch, if that's easier; I can do the rebasing.

xlgui/cover.py
genodeftest: Oh, I've chosen `__stack` because I prefer to have fields private where possible. And I did the same in .ui files because they might be migrated to GtkTemplate eventually. What do you think? Still change it?

genodeftest:
I changed the code according to your request. Thank you and thanks to `git cherrry-pick`!

sjohannes:
All merged, thanks.

PR #314, opened 2016-12-18
Cover window: Fix dialog being too wide

genodeftest:
Fixes a bug where the track properties window would become too wide for the
screen so user experience would suffer due to unreachable buttons.

Steps to reprodce the original bug:
1. have a file with comment containing some thousand characters in one line
2. open exaile's track properties dialog on it

What happens:
Dialog is very wide to fit the longest comment line without breaking it.

What should happen:
Either wrap the line (realized solution) or scroll horizontally (worse user experience because more user interaction needed for reading the text).

Additional info:
This solution has a downside: Shrinking the track properties dialog is very CPU intensive due to recalculation of text inside the GtkTextView.


virtuald:
I observed the problem you mentioned. I think I'd rather have more correct behavior, as I suspect most people don't have comments of this nature.

PR #316, opened 2016-12-19
Implement async file copy (work around #198)

genodeftest:
Because of a GObject introspection bug (GNOME bug #616044, #708297, #746111).
This patch works around the issue by implementing the async copy behavior in a separate thread.

How to test / how I tested this patch:
1. open any playlist
2. right-click on playlist tab, select "Export Files"
3. select a directory:
    * one without the files
    * one with the files already present
    * one without write permissions (e.g. `/usr/`)
4. If file already exists, try each button

### DIFF
--- a/xlgui/widgets/dialogs.py
+++ b/xlgui/widgets/dialogs.py
@@ -51,6 +51,7 @@ from xl.playlist import (
 from xl.nls import gettext as _
 
 from xlgui.guiutil import GtkTemplate
+from threading import Thread
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


genodeftest:
After this patch is merged, #198 should no longer block the Exaile 4.0 release milestone. I don't care whether you want to close #198 or keep it open for future reference.

virtuald:
I won't be able to test this until later tonight, but this seems like a good workaround. I dislike spawning a separate thread for each operation (a threadpool would be better... but we're in python so that's harder than I would like), but for a high latency operation such as this I think it's probably not a big deal.

virtuald:
Works for me, thanks!

PR #317, opened 2016-12-19
Lyrics improvements

genodeftest:
Fixing minor bugs, improving code quality, removing old and unused stuff.


genodeftest:
By the way: What do you think of putting Lyrics Viewer into "core" exaile, i.e. have it not be a plugin, but live below `xlgui.panel`? I would prefer that because other plugins (lyricswiki, lyricsmania) depend on it. Plugin dependencies don't really work well: If a user enabled the plugin it might still be dysfunctional because no content providers are available or because the panel is hidden. Putting it into exaile core codebase would ease this a bit.

virtuald:
I'm undecided on your suggestion.

genodeftest:
What about the pull request? The suggestion can wait and is mostly unrelated.

PR #321, opened 2016-12-22
Fix cd plugin/panel and transcoder

### LINKED ISSUE #320: CD ripping / "Importing": Unable to set format
transcoder.py specifies a `FORMATS` array of encoder information. But on the "Audio Disc" panel there is no way to set any format. I couldn't find one in preferences either.

Furthermore, the "steps" metadata is outdated, at least for FLAC.

genodeftest:
Fixes #320 and a few other issues related to the CD plugin/panel, making it work. Still room for improvement, but enough to ship in Exaile 4.0 pre-release for now in my opinion.

flac quality levels: There is a related GStreamer bug: https://bugzilla.gnome.org/show_bug.cgi?id=776413
tl;dr: GStreamer says, flac has quality levels 0…9 but it really has just level 0…8 so setting level 9 will always fail.

### DIFF
--- a/plugins/cd/cdprefs.py
+++ b/plugins/cd/cdprefs.py
@@ -25,12 +25,9 @@
 # from your version.
 
 from __future__ import with_statement
-from gi.repository import GObject
-from gi.repository import Gtk
 import os
-import warnings 
 
-from xl import event, settings, transcoder
+from xl import settings, transcoder
 from xl.nls import gettext as _
 from xlgui.preferences import widgets
 
--- a/plugins/cd/cdprefs.py
+++ b/plugins/cd/cdprefs.py
@@ -63,7 +63,11 @@ class OutputQualityPreference(widgets.ComboPreference, widgets.Conditional):
             :rtype: bool
         """
         model = self.widget.get_model()
-        format = self.condition_widget.get_active_text()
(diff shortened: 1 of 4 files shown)


plugins/cd/cdprefs.py
```diff
@@ -25,14 +25,12 @@
...
-from gi.repository import GObject
-from gi.repository import Gtk
 import os
-import warnings 
 
-from xl import event, settings, transcoder
+from xl import settings, transcoder
 from xl.nls import gettext as _
 from xlgui.preferences import widgets
+from __builtin__ import str
```
virtuald: Why is this here?

genodeftest: Probably a bug in PyDev.

plugins/cd/cdprefs.py
```diff
@@ -30,7 +30,6 @@
 from xl import settings, transcoder
 from xl.nls import gettext as _
 from xlgui.preferences import widgets
-from __builtin__ import str
```
genodeftest: Or should stash this change into the original pull request which introduced this change?

virtuald: I personally like rebasing such changes, but I'll leave it up to you.

genodeftest:
More compact version now.

Issue #35, opened 2015-01-11
"no suitable plugins found" message is not quite helpful


genodeftest:
I was unable to play MP3 files and got this message in GUI: `no suitable plugins found`. I didn't know why. I don't think a normal user would find out what to do in this case.

On console I got this message:

```
<gst.Message GstMessageError, gerror=(GError)NULL, debug=(string)"gstdecodebin2.c\(3576\):\ gst_decode_bin_expose\ \(\):\ /GstPlayBin2:player/GstURIDecodeBin:uridecodebin3/GstDecodeBin2:decodebin23:\012no\ suitable\ plugins\ found"; from decodebin23 at 0x7ff0b4068b10> ['__class__', '__cmp__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__grefcount__', '__gstminiobject_init__', '__gtype__', '__hash__', '__init__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'copy', 'flags', 'parse_async_start', 'parse_buffering', 'parse_buffering_stats', 'parse_clock_lost', 'parse_clock_provide', 'parse_duration', 'parse_error', 'parse_info', 'parse_new_clock', 'parse_qos', 'parse_qos_stats', 'parse_qos_values', 'parse_request_state', 'parse_segment_done', 'parse_segment_start', 'parse_state_changed', 'parse_step_done', 'parse_step_start', 'parse_stream_status', 'parse_structure_change', 'parse_tag', 'parse_tag_full', 'parse_warning', 'set_buffering_stats', 'set_qos_stats', 'set_qos_values', 'set_seqnum', 'src', 'structure', 'timestamp', 'type'] (xl.player._base)
```

After installing gstreamer-plugins-ugly this error message was gone.

It would be useful to add a hint that the user should install gstreamer plugins and possibly add the file format. And/or a link to https://exaile.readthedocs.org/en/latest/user/faq.html#error-no-suitable-plugin-found-when-playing-a-mp3-m4a-etc

genodeftest:
for m4a files this is gstreamer-ffmpeg
for wma files this is gstreamer-plugins-bad
for aac there seems to be no plugin for gstreamer 0.10

For more details see 
http://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-good-plugins/html/
[...]

PR #335, opened 2017-04-25
Installer using MSYS2

dangmai:
Updated Exaile to use with the new `python-gtk3-gst-sdk`. The installer is created fine, however it is a bit large since it contains all the Gst plugins (~30mb out of the ~50mb generated installer). I'm not 100% sure which plugins we use, so I'll leave all of them in for now, but we can delete them before release if needed. Also, right now PyInstaller is pointing to my fork, we'll need to change that once they merge it in.

### DIFF
--- a/tools/installer/_build.sh
+++ b/tools/installer/_build.sh
@@ -1,3 +1,9 @@
+set -e
+
+# We add the bin dirs from the build root to the PATH, because there
+# are a lot of instances in the legacy code where we make assumptions
+# about where the programs are (e.g. help2man)
+init_env
 
 if [ "$TARGET" == "" ]; then
   TARGET=`pwd`
@@ -11,6 +17,18 @@ else
   DIST=_dist
 fi
 
+function build_git {
+  "${BUILD_ROOT}"/usr/bin/git.exe "$@"
+}
+
+function build_tar {
+  "${BUILD_ROOT}"/usr/bin/tar.exe "$@"
+}
+
+function build_make {
+  "${BUILD_ROOT}"/usr/bin/make.exe "$@"
+}
+
 EXAILE_DIR="$TARGET"/../..
 COPYDIR="$TARGET"/_copy
 DESTDIR="$TARGET"/_inst
@@ -22,12 +40,16 @@ for d in _copy _inst _build _build_osx $DIST; do
 done
 
 pushd "$EXAILE_DIR"
-git archive HEAD --prefix=_copy/ | tar -x -C tools/installer/
+build_git archive HEAD --prefix=_copy/ | build_tar -x -C tools/installer/
 popd
 
 pushd "$COPYDIR"
-make
-PREFIX=/usr DESTDIR="$DESTDIR" make install
+# Our Makefile relies on $PYTHON2_CMD to run Python commands
+export PYTHON2_CMD="${BUILD_ROOT}"/"${MINGW}"/bin/"${PYTHON_ID}".exe
+# help2man relies on perl-Locale-gettext, which lives in this directory
+export PERL5LIB="${BUILD_ROOT}"/usr/lib/perl5/vendor_perl
+build_make
+PREFIX=/usr DESTDIR="$DESTDIR" build_make install
 
 # Copy things that the unix install doesn't require..
 if [ "$SDK_PLATFORM" == "darwin" ]; then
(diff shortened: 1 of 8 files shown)


tools/installer/requirements.txt
```diff
@@ -1,3 +1,5 @@
+git+git://github.com/exaile/pywin32-ctypes.git@pyinstaller#egg=pywin32-ctypes
+git+git://github.com/dangmai/pyinstaller.git@msys2#egg=PyInstaller
 musicbrainzngs
 mutagen==1.31
```
virtuald: We should bump the mutagen requirement.

Have you tried running this from scratch? I suspect that pyinstaller and pywin32-ctypes will need to be installed separately, as pip doesn't perform the install in order IIRC, so pyinstaller will add pywin32 to the dependency list.

genodeftest: Shouldn't this be a minimum reqirement? `mutagen>=1.31`?

dangmai: I just kept the requirement that we have currently, no opposition to bump it at all. Although I do think pinning our dependencies to a particular version is a good idea so that we can have a reproducible build. On the other hand we're running pacman to get the binary dependencies for the Windows build without pinning anyway, so maybe what we actually need a good integration test suite to detect regression (but ain't no body got time for that 😛 )

genodeftest: yeah, the integration test suite to detect regressions… I'd like to have that too ;)

virtuald: Well, there's a very insufficient set of unit tests. That's a start?

dangmai:
Just rebased this on top of `master`, and squashed away all the commits 🐕

genodeftest:
Sorry, I can't review this, I'm not using Exaile on Windows at the time.

virtuald:
No worries, I told him I was going to review it at some point. Just hadn't done so yet. I think with the gst thing out of the way for now I can do this tonight.

tools/installer/_build.sh
```diff
@@ -22,12 +40,16 @@ for d in _copy _inst _build _build_osx $DIST; do
...
+build_git archive HEAD --prefix=_copy/ | build_tar -x -C tools/installer/
 popd
 
 pushd "$COPYDIR"
-make
-PREFIX=/usr DESTDIR="$DESTDIR" make install
+# Our Makefile relies on $PYTHON2_CMD to run Python commands
+export PYTHON2_CMD="${BUILD_ROOT}"/"${MINGW}"/bin/"${PYTHON_ID}".exe
+# help2man relies on perl-Locale-gettext, which lives in this directory
+export PERL5LIB="${BUILD_ROOT}"/usr/lib/perl5/vendor_perl
```
virtuald: We don't really need to build a manpage on Windows. Maybe we can add a way to skip it?

virtuald: If we change the build command to be `build_make compile locale` then we wouldn't need to worry about the manpage nonsense.

tools/installer/_build.sh
```diff
@@ -22,12 +40,16 @@ for d in _copy _inst _build _build_osx $DIST; do
...
+build_make
```
dangmai: I'm not super versed about Make in general, do you mean I should change this to `build_make compile locale`, @virtuald ?

virtuald: Yeah, then you don't have to deal with the help2man stuff.

dangmai: Not working:

```
install -m 644 exaile.1.gz /home/DMai/Dev/exaile/tools/installer/_inst/usr/share/man/man1/
install: cannot stat 'exaile.1.gz': No such file or directory
make: *** [Makefile:144: install-target] Error 1
```

That's part of `install-target`, which (I think) is cross-platform.

virtuald: Oh that's stupid. We can fix that.

sjohannes: Should work now; please try again.

dangmai:
The installer works fine on my branch right now; however when I rebased it on top of `master`, it no longer works. I've traced it down the inclusion of `GstPbutils` mentioned [here](https://github.com/exaile/exaile/issues/35). I ran the code by @genodeftest in that issue fine, so my guess is that it's not being included by PyInstaller. Here's the stack trace:

```
  File "C:/msys64/home/dangmai/Dev/exaile/tools/installer/_inst/usr/lib/exaile/exaile_win.py", line 110, in main
    exaile.main()
  File "C:/msys64/home/dangmai/Dev/exaile/tools/installer/_inst/usr/lib/exaile/exaile.py", line 57, in main
    exaile = main.Exaile()
  File "C:/msys64/home/dangmai/Dev/exaile/tools/installer/_inst/usr/lib/exaile/xl/main.py", line 256, in __init__
    _do_heavy_imports()
  File "C:/msys64/home/dangmai/Dev/exaile/tools/installer/_inst/usr/lib/exaile/xl/main.py", line 54, in _do_heavy_imports
    gi.require_version('GstPbutils', '1.0')
  File "C:/msys64/home/dangmai/Dev/exaile/tools/installer/_build_root/mingw32/lib/python2.7/site-packages/gi/__init__.py", line 118, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace GstPbutils not available
```

virtuald:
I can dig into that.

dangmai:
Thanks! When you rebase/merge with `master` make sure to pass `PREFIX` and `DESTDIR` into Make instead of using environmental variables, I got stuck there for quite a while :)

virtuald:
When I run the packed app I get errors about libass and libx264... but I get those errors when I install msys2 separately anyways.

If we create a script to remove the plugins we don't care about, then the error will go away... probably need to create a 'remove_video_plugins' function in the SDK.

virtuald:
Actually, I don't even think we should really care about GstPbutils... it's not like the user can upgrade the package anyways? Maybe we disable it in a frozen build.

virtuald:
Just needed additional hooks in pyinstaller: https://github.com/pyinstaller/pyinstaller/pull/2603

genodeftest:
`GstPbutils` is not quite useful on Windows, that's true.

PR #336, opened 2017-04-25
GStreamer Engine: Assist user on installing plugins

genodeftest:
This is a proposal for fixing #35, which is not yet ready to merge.

Assumptions:
* GstPbutils is always available. This might change requirements so that GStreamer plugins base is required by Exaile. This is relevant for distributions and packagers. I've added this to the spec file.

TODO, please help me on these issues / decisions:
* `__installer_finished_callback`: Should Exaile continue playback after successful installation? Should it bring its window to the front?
* ~~`__installer_finished_callback`: Should we handle any of the "failing" return codes?~~
* ~~`__handle_plugin_missing_message`: I have a "u" prefix in the message, how do I get rid of it?~~ solved.
* `__handle_plugin_missing_message`: This message looks bad (i.e. bad UI/UX), do you have any suggestions? Can we add an URL? How? Which URL?

Known issues:
* At least on Fedora 25, the installer process returns immediately, rendering the asynchronity useless and makes the result unreliable. I've reported this bug against PackageKit: https://bugs.freedesktop.org/show_bug.cgi?id=100791

Known side-effects:
* This change increases the minimum required GStreamer version from 1.4 to 1.6, released in early 2014. I doubt this is a problem, but tell me if I'm wrong.

Out of scope:
* Exaile could recommend a list of GStreamer plugins on first start.

### DIFF
--- a/xl/player/gst/engine.py
+++ b/xl/player/gst/engine.py
@@ -43,7 +43,10 @@ from .sink import create_device
 
 from xl.player.engine import ExaileEngine
 from xl.player.track_fader import TrackFader
+from xl.player.gst import missing_plugin
 
+
+logger = logging.getLogger(__name__)
 
 
 class ExaileGstEngine(ExaileEngine):
@@ -594,40 +597,54 @@ class AudioStream(object):
                 self.playbin.notify("volume")
         
         elif message.type == Gst.MessageType.ERROR:
-            
-            # Error handling code is from quodlibet
-            gerror, debug_info = message.parse_error()
-            message_text = ""
-            if gerror:
-                message_text = gerror.message.rstrip(".")
-                 
-            if message_text == "":
-                # The most readable part is always the last..
-                message_text = debug_info[debug_info.rfind(':') + 1:]
-                
-                # .. unless there's nothing in it.
-                if ' ' not in message_text:
-                    if debug_info.startswith('playsink'):
-                        message_text += _(': Possible audio device error, is it plugged in?')
-            
-            self.logger.error("Playback error: %s", message_text)
-            self.logger.debug("- Extra error info: %s", debug_info)
-            
-            envname = 'GST_DEBUG_DUMP_DOT_DIR'
-            if envname not in os.environ:
-                import xl.xdg
-                os.environ[envname] = xl.xdg.get_logs_dir()
+            self.__handle_error_message(message)
+        
+        elif message.type == Gst.MessageType.ELEMENT:
+            if not missing_plugin.handle_message(message, self.engine):
+                logger.debug(
+                    "Unexpected element-specific GstMessage received from %s: %s"
+                    % (message.src, message))
+        
+        elif message.type == Gst.MessageType.WARNING:
+            # TODO there might be some useful warnings we ignore for now.
+            gerror, debug_text = Gst.Message.parse_warning(message)
+            logger.warn("Unhandled GStreamer warning received:\n\tGError: %s\n\tDebug text: %s",
+                        gerror, debug_text)
+        
+        else:
+            # TODO there might be some useful messages we ignore for now.
+            logger.debug("Unhandled GstMessage of type %s received: %s",
+                         message.type, message)
+    
+    def __handle_error_message(self, message):
+        # Error handling code is from quodlibet
+        gerror, debug_info = message.parse_error()
+        message_text = ""
+        if gerror:
+            message_text = gerror.message.rstrip(".")
              
-            Gst.debug_bin_to_dot_file(self.playbin, Gst.DebugGraphDetails.ALL, self.name)
-            self.logger.debug("- Pipeline debug info written to file '%s/%s.dot'",
-                              os.environ[envname], self.name)
+        if message_text == "":
+            # The most readable part is always the last..
+            message_text = debug_info[debug_info.rfind(':') + 1:]
             
-            self.engine._error_func(self, message_text)
+            # .. unless there's nothing in it.
+            if ' ' not in message_text:
+                if debug_info.startswith('playsink'):
+                    message_text += _(': Possible audio device error, is it plugged in?')
         
-        # TODO: Missing plugin error handling from quod libet
-        # -- http://cgit.freedesktop.org/gstreamer/gstreamer/tree/docs/design/part-missing-plugins.txt
+        self.logger.error("Playback error: %s", message_text)
+        self.logger.debug("- Extra error info: %s", debug_info)
         
-        return True
+        envname = 'GST_DEBUG_DUMP_DOT_DIR'
+        if envname not in os.environ:
+            import xl.xdg
+            os.environ[envname] = xl.xdg.get_logs_dir()
+         
+        Gst.debug_bin_to_dot_file(self.playbin, Gst.DebugGraphDetails.ALL, self.name)
+        self.logger.debug("- Pipeline debug info written to file '%s/%s.dot'",
+                          os.environ[envname], self.name)
+        
+        self.engine._error_func(self, message_text)
         
     def on_source_setup(self, playbin, source, track):
         # this is for handling multiple CD devices properly
--- a/xl/player/gst/missing_plugin.py
+++ b/xl/player/gst/missing_plugin.py
@@ -1,0 +1,187 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2, or (at your option)
+# any later version.
+#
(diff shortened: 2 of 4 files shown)


xl/player/gst/missing_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+
+    desc = GstPbutils.missing_plugin_message_get_description(message)
+    installer_details = GstPbutils.missing_plugin_message_get_installer_detail(message)
+    logger.warn("A plugin for %s is missing", desc)
+
+    user_message = _(
+        "A GStreamer 1.x plugin for %s is missing. "
+        "Without this software installed, Exaile will not be able to play the current file. "
+        "Please install the required software on your computer."
+        % desc)
```
virtuald: The `% desc` should be outside the parentheses.

xl/player/gst/engine.py
```diff
@@ -594,40 +597,61 @@ def on_message(self, bus, message):
...
+            gerror, debug_text = Gst.Message.parse_warning(message)
+            print(gerror)
+            print(debug_text)
+            warning_details = Gst.Message.parse_warning_details(message)
+            if warning_details:
+                print(warning_details)
+            # TODO what's this about?
+        
+        else:
+            logger.warn("Unexpected GstMessage of type %s received: %s"
```
virtuald: Use commas, not % when using a logger.

xl/player/gst/missing_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+
+import logging
+
+from xl.nls import gettext as _
+
+import gi
+from gi.repository import Gst
+gi.require_version('GstPbutils', '1.0')
+from gi.repository import GstPbutils
+GstPbutils.pb_utils_init()
```
virtuald: quod doesn't call this, is it necessary in python?

xl/player/gst/missing_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+@author: Christian Stadelmann
+'''
+
+import logging
+
+from xl.nls import gettext as _
+
+import gi
+from gi.repository import Gst
+gi.require_version('GstPbutils', '1.0')
```
virtuald: This should probably go where the other `require_version`'s are in main.py?

virtuald: I could go either way on that I guess, since nothing else will probably use it.

virtuald:
I would say if exaile can detect that the install has succeeded, that at the minimum it should display a message to that effect. I think there's an InfoBox in the main UI that can be used for this.

If the failing return codes happen, I think once again it should show that message in the infobox (which I believe is what happens now if a failure occurs).

If you think it's bad UI/UX I'm open to changes if you want to make them and they make sense. I don't mind the infobox, and I prefer the infobox to a modal popup.

virtuald:
I don't think it makes sense for Exaile to continue playback after install. Maybe in the 'install succeeded' message it shows something like 'try again'? Does gstreamer have the ability to find the new plugin after the install finishes?

xl/player/gst/missing_plugin.py
genodeftest: Thanks!

xl/player/gst/engine.py
genodeftest: Thanks!

xl/player/gst/missing_plugin.py
genodeftest: Oh, it does not look like it is needed, at least I don't get any different behavior or warnings. I'll remove it.

xl/player/gst/missing_plugin.py
genodeftest: Ok, moving to `main.py`. This way we will see the error earlier, because most people hopefully never run the code I added.

genodeftest:
> If the failing return codes happen, I think once again it should show that message in the infobox (which I believe is what happens now if a failure occurs).

Right now, the code cannot fail (at least on my system) due to the bug in PackageKit. And no, failures are ignored. I did this on purpose, because the user did not trigger the plugin installation procedure directly, so giving him/her any warning or error messages about that will confuse the user. I'd rather not add any error message on that. At least not until I can test them.

> If you think it's bad UI/UX I'm open to changes if you want to make them and they make sense. I don't mind the infobox, and I prefer the infobox to a modal popup.

My comment was not targeted on using the infobox, but on the text itself. I think the text (string) needs improvement.

> I would say if exaile can detect that the install has succeeded, that at the minimum it should display a message to that effect. I think there's an InfoBox in the main UI that can be used for this.

How do I get a handle on that? I'd say there should be a signal (with a handler in xlgui.main) and my code should send an event on that signal. How about that?

> I would say if exaile can detect that the install has succeeded, that at the minimum it should display a message to that effect. I think there's an InfoBox in the main UI that can be used for this.

Yes, sounds like a good idea. I would not resume playback though, because that is not tested well enough and probably never will be.

virtuald:
> How do I get a handle on that? I'd say there should be a signal (with a handler in xlgui.main) and my code should send an event on that signal. How about that?

I would use the same mechanism that errors use, calling up through the player which knows which signal to send.

genodeftest:
> I would use the same mechanism that errors use, calling up through the player which knows which signal to send.

The player only knows how to send an error message, i.e. `ExailePlayer.engine_notify_error()` in xl/player/player.py, which emits a 'playback_error' signal which causes `MainWindow.on_playback_error()` in xlgui/main.py to run. There is no such thing for non-error messages, especially not for questions.

Also, there is no way we can import xlgui/main.py to directly access `main.mainwindow().message`, since that results in circular import dependencies and is _very_ bad style anyway.

I have one other idea, but this will introduce major changes: Add a signal API to xlgui/main.py to make it receive any kind of message xlgui/widgets/dialogs.py:MessageBar supports. Isn't there an easier solution?

> Does gstreamer have the ability to find the new plugin after the install finishes?

Yes, it does, by [Gst.update_registry()](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/gstreamer-Gst.html#gst-update-registry). But the user might have skipped a track and continued playback with another file, so that restarting the GStreamer engine would interrupt playback. I guess this is not what we want. So either we need to ask the user whether we should restart GStreamer or we have to wait until the next pause or stop, which might take _very_ long in gapless mode. This is why I prefer asking.

virtuald:
I'm not opposed to adding a new signal for communicating useful information.

genodeftest:
Ok, but this will take a few days.

genodeftest:
> > I'm not opposed to adding a new signal for communicating useful information.

> Ok, but this will take a few days.

I'd rather not do this in this pull request but at a later time, because it requires reworking much more code in xlgui, which is out of scope for this change.

In my opinion, this pull request is ready to merge. I've rebased it to the latest git master.

genodeftest:
Travis is wrong:
```
tests/xl/player/gst/test_fader.py:5: in <module>

    from xl.player.track_fader import TrackFader, FadeState

xl/player/__init__.py:46: in <module>

    PLAYER = player.ExailePlayer('player')

xl/player/player.py:79: in __init__

    self._setup_engine()

xl/player/player.py:96: in _setup_engine

    from .gst.engine import ExaileGstEngine

xl/player/gst/engine.py:46: in <module>

    from xl.player.gst import missing_plugin

xl/player/gst/missing_plugin.py:101: in <module>

    @common.idle_add()

E   AttributeError: 'NoneType' object has no attribute 'idle_add'
```
but `common.idle_add()` is available at runtime. This issue is probably caused by `common` being definded as `None` in `xl/main.py`.

xl/player/gst/missing_plugin.py
```diff
@@ -0,0 +1,189 @@
...
+
+    engine.stop()
+    __notify_user_on_error(user_message, engine)
+    if GstPbutils.install_plugins_supported():
+        if __run_installer_helper(installer_details):
+            return
+    LOGGER.warn("Installation of GStreamer plugins not supported on this platform.")
+
+
+@common.idle_add()
```
virtuald: Is there a particular reason why you're adding `@idle_add` here? `event.log_event` is safe to call from any thread.

genodeftest: Ok, I will change it. I didn't know that. Are there any rules on what is safe to do and what is not?

genodeftest: Done.

virtuald: The primary constraint I'm aware of are that one shouldn't interact with UI elements directly if you're not on the main UI thread. We do need to document thread safety a little bit more. I'll add a note to xl.event

PR #337, opened 2017-04-26
Plugin moodbar: rework cache filenames, cleanups and fixes

### LINKED ISSUE #297: moodbar: Filename length limit too easily reached on files with Unicode names
When you have a non-Latin filename where each character encodes to multiple UTF-8 bytes, the .mood filename length blows up. For example, a CJK character usually encodes to 3 bytes, each then %-encoded, resulting in each input character taking up 9 characters in the output.

Once this gets too long, the filesystem may reject it:

``` pytb
Exception in thread SpectrumMoodbarGenerator:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/exaile/plugins/moodbar/generator.py", line 94, in generate
    callback(uri, data)
  File "/exaile/plugins/moodbar/__init__.py", line 135, in callback
    cache.put(uri, data)
  File "/exaile/plugins/moodbar/cache.py", line 64, in put
    with open(self._get_cache_path(uri), 'wb') as f:
IOError: [Errno 36] File name too long: '[redacted].mood'
```

genodeftest:
Cache:
* replaced URI-based filenames by hash-based filenames. This fixes #297.
* clean up some code

Generator:
* fixed a traceback occurring if the .mood file could not be created and thus the tmpfile does not exist

Painter:
* Fix array indexing on short tracks, make array indexing more flexible to any kind of track size
* MoodbarPainter: had just one function without usage of self, so removed this class

Widget:
* Fallout from Painter

Formatting:
* made sure that variables are initialized until `__init__()` finishes
* renamed some variables to not overlap with builtin symbols (format)
* renamed some variables to make the name more readable (e.g. `f` for file, `t` for thread)
* replaced old-style classes by new-style classes

I am ready to merge this. Comments?

### DIFF
--- a/plugins/moodbar/cache.py
+++ b/plugins/moodbar/cache.py
@@ -17,17 +17,12 @@
 
 from __future__ import division, print_function, unicode_literals
 
+import hashlib
 import os
-import sys
 
 
-if sys.platform == 'win32':
-    import string
-    invalid_chars = b'*/:<>?\\'
-    SANITIZE_TRANS_TABLE = string.maketrans(invalid_chars, b'-' * len(invalid_chars))
-
-
-class MoodbarCache:
+class MoodbarCache(object):
+    """ Abstract class for a cache for moodbar metadata files """
     def get(self, uri):
         """
         :type uri: bytes
@@ -44,40 +39,33 @@ class MoodbarCache:
 
 
 class ExaileMoodbarCache(MoodbarCache):
+    """ Exaile's implementation of a cache for moodbar files """
+
+    __loc = None
+
     def __init__(self, location):
         try:
             os.mkdir(location)
         except OSError:
             pass
-        self.loc = location
+        self.__loc = location
 
     def get(self, uri):
         try:
-            with open(self._get_cache_path(uri), 'rb') as f:
-                return f.read()
+            with open(self._get_cache_path(uri), 'rb') as cachefile:
+                return cachefile.read()
         except IOError:
             return None
 
     def put(self, uri, data):
         if data is None:
             return
-        with open(self._get_cache_path(uri), 'wb') as f:
-            f.write(data)
+        with open(self._get_cache_path(uri), 'wb') as cachefile:
+            cachefile.write(data)
 
     def _get_cache_path(self, uri):
-        """
-        :type uri: bytes
-        :rtype: bytes
-        """
-        assert isinstance(uri, bytes)
-        if uri.startswith(b'file://'):
-            uri = uri[7:]
-        uri = uri.replace(b"'", b"")
-        if sys.platform == 'win32':
-            uri = uri.translate(SANITIZE_TRANS_TABLE, b'"')
-        else:
-            uri = uri.replace(b'/', b'-')
-        return os.path.join(self.loc, uri + b'.mood')
+        hex_str = hashlib.sha256(uri).hexdigest()
+        return os.path.join(self.__loc, hex_str + b'.mood')
 
 
 # vi: et sts=4 sw=4 tw=99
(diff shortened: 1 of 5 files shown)


virtuald:
This looks fine to me, my only comment is if you're going to use a hash, use sha256 instead of md5.

genodeftest:
> This looks fine to me, my only comment is if you're going to use a hash, use sha256 instead of md5.

May I ask why? I'd understand that for security-relevant code, but this code surely is not related to security at all. Hash collisions might happen, but are **quite** unlikely to happen on just a few thousand songs per Exaile instance.

virtuald:
I haven't tested it, but for the file sizes we deal with, I can't imagine sha256 takes all that much longer to compute than md5. I would be fine with sha1, since that's good enough for git. Google searching indicates that sha1 is only slightly slower than md5 to compute.

I realize this isn't a crypo problem, I just like to avoid the potential for hash collisions. :)

genodeftest:
Done, with `sha256`. It is filenames only anyway, so you are right.

genodeftest:
I also changed floating point division to integer division.

plugins/moodbar/painter.py
```diff
@@ -17,30 +17,29 @@
...
+    surf = cairo.ImageSurface(cairo.FORMAT_RGB24, limit, 1)
+    arr = surf.get_data()
+    for index in xrange(0, limit):
+        index4 = index * 4
+        index3 = index * 3
+        # Cairo RGB24 is BGRX
+        arr[index4 + 0] = data[index3 + 2]
+        arr[index4 + 1] = data[index3 + 1]
+        arr[index4 + 2] = data[index3 + 0]
+    return surf
```
sjohannes: Replace `limit` with `width`.

plugins/moodbar/widget.py
```diff
@@ -20,6 +20,8 @@
 import collections
 
 import gi
+from __builtin__ import int
+from plugins.moodbar import painter
```
sjohannes: Move this to after the other imports.

plugins/moodbar/widget.py
```diff
@@ -117,7 +118,7 @@ def do_draw(self, cr):
         if self.surf:
             # Mood
             cr.save()
-            cr.scale(width / 1000, height)
+            cr.scale(width / (len(self.surf.get_data()) // 4), height)
```
sjohannes: I think that's equivalent to `width / self.surf.get_width()`.

plugins/moodbar/cache.py
```diff
@@ -44,40 +39,33 @@ def put(self, uri, data):
...
-        if uri.startswith(b'file://'):
-            uri = uri[7:]
-        uri = uri.replace(b"'", b"")
-        if sys.platform == 'win32':
-            uri = uri.translate(SANITIZE_TRANS_TABLE, b'"')
-        else:
-            uri = uri.replace(b'/', b'-')
-        return os.path.join(self.loc, uri + b'.mood')
+        hex_str = hashlib.sha256(uri).hexdigest()
+        return os.path.join(self.__loc, hex_str[3:] + b'.mood')
```
genodeftest: Oh, thanks!

genodeftest:
@sjohannes : Thanks, I've included all your suggestions.

sjohannes:
Thanks for this.

The biggest issue with the moodbar plugin right now is that Moodbar itself requires gst0.10.... I wanted to rip out the updated code (which uses gst1) from Clementine but it wasn't working correctly and I haven't had the time to work on it again.

PR #338, opened 2017-04-26
panel notebook: Do not show context menu everywhere

### LINKED ISSUE #291: Panel: Pressing Menu key on a treeview should trigger the treeview's context menu, not the notebook's
1. Select something in a panel treeview.
1. Press the Menu key.

Expected: Context menu for the selected entry shows, just like right-clicking on it.

Actual: The notebook context menu (to select which panels to show/hide) shows.

Less obviously, this also happens when the focus is on a button, though you wouldn't normally press Menu there. It does not happen when the focus is on an entry (the entry's context menu shows fine).

genodeftest:
Before this commit, the context menu of the notebook panel would not only be
shown on notebook tab, but also inside any panel when pressing the "menu" key.

After this commit, the context menu of the notebook panel is not shown any more
when pressing the "menu" key.

Opening the context menu by secondary mouse click is not affected by this commit,
since the code for that is in NotebookTab.

This is an attempt to fix #291.
This change basically reverts 77ca3dcaf7dbf357bf92de8ab8f3dfd9313a51a9.
As a side effect, the menu key does not show the tab panel menu any more at all. The only accessible way to add and remove panels is the main menu now.

### DIFF
--- a/xlgui/widgets/notebook.py
+++ b/xlgui/widgets/notebook.py
@@ -31,7 +31,6 @@ from gi.repository import Pango
 
 from xl.nls import gettext as _
 from xl import providers
-from xlgui import guiutil
 from xlgui.widgets import menu
 
 # Custom tab style; fixes some Adwaita ugliness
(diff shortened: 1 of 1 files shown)


virtuald:
The original menu key stuff was added for accessibility reasons, this seems like a bad thing to remove. What's an example of the bug we're trying to fix? The behavior I've observed seems to be correct as is.

genodeftest:
Ok, let's keep discussing this in #291.

genodeftest:
Is this still of interest? Otherwise it could be closed.

virtuald:
We can revisit it in the future.

PR #349, opened 2017-04-28
splash.ui: Splash screen is no popup window

### LINKED ISSUE #342: Splash screen isn't showing on OSX/Fedora 25
GTK 3.22. On Linux there's a warning (see below). No warning on OSX, but no splash screen either.

```
Gdk-Message: Window 0x560cc8b08320 is a temporary window without parent, application will not be able to position it on screen.
Gdk-Message: Window 0x560cc8b08320 is a temporary window without parent, application will not be able to position it on screen.
```

genodeftest:
This change should fix #342 or at least get around the warnings.

@virtuald : Can you please test under OS X, if this fixes #342?

### DIFF
--- a/data/ui/splash.ui
+++ b/data/ui/splash.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.20.0 -->
 <!--*- mode: xml -*-->
 <interface>
   <requires lib="gtk+" version="3.10"/>
@@ -7,10 +7,8 @@
     <property name="width_request">413</property>
     <property name="height_request">163</property>
     <property name="can_focus">False</property>
-    <property name="type">popup</property>
     <property name="resizable">False</property>
     <property name="window_position">center</property>
-    <property name="destroy_with_parent">True</property>
     <property name="type_hint">splashscreen</property>
     <property name="skip_taskbar_hint">True</property>
     <property name="skip_pager_hint">True</property>


virtuald:
No difference on OSX.

virtuald:
Actually, I take it back, it showed up behind everything.

genodeftest:
So is there a `Gdk.Window.present()` missing?

Btw: I do not see any splash screen, neither with nor without this change. My music library is quite small nowadays, so this might be the cause.

Anyway, this change gets rid of the warnings at startup.

virtuald:
So, this does fix the warning on F25, but as you observed it doesn't actually cause the Window to show. I note that if I get rid of the thing that destroys the window, it eventually does show, so there must be something about showing the splash screen during the start up process that's broken in wayland. There seem to be other reports of this behavior elsewhere.

You can play with this example to see what happens if you're doing a lot of non-ui stuff: it just doesn't show the startup screen:

```
import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk, GLib


win = Gtk.Window()
win.connect("delete-event", Gtk.main_quit)
win.show_all()

import time
for i in range(2):
    
    if win.props.window:
        print("Realized")
        #break
    Gtk.main_iteration()
    time.sleep(3)

print("Done")
Gtk.main()

```

You can see that if you don't let the Gtk loop do it's thing, then you'll never see the Window. I'm not sure there's much we can do there other than initializing everything in a different thread... but that sounds like a fools errand.

genodeftest:
> I'm not sure there's much we can do there other than initializing everything in a different thread... but that sounds like a fools errand.

I actually think it is possible, if all of the initialization that is currently run after the splash screen is shown is being moved into several functions added to the main loop. See my [genodeftest/demo-early-glib-mainloop](https://github.com/genodeftest/exaile/tree/demo-early-glib-mainloop) branch. This is just a techdemo though, I have not tested running exaile without GUI for example.

virtuald:
Well, strictly speaking, as long as you don't load the main GUI, you could do everything else in a thread, right? Then do the rest of the initialization back in the main UI thread.

genodeftest:
> Well, strictly speaking, as long as you don't load the main GUI, you could do everything else in a thread, right? Then do the rest of the initialization back in the main UI thread.

In case we ever want a progress bar in the splash screen, we probably want this anyway.

PR #355, opened 2017-04-29
Add github issue template to guide new bug reports

virtuald:
Let me know if this sounds good, or if it needs adjustment.


sjohannes:
The headers look a bit too big, perhaps h3 or h4 would be better.

I usually put the steps to reproduce before expected and actual behaviours.

It may be useful to tell the user to check the console output and attach any relevant error message.

PR #358, opened 2017-05-01
Some preparations to fix the OSD plugin

genodeftest:
During fixing the OSD plugin, I ran into a few issues I want to clear before committing the changes to the plugin itself. I'm still learning, so depending on your answers to my questions this pull request is not yet ready to merge.

> TrackInfoPane widget: Fix toggling progress bar visibility

No questions.

> preferences widgets: bugfix for TextViewPreference

Did this API change happen from 2.x to 3.x or within the 3.x series? Do we need to have version-specific code? If you know how to get this kind of information, please add a guide to the wiki.

> preferences widgets: Add new RGBAButtonPreference 

Should I store the value as `Gdk.RGBA` so it does not need to be parsed all the time? This will need a change in `xl.settings.TYPE_MAPPING`. Do want that?

There is only one other usage of `ColorButtonPreference` over the Exaile source tree, in `plugins/ipconsole/ipconsoleprefs.py`. I'd fix that too once these changes here are merged.

### DIFF
--- a/xlgui/preferences/widgets.py
+++ b/xlgui/preferences/widgets.py
@@ -793,15 +793,15 @@ class TextViewPreference(Preference):
         buf = self.widget.get_buffer()
         start = buf.get_start_iter()
         end = buf.get_end_iter()
-        return unicode(buf.get_text(start, end), 'utf-8')
+        return unicode(buf.get_text(start, end, True), 'utf-8')
 
     def _set_value(self):
         """
             Sets the value of this widget
         """
         self.widget.get_buffer().set_text(str(
-            self.preferences.settings.get_option(self.name,
-            default=self.default)))
+            self.preferences.settings.get_option(
+                self.name, default=self.default)))
 
     def _get_value(self):
         """
--- a/xlgui/preferences/widgets.py
+++ b/xlgui/preferences/widgets.py
@@ -881,6 +881,8 @@ class IntPreference(FloatPreference):
 class ColorButtonPreference(Preference):
     """
         A class to represent the color button
+        
+        Gdk.Color is deprecated, please use RGBAButtonPreference instead.
     """
     def __init__(self, preferences, widget):
         Preference.__init__(self, preferences, widget)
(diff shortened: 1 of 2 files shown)


sjohannes:
> > preferences widgets: bugfix for TextViewPreference
>
> Did this API change happen from 2.x to 3.x or within the 3.x series?

I assume you're referring to the extra required argument on

```diff
-        return unicode(buf.get_text(start, end), 'utf-8')
+        return unicode(buf.get_text(start, end, True), 'utf-8')
```

The change was not from GTK+ itself ([v2](https://developer.gnome.org/gtk2/stable/GtkTextBuffer.html#gtk-text-buffer-get-text), [v3](https://developer.gnome.org/gtk3/stable/GtkTextBuffer.html#gtk-text-buffer-get-text)) but rather going from [PyGTK](https://developer.gnome.org/pygtk/stable/class-gtktextbuffer.html#method-gtktextbuffer--get-text) (note the `include_hidden_chars=True`) to the [new PyGObject](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/TextBuffer.html#Gtk.TextBuffer.get_text)/PyGI (which doesn't have that default value).

genodeftest:
> I assume you're referring to the extra required argument on

Exactly. Thanks for the explanation! I just wanted to make sure my code will also run on older Gtk+ releases in the 3.x series then what I'm using.

> Looks like settings has a version associated with it: https://github.com/exaile/exaile/blob/master/xl/settings.py#L63, so if you wanted to make an adjustment to type mapping you'd need to bump it I think.

> If the setting isn't used very often, I'd rather use RGBA.parse et al to store/retrieve it, that way we don't have to mess with migrating settings. But if there's a good reason to do so, then I'm ok with doing it.

Ok, then I'll leave it that way. This setting is only used by two plugins, osd and ipconsole, so not worth the risk.

Can you please merge this? I'll add the code in plugin `osd` later on a different pull request.

Issue #346, opened 2017-04-28
Handle removal of deprecated Gtk+ API


genodeftest:
Some parts of the Gtk+ 2.x API have been removed in Gtk+ 3. We have to remove all references to that or it will break Exaile.

- [x] References to `Gdk.Pixmap`
  - [x] affects `plugins/osd/__init__.py`:238 (#360)
  - [x] affects `xlgui/icons.py`:793 (#359)

Old Gdk.Pixmap documentation: https://developer.gnome.org/gdk2/2.24/gdk2-Bitmaps-and-Pixmaps.html

genodeftest:
I'm working on this, but I want to merge #358 first so I know on which base the OSD plugin stands.

genodeftest:
Done in #359, #360.

PR #359, opened 2017-05-01
Remove unused broken function

genodeftest:
This function was broken ever since porting Exaile to Gtk+ 3.
It is not used anywhere, so we may just remove it.

This is part one of fixing #346.


genodeftest:
I made the changes you requested.

PR #360, opened 2017-05-02
plugins/osd: Port to Gtk+ 3, fix bugs, rework

genodeftest:
## Changes
* Stop using Gdk.Pixmap, which has been removed in Gtk+ 3 (See #346)
* Use signals instead of overriding Gtk.Window methods.
  * Works around removal of some Gtk.Window methods
* Make fade out transition smoother
  * 30ms instead of 50ms delay
* Make fade out transition longer
  * 30ms/0.05 = 600ms instead of 50ms/0.1 = 500ms
* Replace the custom `alphacolor` implementation by Gdk.RGBA. Less code to maintain now, yay!
* Replace the custom drawing function by GtkStyleContext and GtkCssProvider
* Disable window shaping code for now
  * Getting the code back requires more work, including bugfixes in PyCairo
  * Known backdraw: The window corners are transparent (not drawn to) but still part of the OSD window. This is a minor usability issue.
* Get rid of the cairo dependency (as a result of other changes)
* Preferences: Make hack for force-showing OSD less ugly
* Preferences: Utilize the new RGBAButtonPreference
* Preferences Glade file: Fix bug where label with pango markup and wrapping enabled would create a very large window
  * Solution: Specify minimum character width so it does not allocate one line per character before being added to the window
* Completely rework the show/hide, fade logic
  * including logic to prevent hiding or fading
  * do not utilize Exceptions that much, rely on good code instead
* Put plugin functions into a OSDPlugin class to
  * allow usage of on_gui_loaded() callback
  * remove global variable OSDWINDOW
* make private fields private
* rename some methods which overlapped with Gtk.Window or Gtk.Widget methods
* remove do_configure_event() method to stop spamming xl.settings with position updates
* some style cleanups according to pep8, pylint et al
  * prevent overlapping names
* make sure to only connect to events after Exaile has started (greatly reduces startup time)
* I could not manage to get resizing code fixed, so I removed it
* Remove old migration code from 2012 since no user will migrate his settings from that long ago
* Add workaround for Gtk+ crasher with opacity: https://bugzilla.gnome.org/show_bug.cgi?id=674449
* Add workaround for a bug in Gtk+ on Linux which will mess up
  * transparent parts of otherwise non-transparent windows
  * fade out transitions
* Use `__future__` division
* Change default OSD color
* Do not specify defaults twice on different locations

## Questions / minor issues
* Text is sometimes selected on start if mouse was pressed down when OSD is shown for the first time. I don't really care about that.
* Should I remove the assertions?
* ~~is opacity still crashing Gtk on Windows? The original code contained a link to this bug report: https://bugs.freedesktop.org/show_bug.cgi?id=54682 stating that any kind of opacity would crash Gtk+ 2.x as shipped with the old Exaile versions on Windows. https://bugzilla.gnome.org/show_bug.cgi?id=674449 indicates this crasher has been present on the Gtk+ 3.x series too. Can someone please test this on Windows or on Mac OS?~~
  * I disabled opacity on windows and also added code to detect whether opacity is supported.
* ~~I could not figure out how to get a pointer position and whether it is inside the OSD (see TODO in `osd.show_first_after_init()`. If you know how, I'd like to fix that.~~
* I decided not to show any errors on playback error, but I could change that if you want.


genodeftest:
Ok, I broke the migration. I'll remove the old migration code from 2012 and place new code there.

plugins/osd/__init__.py
```diff
@@ -117,111 +127,129 @@ def __init__(self):
...
         """
             Cleanups
         """
-        event.remove_callback(self.on_option_set, 'plugin_osd_option_set')
-        event.remove_callback(self.on_track_tags_changed, 'track_tags_changed')
-
+        self.__hide_immediately()
+        event.remove_callback(self.__on_option_set, 'plugin_osd_option_set')
+        if self.__fadeout_id:
+            GLib.source_remove(self.__fadeout_id)
```
virtuald: I know it's being destroyed, but good practice to set __fadeout_id to None

genodeftest: Ok, done.

genodeftest:
Still not ready to merge, I just found another buggy behavior.

genodeftest:
Ok, I'm done now. Migration works too, as `Gdk.RGBA` can also parse old-style `Gdk.Color` strings like `rgb(12, 24, 48)`. (← nice Github feature by the way… )

Stuff I changed since your review:
* Remove old migration code from 2012 since no user will migrate his settings from that long ago
* Add workaround for Gtk+ crasher with opacity: https://bugzilla.gnome.org/show_bug.cgi?id=674449
* Add workaround for a bug in Gtk+ on Linux which will mess up
  * transparent parts of otherwise non-transparent windows
  * fade out transitions
* Use `__future__` division
* Change default OSD color
* Do not specify defaults twice on different locations

virtuald:
It seems to work great on Linux -- with the exception that as you mentioned, I can't actually move the Window. If you can address that, feel free to merge it (I think you can click merge after someone who isn't you approves a PR). 

This is great work so far, thanks!

genodeftest:
Moving the window is, in fact, impossible on Wayland. There is no way an application can position itself on the screen, the wayland protocol does not have a way to do that. This plugin cannot work on wayland.

I'll make a note in the plugin description and warn the user when enabling the plugin. And I'll recommend to use the "notify" plugin instead.

Also, I found a few other issues with the plugin (and in my changes), which I fixed. It will take a few more days before being ready to merge though.

genodeftest:
I'm done with this plugin for now, it is ready to merge imho.

PR #363, opened 2017-05-08
gstreamer windows: Call AvSetMmThreadCharacteristics to boot thread priority

virtuald:
@sjohannes what do you think... I can't think of a better way to inject this functionality. This seems to work on my Windows 7 VM. The only question I have is whether there should be a setting to enable/disable it... or maybe an exception handler in case it fails? But... it shouldn't really fail as far as I can see. Only available on Windows Vista and above, but I think GST requires Win7 anyways..

@dangmai you mentioned you were able to get stutter to happen on your machine at some point? Can you see if you can stimulate your machine to stutter with GST 1.x? I didn't have much luck with my machine. It would be interesting to know whether this branch makes a difference.

It's worth noting that for this to be even more effective, you really need a directsoundsink built with the patches in https://bugzilla.gnome.org/show_bug.cgi?id=773681 applied. I can post a binary somewhere if you want.

### DIFF
--- a/xl/player/gst/engine.py
+++ b/xl/player/gst/engine.py
@@ -39,7 +39,7 @@ from xl.nls import gettext as _
 
 from . import gst_utils
 from .dynamic_sink import DynamicAudioSink
-from .sink import create_device
+from .sink import create_device, priority_boost
 
 from xl.player.engine import ExaileEngine
 from xl.player.track_fader import TrackFader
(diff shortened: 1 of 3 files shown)


sjohannes:
> The only question I have is whether there should be a setting to enable/disable it...

Meh, I don't think that's necessary. If Chrome does it anytime it's playing audio then we should be fine doing it as well.

> or maybe an exception handler in case it fails?

The documentation says the function returns 0 on failure, which your code handles just fine.

virtuald:
One way that I've been able to induce stutter is by changing the power settings on my laptop to throttle the maximum CPU usage to something ridiculous like 5%. Of course, at 5%, even Windows Media Player experiences some glitching, so it's not a great test.

PR #364, opened 2017-05-09
plugins/ipconsole: fix colors, replace deprecated code, clean up

genodeftest:
* ipthemes did not work, so they were removed
* disable opacity on Win32 because it could cause a segfault
* utilize the new RGBAButtonPreference instead of ColorButtonPreference
  * get rid of the deprecated Gdk.Color
* `ipython_view.py`: replace deprecated `os.popen4()`
  * improves compatibility with python 3 where `os.popen4()` has been removed
* code cleanup and imports resorted
* `__init__.py` is not being loaded from terminal
  * it does not need an import hack
* add logging
* put preferences methods into plugin class
* utilize Gtk.StyleContext with CSS replacing deprecated methods from Gtk.Widget
  * `modify_font()`, `modify_base()`, `modify_text()`
  * this fixes coloring for font
* `xlgui/guiutil.py`: add function to convert font description from Pango to CSS
* utilize new css functions in `xlgui/guiutil.py`
* use `__future__` division for python 3 compatibility

`xlgui/preferences/widgets.py`: remove unused ColorButtonPreference
* After ipconsole is merged, there is no use of the ColorButtonPreference any more.
* FontButtonPreference inherited from ColorButtonPreference, but there is no reason to do so any more.

### DIFF
--- a/plugins/ipconsole/__init__.py
+++ b/plugins/ipconsole/__init__.py
@@ -18,32 +18,31 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+from __future__ import division
+
+import logging
 import sys
+import site
+
 from gi.repository import Gdk
 from gi.repository import Gtk
 from gi.repository import GLib
-import ipconsoleprefs
-from xl import settings, providers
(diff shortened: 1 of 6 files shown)


virtuald:
It works for me in Linux. @sjohannes is probably more familiar with this code than I am.

PR #367, opened 2017-05-10
plugins/lyricsviewer: remove deprecations, clean up

genodeftest:
`lyricsviewer.ui`:
* Remove toplevel panel window
* toplevel box becomes a template
* utilize GtkStack to switch refresh_button child widgets

`lyricsviewer_prefs.ui`:
* Remove deprecated "use_action_appearance" property
* use font and size in GtkFontButton

`__init__.py`:
* clean up code, resort `import`s
* reflect changes in .ui file, including port to GtkTemplate
* create plugin class to utilize `on_gui_loaded()`
* remove unused function `modify_textview_look()`
* remove unused function `open_url()`
* integrate `update_source_text()` into `update_lyrics_text()`
* use GtkStyleContext with GtkCssProvider
  * replaces deprecated method `Gtk.Widget.modify_font()`
  * replaces deprecated method `Gtk.Widget.get_default_style()`

`lyricsviewer_prefs.py`:
* clean up code, resort `import`s

### DIFF
--- a/plugins/lyricsviewer/__init__.py
+++ b/plugins/lyricsviewer/__init__.py
@@ -24,265 +24,256 @@
 # do so. If you do not wish to do so, delete this exception statement
 # from your version.
 
-from gi.repository import Gdk
-from gi.repository import Gtk
-from gi.repository import GLib
-from gi.repository import Pango
-
 import os
 import webbrowser
 
-from gi.repository import GdkPixbuf
+from gi.repository import Gtk
+from gi.repository import GLib
+from gi.repository import Pango
 
 from xl.nls import gettext as _
 from xl import (
     common,
     event,
-    lyrics,
     player,
     providers,
-    settings
 )
+from xl import settings as xl_settings
+from xl import lyrics as xl_lyrics
 from xlgui import guiutil
+from xlgui.guiutil import GtkTemplate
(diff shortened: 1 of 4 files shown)


genodeftest:
Also, how about merging lyricsviewer into the xlgui package? Right now, this is a plugin which is being depended on by other plugins – this might fail easily.

virtuald:
That makes sense to me.

genodeftest:
> Were you going to stick this in xlgui? Otherwise, this is good to go, feel free to merge it.

I think so. Is it ok if I get rid of the font preference? this way we can get rid of the lyrics viewer preferences pane. Otherwise I'd have to add some migration code and we'd have a nearly empty preferences pane. I don't know why anybody would want to change the lyrics viewer font anyway.

If you don't support removing the font preference, may I put it on the "appearance" page?

virtuald:
I like having font preferences available when possible, since we really don't have any theming/styling support available at this time.

I think if it's a lot of extra work, let's just keep it as a plugin for now, it's not important enough to mess with it at this time.

genodeftest:
Ok, then keep it as a plugin. That's less work for now.

genodeftest:
> Otherwise, this is good to go, feel free to merge it.

I can't merge it (not authorized). Can you please do that for me? Thanks!

Issue #371, opened 2017-05-15
Improve code quality


genodeftest:
Currently, there are quite many warnings and errors when Exaile's python code is analyzed by `pep8` or `pylint`. I would like to improve the code quality.

### Steps to Reproduce (for bugs)
Run code analysis on Exaile code – or just open the files and read code.

### Expected Behavior
Clean, well-formatted code. According to [doc/dev/code_guidelines.rst](https://github.com/exaile/exaile/blob/master/doc/dev/code_guidelines.rst) the code should comply to pep8

### Current Behavior
Many warnings when running code analysis tools such as pep8 or pylint.

### Possible Solution
[ ] `pep8` and `autopep8`:
- [ ] run `autopep8` over the whole tree
- [ ] run `pep8` and manually fix remaining warnings
- [ ] add a `pep8`-based code checker to continuous integration tests
[ ] `pyflakes`, `pychecker`
[ ] `pylint` – repeat steps above (?)

### Environment
* Exaile version: current git master

### Additional info:
https://flake8.readthedocs.io/en/latest/
https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes

### Discussion:
In [a comment](https://github.com/exaile/exaile/pull/310#issuecomment-267244829), @virtuald voted against enforcing the 80-characters rule too strong. Is there a consensus for that? If yes, we should change the docs.

virtuald:
I think if you're going to take the time to clean it up, then we should add some tests to travis to enforce it on new code too.. otherwise it'll just get messy again?

I'm still very against an 80-character limit. I don't mind sensible code style that helps readability, but sometimes it just gets in the way.

genodeftest:
> I think if you're going to take the time to clean it up, then we should add some tests to travis to enforce it on new code too.. otherwise it'll just get messy again?

That's what I wrote above. I don't know how to do this (yet), but there will be a way. I'll do that after the manual cleanup.

genodeftest:
> I'm still very against an 80-character limit. I don't mind sensible code style that helps readability, but sometimes it just gets in the way.

So, is a 100-character limit ok for you? or 120?

genodeftest:
With pycodestyle (formerly called pep8), I'd like to ignore these "errors" for now:
* E241 – multiple spaces after ‘,’
* E265 – block comment should start with ‘# ‘
* E266 – too many leading ‘#’ for block comment
* E402 – module level import not at top of file
* E501 – line too long (82 > 79 characters)

[...]

PR #372, opened 2017-05-15
Run autopep8 over the whole tree

genodeftest:
This is the first step of #371.

PR #373, opened 2017-05-16
Improve python code quality using pycodestyle

javiercantero:
Use pycodestyle's errors and warnings over the core source tree (`xl/` and `xlgui/` directories) to fix simple errors and improve the general code quality (#371). Only a sensible subset of the checks are used.

### DIFF
--- a/xl/collection.py
+++ b/xl/collection.py
@@ -715,7 +715,7 @@ class Library(object):
                 tr.set_tag_raw('__modified', mtime)
         else:
             tr = trax.Track(uri)
-            if tr._scan_valid == True:
+            if tr._scan_valid:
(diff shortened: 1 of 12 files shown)


virtuald:
Naturally, conflicts occurred with #372 . I assume this is easy enough to redo.

javiercantero:
> I assume this is easy enough to redo.

It is not without a careful review of every conflict, that was precisely the point of my warning: to avoid forcing contributors to spend additional time fixing conflicts they haven't caused (especially because there are a lot of potential conflicts).

I'm not going to waste more time with this, as this PR was an attempt to show a "small incremental patches" based approach but all of them should be already included in the other branch #371.

PR #374, opened 2017-05-19
Autopep8: More automatic fixes

genodeftest:
Some more fixes I forgot to add to #372 in time.

Not important as they can easily be redone.

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -222,7 +222,7 @@ class Track(object):
                 state. not for normal use.
         """
         # don't re-init if its a reused track. see __new__
-        if self._init == False:
+        if self._init is False:
             return
 
         self.__tags = {}
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -485,7 +485,7 @@ class Track(object):
 
     def set_tag_raw(self, tag, values, notify_changed=True):
         """
-            Set the raw value of a tag. 
+            Set the raw value of a tag.
 
             :param tag: The name of the tag to set.
             :param values: The value or values to set the tag to.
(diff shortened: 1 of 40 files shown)


xl/trax/track.py
```diff
@@ -222,7 +222,7 @@ def __init__(self, uri=None, scan=True, _unpickles=None):
                 state. not for normal use.
         """
         # don't re-init if its a reused track. see __new__
-        if self._init == False:
+        if not self._init:
```
javiercantero: I like using `is False` in some instances like this one only because it's more readable.

xlgui/main.py
```diff
@@ -1074,7 +1074,7 @@ def window_state_change_event(self, window, event):
             # else
             #    destroy tray
 
-            if self.minimized != prev_minimized and self.minimized == True:
+            if self.minimized != prev_minimized and self.minimized:
```
javiercantero: Another instance where the use of is (`is True`) could improve the readibility.

xl/trax/track.py
genodeftest: Yes, that's true. Your solution has better readability than the pep8-suggested default. Will fix it.

xlgui/main.py
virtuald: I prefer the implicit comparison myself, but either way is fine.

xl/trax/track.py
virtuald: If you fix it, I'll merge, everything else looks good.

xlgui/main.py
javiercantero: The reason I prefer the explicit comparison here is because the expression `self.minimized` is already in the first part of the condition (before the `and`), so the instance in the second part is like something is lost after that. Anyway, it's only a matter of personal taste/style.

genodeftest:
I've included the changes you requested.

PR #385, opened 2017-05-25
Sync spydaap upstream

genodeftest:
This change is a huge step towards bringing our local copy of spydaap and upstream closer together (see also: egh/spydaap#3 and egh/spydaap#4). Based on this, it should be possible to unbundle our copy of spydaap, but that will be subject to a separate change.

I don't own an Apple device (and don't want to), so I could not check this code thoroughly. Is someone here who can test it?


virtuald:
I have an apple computer, but I'm not really familiar with daap/iTunes... honestly, if it seems to run and doesn't crash anything, I think that's good enough until we have someone use it. We can't support everything.

There's a server and a client, right? I expect if we could prove that two difference instance of Exaile could talk to each other using those plugins, then that's good enough?

afs2015:
I have a Mac, to clarify you just need someone to run Exaile and make sure it doesn't crash Apple device?

genodeftest:
> I have a Mac, to clarify you just need someone to run Exaile and make sure it doesn't crash Apple device?

No. I'd need someone to try to transfer files between a daap client and a daap server, where Exaile can be either one of them, and iTunes the other one. You need to enable sharing your music over the network in iTunes to enable the server. But since I don't use iTunes I have no clue how to do so.

genodeftest:
Current state of this pull request (tested with 2 exaile instances, one acting as server, one as client):
- [x] Connect client to server
- [x] Fetch metadata (whole collection) over daap
  - [ ] Encoding issues on the server with file names. What is the default encoding in DAAP? As a result, these files cannot be displayed to the client and thus never be played.
- [x] Playback of all correctly shown files works fine
  - [ ] When scrolling on client in long audio files, the server throws a `broken pipe` Exception in a separate thread. It does not look like any of exaile's code is involved.
- [x] Server can shut down without any warning printed
  - [ ] Except when a client is still connected
- [x] Client can shut down and disconnect cleanly from server
  - [ ] Client does not break when server disappears

genodeftest:
Remaining unclear:
* Is it ok to encode filenames to UTF-8? http://daap.sourceforge.net/docs/content-codes.html says so. Since filenames never appear in HTTP GET requests that might be ok.

genodeftest:
I'm putting the client-related stuff into a new pull request later on, so this one can be finished. Can someone please review the changes made above?

genodeftest:
> [ ]  Can jump time on file playback

Seeking on files played from DAAP shares has to be implemented in xl.* as there is no daapclient-specific code used when playing any file over daap, instead it just uses the URL.

genodeftest:
@afs2015 
> I have a Mac, to clarify you just need someone to run Exaile and make sure it doesn't crash Apple device?

Now would be a nice time to re-test. Could you please run Exaile and your Mac at the same time and connect them to each other, i.e. following these steps:
1. start both applications
2. in Exaile, enable the daapserver plugin
3. in iTunes, connect to Exaile's daap share
4. play a file from iTunes
and please report back if you run into any issues.

virtuald:
If you think it's good to go, then let's merge it. There's a lot of code here, I don't think I'd be able to give it a proper review right now.

genodeftest:
> If you think it's good to go, then let's merge it.

I am. Let's merge.

PR #394, opened 2017-05-28
Enhance accelerators

virtuald:
* Menu items can pull its callback/label data from an accelerator instead of typing it twice
* Adds a dialog showing all existing shortcuts

### DIFF
--- a/xlgui/widgets/dialogs.py
+++ b/xlgui/widgets/dialogs.py
@@ -139,6 +139,36 @@ class AboutDialog(Gtk.AboutDialog):
         self.destroy()
 
 
+@GtkTemplate('ui', 'shortcuts_dialog.ui')
+class ShortcutsDialog(Gtk.Dialog):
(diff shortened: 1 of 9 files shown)


xlgui/widgets/dialogs.py
```diff
@@ -139,6 +139,35 @@ def on_response(self, *_):
...
+    
+    __gtype_name__ = 'ShortcutsDialog'
+    
+    def __init__(self, parent=None):
+        Gtk.Dialog.__init__(self)
+        self.init_template()
+        
+        self.set_transient_for(parent)
+        
+        for a in sorted(providers.get('mainwindow-accelerators'), key=lambda a: a.name[-1] + a.name):
```
sjohannes: Suggestion: `key=lambda a: (a.key, a.mods)` so it works for things like `F1`, `Delete`, etc.

Issue #319, opened 2016-12-20
NotebookTab theming and orientation issues


genodeftest:
From [the wiki page](https://github.com/exaile/exaile/wiki/PyGI-migration-tasklist-(GTK3-GStreamer-1.x)):

> If _Prefs > Appearance > Tab placement_ is Left or Right, it looks terrible.
> We force 6px padding to deal with Adwaita's massive padding, but 6px is actually larger than normal on some other themes. Hmm....

There are two issues here:
# Theming
We should not apply Adwaita theme fixes to non-Adwaita themes. Also, the Adwaita theme fixes are only working for `Gtk.PositionType.LEFT` tabs, but not for `RIGHT`.

Solutions:
1. It is possible to apply our theming fixes only if the theme is Adwaita and let all the other themes handle it however they like. A fix for this is in my branch.
2. I also wrote a fix for `Gtk.PositionType.RIGHT`. Compatible with the other solution.

~# Widget orientation~
~When setting `Gtk.PositionType.LEFT` or `RIGHT`, the tab widgets (label, etc.) are still oriented horizontally.~

~Some options here:~
1. ~disable changing tab placement. Remove a feature to simplify code. Panel tabs still need to have custom theming, but my fix won't be necessary.~
2. ~In my branch, the tab widget is also rotated, including the label. This solution is not perfect now, because~
    * ~images are not rotated, which looks weird. But I'm not sure it will get better when being rotated. do we want them to rotate?~
    * ~renaming the tab gets __very__ ugly now, because you can't easily rotate a `Gtk.Entry` and you probably shouldn't anyway. How about replacing it by a `Gtk.Popover`? Note that this will increase the version requirement on Gtk+ to 3.12. I think we can do that, because [Debian stable (8.0 Jessie) is at 3.14](https://distrowatch.com/dwres.php?firstlist=ubuntu&secondlist=debian&firstversions=6&secondversions=2&resource=compare-packages&view=major&refresh=Refresh) and CentOS too, but we'll lose Ubuntu 14.04 LTS compatibility. You'll need to decide whether you want that.~

As one can see below, @sjohannes didn't really want to rotate the tabs other 

My branch: https://github.com/genodeftest/exaile/tree/notebooktab

sjohannes:
Regarding the orientation, there should be a difference between the panel-style layout (what the current code calls "vertical") where the items are laid out vertically, and the tracklist-style layout where they are laid out horizontally (even when the tab bar is on the left).

In ASCII art form:

[...]

PR #396, opened 2017-05-29
xlgui/notebook: Fix appearance of left/right tracklist tab labels.

sjohannes:
Left/right tab labels should:

* Not expand vertically.
* Have a larger minimum width.

See: https://github.com/exaile/exaile/issues/319


virtuald:
Looks good, but it only fixes the top view, not any additional playlist views.

virtuald:
Oh weird. It works if you start with multiple pages, but it seems like removing all of the pages causes it to break?

Issue #386, opened 2017-05-25
xl.trax.trackdb fails to save database


genodeftest:
### Steps to Reproduce (for bugs)
1. use the new 4.0.0-beta1 installer
2. start Exaile
3. have a look at the log file

### Expected Behavior
No tracebacks. No errors.

### Current Behavior
Upon opening the database (with 'c' flag for creating it if not present), I get a traceback:
```
18:56:52 INFO    : Loading Exaile 4.0.0beta1... (xl.main)
18:56:52 INFO    : Using Python 2.7.13 (xl.version)
18:56:52 INFO    : Using PyGObject 3.24.1 (xl.version)
18:56:52 INFO    : Loading settings... (xl.main)
18:56:52 INFO    : Using Locale de_DE cp1252 (xl.version)
18:56:52 INFO    : Using Mutagen 1.37 (xl.version)
18:56:54 INFO    : Using GStreamer 1.10.4 (xl.version)
18:56:54 INFO    : Reconfiguring crossfading (xl.player.gst.engine [player])
18:56:54 INFO    : Crossfade: disabled (xl.player.gst.engine [player])
18:56:54 INFO    : Using GTK+ 3.22.15 (xl.version)
18:56:55 INFO    : Loading plugins... (xl.main)
18:56:56 INFO    : Loading collection... (xl.main)
18:56:56 ERROR   : Failed to open music DB. (xl.trax.trackdb)
Traceback (most recent call last):
  File "C:/msys32/home/IEUser/exaile/tools/installer/_inst/usr/lib/exaile/xl/trax/trackdb.py", line 186, in load_from_location
  File "C:/msys32/home/IEUser/exaile/tools/installer/_build_root/mingw32/lib/python2.7/shelve.py", line 243, in open
  File "C:/msys32/home/IEUser/exaile/tools/installer/_build_root/mingw32/lib/python2.7/shelve.py", line 227, in __init__
  File "C:/msys32/home/IEUser/exaile/tools/installer/_build_root/mingw32/lib/python2.7/anydbm.py", line 82, in open
error: db type could not be determined
```

### Possible Solution
No clue.

### Environment
* Operating System and version: Windows 10
* Exaile Version: 4.0.0-beta1 from [official installer](https://github.com/exaile/exaile/releases/tag/4.0.0-beta1)

genodeftest:
This issue has been reported to python upstream developers: https://bugs.python.org/issue23174. They claim it is not a bug, but the behavior differs from the docs:

[...]

PR #398, opened 2017-06-01
trackdb: Insist on creating new DBs using bsddb, not any other dbm backend

sjohannes:
I think the gdbm module in MSYS2 is [broken](https://github.com/exaile/exaile/issues/386), so this completely sidesteps it.

On other platforms bsddb also has advantages. [The only benchmark I can find](http://tokyocabinet.sourceforge.net/benchmark.pdf) says Berkeley DB is marginally better than GDBM. ~~It's also more likely to already be installed (only talking about the library, not the Python module), according to a quick reverse dependency search on Debian and Arch~~ (**update:** this is wrong; lots of things require gdbm as well).

If the user really really wants to use gdbm or any of the other dbm backends, they can create their own music.db with that backend; when the bsddb part of this patch fails, it falls back to using `shelve.open`.

### DIFF
--- a/xl/trax/trackdb.py
+++ b/xl/trax/trackdb.py
@@ -181,13 +181,23 @@ class TrackDB(object):
         logger.debug("Loading %s DB from %s.", self.name, location)
 
         try:
+            # New DBs will only be created as Berkeley DB Hash databases using
+            # either bsddb3 (external) or bsddb (stdlib but sometimes removed).
+            # Existing DBs created with other backends will continue to work.
+            # We do this because BDB is generally considered more performant,
+            # and because gdbm currently doesn't work at all in MSYS2.
             try:
-                pdata = shelve.open(location, flag='c',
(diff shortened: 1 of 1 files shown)


sjohannes:
Note that if your music.db is already created broken (empty file), you need to delete it first.

virtuald:
Hm. The way I was thinking about this, it would try to automatically upgrade you to bsddb3 if it detected a different type.

Also, the lyrics cache needs to be updated also. I was thinking maybe we move this to a utility class?

genodeftest:
> Hm. The way I was thinking about this, it would try to automatically upgrade you to bsddb3 if it detected a different type.

+1, preferably with a new name to make database type detection easier. (migration).

Still, the current solution in 6cfd5fe fixes the core issue. The migration code can be added later.

virtuald:
I'm going to steal this code and move it into xl.common, and open a new PR in a few minutes.

PR #399, opened 2017-06-02
Change trackdb to force bsddb

virtuald:
- Migrates old data when opened

Proposed as an alternative to #398


xl/common.py
```diff
@@ -376,6 +390,52 @@ def open_file_directory(path_or_uri):
         subprocess.Popen(["xdg-open", f.get_parent().get_parse_name()])
 
 
+def open_shelf(path):
+    '''
+        Opens a python shelf file, used to store various types of metadata
+    '''
+    # New DBs will only be created as Berkeley DB Hash databases using
+    # either bsddb3 (external) or bsddb (stdlib but sometimes removed).
+    # Existing DBs created with other backends will continue to work.
```
sjohannes: s/continue to work/be migrated/

xl/migrations/database/to_bsddb.py
```diff
@@ -0,0 +1,63 @@
...
+    new_shelf = shelve.BsdDbShelf(db, protocol=common.PICKLE_PROTOCOL)
+    
+    for k,v in old_shelf.iteritems():
+        new_shelf[k] = v
+    
+    new_shelf.close()
+    old_shelf.close()
+    
+    try:
+        common.replace_file(bak_path, path)
```
sjohannes: The fun thing here is that some dbm-style backends (I think dbm and dumbdbm) create multiple files (`music.db` is stored as `music.db.*`), and this would leave the old files behind.

sjohannes: For reference:

* dbm:
  * When compiled against ndbm or gdbm (e.g. in Arch and Fedora), it creates `music.db.dir` and `music.db.pag`.
  * When compiled against Berkeley DB (e.g. in Debian), it creates `music.db.db` (note the extra `.db`).
* dumbdbm creates `music.db.bak`, `music.db.dat` and `music.db.dir`.

`dbm` is only used by `shelve.open` if neither `bsddb` nor `gdbm` can be imported. If that also fails, `dumbdbm` is used.

virtuald: Hm. So maybe we should copy music.db.*

virtuald: Fun side effect that occurs here: if there isn't a `music.db` file (such as in the case of dumbdbm), then forcing an open with bsddb will just create a new file because it won't notice another file present.

sjohannes: Oh right, that didn't even occur to me. Hm....

The dbm and dumbdbm cases can only happen if the user doesn't have either the bsddb or gdbm module. I really don't know how common that is. If it's rare, we could just add something to the release notes about what to do if the collection disappears on upgrading.

virtuald: I'm sure it's not super common. I think the way to deal with it is by checking if the path we asked for exists; if it doesn't (only true in the case of migration needed) then use whichdb to double check that it's not one of the weird ones. If whichdb finds something, then we perform a migration and clear the old files.

virtuald: Note: whichdb has a bunch of checks to see if any of the files exist.

virtuald: That approach seems to work; I'm going to add some tests just to make sure I didn't miss anything.

virtuald: Heh. Seems like the shelve module is a bit broken. That's annoying. Won't be able to look into this more until Wednesday I suspect.

virtuald: The exaile testimg is running Ubuntu 16.04... maybe that means we should create test images for other OS too just for completeness.

PR #402, opened 2017-06-23
Plugins: bug fixes and cleanups

genodeftest:


plugins/bookmarks/__init__.py
```diff
@@ -244,84 +230,73 @@ def delete_bookmark(self, widget, targets):
 
         self.save_db()
 
+    @common.idle_add()
```
virtuald: Why is this idle_add?

genodeftest: > Why is this idle_add?

Because it does I/O operations which might block, e.g. during swapping/paging or because of disk suspend.

genodeftest: > Because it does I/O operations which might block, e.g. during swapping/paging or because of disk suspend.

Maybe it should be threaded instead…. Does python have async I/O? Or should I use GLib's GIO instead? I've read that the latter does most stuff async so it should be safe here.
If you are fine with the other changes, I could cherry-pick them on master and fix this one only.

virtuald: idle_add only ensures that something runs on the main UI thread -- so in fact by using it you will ensure that the main thread is blocked. :)

The GIO async stuff requires you use the _async version of whatever calls that exist.

`@threaded` is what you want here for now. Using the GIO async stuff is a bit awkward, and to do it well we really need to wrap it with coroutines of some kind...

genodeftest: Ok, then I'll put it into @threaded.

genodeftest: I'm sorry it took so long.

> `@threaded` is what you want here for now. Using the GIO async stuff is a bit awkward, and to do it well we really need to wrap it with coroutines of some kind...

Done.

PR #404, opened 2017-06-24
Plugins: merge notifyosd into notify, rework notify

genodeftest:
> Both plugins were quite outdated, especially notify was using non-existent API (attaching the notification to a widget or tray icon).
> 
> Keep some settings from notifyosd and merge them into notify.
> 
> Custom format strings have been removed because they are too hard on the user. If a user wishes to change them, he might also change the code.
> 
> Carefully handle the case that no notify-enabled server is available, which will otherwise freeze Exaile for >10 seconds. Use threads very carefully to prevent race conditions.
> 
> Take more care of the Notify API, handle special cases like missing support for markup. Activate Exaile when clicking on the notification.
> Remove overlapping signal handlers.
> 
> Port plugin to plugin class

Some remaining issues (which should not block the merge, but if you have an idea how to fix them, I'd like to do so; marked as TODO in code):
* Even with `resize_covers=False`, icons don't get larger than 48px, which is very small.
* We could add more actions later.
* When enabling the plugin fails, the user should get notified. API is missing to do so.


plugins/notifyosd/PLUGININFO
```diff
@@ -1,6 +0,0 @@
-Version='4.0.0'
-Authors=['Adam Olsen <arolsen@gmail.com>', 'Steve Dodier <sidnioulz@gmail.com>']
-Name=_('Notify-osd notifications')
-Description=_('This plugins displays notification bubbles when a song is played/resumed/stopped, with either the song cover or a media icon to indicate the latest action.\n\nDepends: libnotify and its typelib file\nRecommends: notify-osd')
-Category=_('Notifications')
-RequiredModules=['gi:Notify']
```
sjohannes: Merge into notify/PLUGININFO?

plugins/notify/__init__.py
```diff
@@ -1,4 +1,8 @@
-# Copyright (C) 2009-2010 Abhishek Mukherjee <abhishek.mukher.g@gmail.com>
+# Copyright (C) 2009-2010
+# Adam Olsen <arolsen@gmail.com>
+# Abhishek Mukherjee <abhishek.mukher.g@gmail.com>
+# Steve Dodier <sidnioulzg@gmail.com>
+# Christian Stadelmann
```
genodeftest: Ok, will do that.

plugins/notify/__init__.py
```diff
@@ -14,124 +18,335 @@
...
+        self.__old_icon = None
+        self.__tray_connection = None
+
+        notification = Notify.Notification.new("", None, None)
+
+        if 'sound' in caps:
+            notification.set_hint('suppress-sound',
+                                  GLib.Variant.new_boolean(True))
+        # notification.set_hint('suppress-sound', True)
+        self.settings.can_show_markup = 'body-markup' in caps
```
genodeftest: This is the place where I detect whether the notification server supports markup or not.

plugins/notify/__init__.py
```diff
@@ -14,124 +18,335 @@
...
-
-        if artist and album:
-            body_format = self.body_artistalbum
-        elif artist:
-            body_format = self.body_artist
-        elif album:
-            body_format = self.body_album
+        album_str = cgi.escape(track.get_tag_display('album'))
+
+        if self.settings.can_show_markup:
```
genodeftest: > I can't remember why we had two plugins for this, but I sort of remember hearing that some notification daemons (XFCE's maybe?) don't support formatting? I don't know whether that's still a valid issue today.

This is the place where I change the notification text body depending on whether markup is supported or not.

plugins/notifyosd/PLUGININFO
genodeftest: Done in d02e201.

plugins/notify/__init__.py
genodeftest: Done in d02e201.

genodeftest:
> Haven't tested this (on Windows at the moment), but other than the plugin credits the patch looks fine.

I tested it on Windows and it fails correctly, disabling itself immediately. This is the same behavior as on e.g. Weston or an Openbox desktop without a notify server running. There is no dbus server on Windows by default, and even if it was there it would miss a libnotify server.  The situation has been the same even before the [BREAK ALL THE THINGS!](https://github.com/exaile/exaile/commit/f2651a22b224078f8e022a218457d72c048c19f8) commit, i.e. before porting to Gtk+ 3.x, because `pynotify` used in 3.3.0 to 3.4.x is just another python binding to libnotify.

So no, if it ever worked on windows, that  was at least 7 years ago and you always needed extra software installed.

virtuald:
The notifications are inconsistent for me on gnome shell... there are some files that it works on, and some that don't work...

genodeftest:
> The notifications are inconsistent for me on gnome shell... there are some files that it works on, and some that don't work...

Any logging output?

genodeftest:
> The notifications are inconsistent for me on gnome shell... there are some files that it works on, and some that don't work...

@virtuald Any logging output?

virtuald:
There wasn't any logging output, and I'm in the middle of a housing move, so I don't have my normal computer available to me at this time. Soon though.

genodeftest:
> The notifications are inconsistent for me on gnome shell... there are some files that it works on, and some that don't work...

Actually, now I can reproduce. No logging output.

genodeftest:
> The notifications are inconsistent for me on gnome shell... there are some files that it works on, and some that don't work...

I guess this is not a bug, but rather the fact that the plugin only shows notifications by default if Exaile's window is hidden. See the preference "ShowWhenFocused" i.e. `plugin/notify/show_when_focused` for details. It defaults to `False`.

Issue #347, opened 2017-04-28
Fix some Gtk+ Deprecations, improve Gtk+-Related code


genodeftest:
Some stuff changed in Gtk+, which requires changes in Exaile:
- [ ] `Gtk.Widget.popup()` is deprecated. Use `Gtk.Widget.popup_at_[…]()` instead.
- [ ] Do not create any `Gtk.Window` or any class inherting from it (most notably `Gtk.Dialog`) without a parent.
- [ ] `Gdk.Color` is deprecated. 
- [ ] `Gtk.ImageMenuItem` is deprecated. Unclear: How do we replace it? Maybe wait for a while until upstream figures out a replacement?
- [ ] `Gtk.Arrow` is deprecated, use `Gtk.Image` with an icon instead.
- [ ] `Gtk.StatusIcon` is deprecated. Gtk+ developers intend to provide no replacement. Make sure we use [MPRIS](https://standards.freedesktop.org/mpris-spec/latest/) good enough and that we show enough notifications.
- [ ] `Gtk.Stock` is deprecated, including `Gtk.StockItem`.
- [ ] Major parts of the `Gtk.Icon[…]` API are deprecated.
- [ ] `Gtk.Window.resize()` is not deprecated, but its use is discouraged and will result in ever-shrinking or ever-growing windows. Use `Gtk.Window.set_default_size()` instead.
- [ ] Use modern signals API: most usages of `button-press-event`, `button-release-event`, `key-press-event` or `key-release-event` are overly complicated custom build solutions and should be replaced by the `activated` signal, the `popup-menu` signal or any of the `drag-` signals. This will also improve touch screen support a little. Make sure to use touch events too!
- [ ] [Test for modifier keys correctly](https://developer.gnome.org/gtk3/stable/checklist-modifiers.html)

virtuald:
The popup_at_* APIs are introduced in GTK 3.22. We'll wait a little while before using them, otherwise we require 3.22... which, I'm not sure what we actually require, but it's older than that.

genodeftest:
> The popup_at_* APIs are introduced in GTK 3.22. We'll wait a little while before using them, otherwise we require 3.22... which, I'm not sure what we actually require, but it's older than that.

Thanks, I've updated the description for that.

sjohannes:
> Gtk.ImageMenuItem

Does it even still work? On Win32 (msys2) I already don't have any menu icons.

[...]

PR #405, opened 2017-06-25
xlgui/panel: Allow panels without a parent window

genodeftest:
Commit message:

> Before this change, all panels are created with a parent window and are being reparented when loaded. This is unnecessary code.
> 
> This change introduces a simple way to allow panels being created without a window. All known panels have been ported to the new feature.
> 
> Also, some panels have been slightly redesigned, reducing the widget count.

This is part of #347.


virtuald:
This seems awesome, but I've been on vacation/busy and won't be able to review anything until next week at the earliest.

PR #411, opened 2017-07-01
Playlist: on_{shuffle,repeat}_mode_set: apply mode only if widget is active

### LINKED ISSUE #410: Playlist: closing shuffle or repeat popup resets its setting to disabled
I have been using the gtk3-enabled git master for a while now, and have observed the following minor annoyance, which is also present in 4.0.0-beta2: whenever I open the playlist's shuffle or repeat popup to check its setting, and then close the popup without selecting anything (e.g., pressing ESC, clicking elsewhere), the corresponding setting gets reset back to 'disabled'.

It would seem that the GtkRadioMenuItem emits the 'activate' signal not only when it becomes checked (active), but on every state change. Furthermore, when the popup is created, the first GtkRadioMenuItem in its group (= 'disabled') is automatically checked (set as active). Therefore, if any other item is checked afterwards due to the current setting, the first item will become unchecked and will emit 'activate' signal, hence resetting the setting. The 'activate' signal from the other item is not connected to the callback yet, so the proper setting is never recovered.

Therefore, the on_shuffle_mode_set() and on_repeat_mode_set() callbacks should check the widget's state before applying the mode. (Alternatively, connecting the item's callback before setting its state during popup construction would probably do the trick as well).

rokm:
The "activate" signal is emitted also when the GtkRadioMenuItem becomes unchecked, so the callbacks need to check the state of the widget before applying the mode.

Fixes #410.

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -577,13 +577,15 @@ class PlaylistPage(PlaylistPageBase):
         """
             Callback for the Shuffle mode menu
         """
-        self.playlist.shuffle_mode = mode
+        if widget.get_active():
+            self.playlist.shuffle_mode = mode
 
     def on_repeat_mode_set(self, widget, mode):
         """
             Callback for the Repeat mode menu
         """
-        self.playlist.repeat_mode = mode
+        if widget.get_active():
+            self.playlist.repeat_mode = mode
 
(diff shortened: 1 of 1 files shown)


genodeftest:
Looks good, can be merged if @sjohannes or @virtuald agree with changing the behavior.

virtuald:
Thanks!

PR #416, opened 2017-07-20
Add support for MATE and Cinnamon screensavers to Pause on Screensaver plugin

twa022:
The plugin currently only supports GNOME and KDE screensavers, but can support MATE and Cinnamon screensavers.


virtuald:
Thanks!

PR #421, opened 2017-07-29
Lyrics viewer: Move plugin code into core xlgui API

genodeftest:
Also, make some attributes private.

This has been discussed in https://github.com/exaile/exaile/pull/367#issuecomment-300592964 before and the idea was ACKed by @virtuald.


genodeftest:
I have not yet migrated the preferences. Should I do that or are those old paths ok?

xl/plugins.py
```diff
@@ -144,6 +144,8 @@ def uninstall_plugin(self, pluginname):
         return False
 
     def enable_plugin(self, pluginname):
+        if pluginname == 'lyricsviewer':
```
virtuald: Hm. I don't really like having a special case in here. Why not just let it fail?

genodeftest: Let's make it fail with an exception then. I wanted to have no exception printed to output/logfile in a case like this when the behaviour was expected, but I don't care much.

PS: I pushed a new commit, which is identical to the old one except for missing these two lines.

PR #427, opened 2017-09-12
Fix shuffle playback missing last track in playlist

### LINKED ISSUE #247: repeat+shuffle: Last track in playlist is not played again any more
Steps to reproduce:
1. create a playlist with at least 2 items, 3 is ideal to show this bug
2. set shuffle mode to "shuffle tracks"
3. set repeat mode to "repeat all"
4. play
5. wait some time or press "next" button

What happens:
Last track in playlist is only played once, no matter how often you press "next".

What should happen:
Ideally (when pressing "next" very often) every track should be played (roughly) the same number of times.

Affects: current git master since at least 5 months

genodeftest:
Prior to this commit, when a playlist had "shuffle tracks" and "repeat all"
enabled, it would not play the last track any more after playing all tracks once.

Thanks to @sbrubes for the hint!

Fixes #247

Please check for unknown side-effects before merging.

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -1125,7 +1125,7 @@ class Playlist(object):
         next_index = -1
 
         if shuffle_mode != 'disabled':
-            if self.current is not None:
+            if current_position != -1:
                 self.__tracks.set_meta_key(current_position,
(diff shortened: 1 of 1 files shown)


virtuald:
Works for me.

PR #445, opened 2017-10-31
[WIP] unbundle spydaap (plugins: daapserver, daapclient)

genodeftest:
Spydaap is managed upstream well, so remove it from our source tree. Also remove the copies of spydaap files in daapclient plugin.

Note: You need to run `make clean` on your local clone after updating beyond this pull request and delete `plugins/daapserver/spydaap` and `plugins/daapclient/spydaap` because git is unable to remove remaining .pyo/.pyc files in those directories. `git reset --hard HEAD` is not enough, neither is `git clean -dfX`.

TODO items:
- [ ] Test build scripts (especially for windows)

__Edit:__
This pull request previously tried to use git submodules, which we now decided against.


virtuald:
I'm not a fan of adding a git submodule.

virtuald:
Let's convince upstream to push it to pip (filed an issue at https://github.com/egh/spydaap/issues/6), otherwise we should fork the repo, and publish to pip for them. I don't want to have to deal with submodules if we don't have to.

genodeftest:
Hm, I don't really like the complexity coming with git submodules either. Pip would be a usable alternative. Should I a new make target to download the files then?

virtuald:
If it were a pip package, then you would just need to import it -- the end user would have to install it separately.

virtuald:
I haven't heard from the author, so I think we should just push our own spydaap pypi package, and relinquish it to them if they ask for it.

genodeftest:
Spydaap is managed upstream well, so remove it from our source tree. Also remove the copies of spydaap files in daapclient plugin.

Note: You need to run `make clean` on your local clone after updating beyond this pull request and delete `plugins/daapserver/spydaap` and `plugins/daapclient/spydaap` because git is unable to remove remaining .pyo/.pyc files in those directories. `git reset --hard HEAD` is not enough, neither is `git clean -dfX`.

TODO items:
- [ ] Test build scripts (especially for windows)

__Edit:__
I deleted all the comments here which were referencing the old approach to use git submodules.

genodeftest:
@virtuald : Can you please test whether this breaks windows builds? I cannot do that on my machine.

virtuald:
If we configure our windows build to install spydaap from pip, then I think we'll be fine.

Unfortunately, we still have the pybonjour pip problem. I think I'll fork that package also and have spydaap depend on the forked version instead.

genodeftest:
> Unfortunately, we still have the pybonjour pip problem.

This problem is related, but not getting introduced or getting worse by my commits, is that correct?

> I think I'll fork that package also and have spydaap depend on the forked version instead.

Or we'd make spydaap use [zeroconf](https://pypi.python.org/pypi/zeroconf/), which is another implementation and seems to be actively developed on ([GitHub repo](https://github.com/jstasiak/python-zeroconf)). It seems they are dropping Python 2 support (see [this commit](https://github.com/jstasiak/python-zeroconf/commit/f22f421e4e6bf1ca7671b1eb540ba09fbf1e04b1)) but maybe the python 3 version will be around long enough to get Exaile with spydaap ported to Python 3.

virtuald:
Where did we leave off on this? I think you're waiting for me for something here...

genodeftest:
Let's continue the pybonjour issue at egh/spydaap#10.

virtuald:
@genodeftest I rebased and force pushed to your branch

I tried it out on Windows, and the daapclient fails because it currently requires dbus (which doesn't exist on Windows). The python-zeroconf has a dependency on netifaces, which requires adding gcc to our set of build dependencies. Not a big deal, but there's something about our pyinstaller configuration which isn't bringing in spydaap, we probably need to tell pyinstaller to scan our plugins for dependencies.

I vote let's just merge this and be done with it. If someone really wants to use it on Windows, we can address it in the future. There are a lot of plugins that just don't work on Windows and that's how it'll be for now.

genodeftest:
> I vote let's just merge this and be done with it. If someone really wants to use it on Windows, we can address it in the future. There are a lot of plugins that just don't work on Windows and that's how it'll be for now.

+1. Let's not block the 4.0 release any longer by this probably barely used plugin. We may fix it later.

You may want to open a new issue for the build issues for windows.

PR #449, opened 2017-11-01
[WIP] Improve build: Makefile and more tests in Travis CI

genodeftest:
Missing items:
- [ ] (optional) Run documentation and style checks only on one machine, preferably a separate one
- [ ] build all the code to check whether it compiles
- [ ] add sphinx to docker images (see exaile/exaile-testimg#2)
- [ ] build sphinx documentation during tests
- [ ] add pychecker to docker images (see exaile/exaile-testimg#2)
- [ ] check code style using pychecker
- [ ] put built .pyc / .pyo files where we can write them.

Code style:
```
$ pycodestyle --exclude=plugins/daapserver/spydaap/*.py --max-line-length=160 --statistics --count --ignore=E203,E241,E265,E266,E305,E402,E722,E731,E741,W293,W503 .
```

Deferred:
- [ ] add pytest-flakes to docker images (through `pip install`, see exaile/exaile-testimg#2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
 
 script:
 - docker run --rm -it -e HOME=/home -v $(pwd):/app -w /app exaile/exaile-testimg:${IMGTAG} make test
+- docker run --rm -it -e HOME=/home -v $(pwd):/app -w /app exaile/exaile-testimg:${IMGTAG} make check-doc
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,7 @@ before_install:
 - docker pull exaile/exaile-testimg:${IMGTAG}
 
 script:
-- docker run --rm -it -e HOME=/home -v $(pwd):/app -w /app exaile/exaile-testimg:${IMGTAG} make test
-- docker run --rm -it -e HOME=/home -v $(pwd):/app -w /app exaile/exaile-testimg:${IMGTAG} make check-doc
+- docker run --rm -it -e HOME=/home -v $(pwd):/app -w /app exaile/exaile-testimg:${IMGTAG} make BUILDDIR=/tmp test check-doc
 
 notifications:
   irc:
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,10 @@ EXAILESHAREDIR = $(DESTDIR)$(DATADIR)/exaile
 EXAILECONFDIR  = $(DESTDIR)$(XDGCONFDIR)/exaile
 EXAILEMANDIR   = $(DESTDIR)$(MANPREFIX)/man
 
-.PHONY: dist test completion coverage clean sanitycheck builddir
+.PHONY: all all_no_locale builddir compile make-install-dirs uninstall \
+	install install_no_locale install-target locale install-locale \
+	plugins-dist manpage completion clean pot potball dist test test_coverage \
+	lint_errors sanitycheck
 
 all: compile completion locale manpage
 	@echo "Ready to install..."
--- a/Makefile
+++ b/Makefile
@@ -226,9 +226,10 @@ dist:
 	./tools/dist.sh
 	rm -rf dist/copy
 
+# See .travis.yml for details on how tests are ran by Travis CI
 check-doc: clean
 	$(MAKE) -C doc html
-	$(MAKE) -C doc linkcheck
+
 
 test:
 	EXAILE_DIR=$(shell pwd) PYTHONPATH=$(shell pwd) $(PYTEST) tests
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,8 @@ EXAILEMANDIR   = $(DESTDIR)$(MANPREFIX)/man
 
 .PHONY: all all_no_locale builddir compile make-install-dirs uninstall \
 	install install_no_locale install-target locale install-locale \
-	plugins-dist manpage completion clean pot potball dist test test_coverage \
-	lint_errors sanitycheck
+	plugins-dist manpage completion clean pot potball dist check-doc test \
+	test_coverage lint_errors sanitycheck
 
 all: compile completion locale manpage
 	@echo "Ready to install..."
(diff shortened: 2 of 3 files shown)


genodeftest:
Missing items:
- [ ] (optional) Run documentation and style checks only on one machine, preferably a separate one
- [ ] add pytest-flakes to docker images
- [ ] put built .pyc / .pyo files where we can write them.

genodeftest:
Any chance we can get updated docker images?

sjohannes:
Which ones are you thinking of updating? It seems the tests are failing across all tested OS versions, most aren't too old.

genodeftest:
@sjohannes According to the warning, python2 must be at least version 2.7.9, but as that has been released 2014-12-10, I think the warning is wrong, so updating probably won't help.

Do you know how to get write permissions?

sjohannes:
The error looks like it's from `pip install pytest-flakes` (because not root?). And I think that's the wrong place to install stuff; my understanding is that you need to put it in [our dockerfiles](https://github.com/exaile/exaile-testimg).

genodeftest:
@sjohannes : Thanks, that's what I was searching for.

genodeftest:
It looks like both Debian and Ubuntu don't have python-pytest-flakes in their repos. Should I use pip to install them in the Dockerfiles?

sjohannes:
Yeah, it doesn't affect users so it doesn't matter how things are installed.

virtuald:
I think if you push to the repos that the docker images are built from it'll get built automatically via docker hub.

.travis.yml
```diff
@@ -12,13 +12,20 @@ env:
 before_install:
 - docker pull exaile/exaile-testimg:${IMGTAG}
 
+install:
+- pip install --user pytest-flakes
```
virtuald: This is happening on the travis VM, not in the docker images. That's most likely why it's failing.

genodeftest:
Debian 8 build fails and I don't understand why. Maybe the "-m" switch has been changed?
```
make -C doc html
make[1]: Entering directory '/app/doc'
/usr/bin/python2: No module named sphinx.__main__; 'sphinx' is a package and cannot be directly executed
```

Debian 9 and Ubuntu 16.04 builds fail differently, which looks like https://bugzilla.redhat.com/show_bug.cgi?id=1498084
```
make -C doc html
make[1]: Entering directory '/app/doc'
  File "/usr/lib/python2.7/dist-packages/sphinx/__main__.py", line 14, in <module>
    sys.exit(main(sys.argv))
  File "/usr/lib/python2.7/dist-packages/sphinx/__init__.py", line 51, in main
    sys.exit(build_main(argv))
  File "/usr/lib/python2.7/dist-packages/sphinx/__init__.py", line 61, in build_main
    from sphinx import cmdline
  File "/usr/lib/python2.7/dist-packages/sphinx/cmdline.py", line 14, in <module>
    import optparse
  File "/usr/lib/python2.7/optparse.py", line 419, in <module>
    _builtin_cvt = { "int" : (_parse_int, _("integer")),
  File "/usr/lib/python2.7/gettext.py", line 642, in gettext
    return dgettext(_current_domain, message)
  File "/usr/lib/python2.7/gettext.py", line 606, in dgettext
    codeset=_localecodesets.get(domain))
  File "/usr/lib/python2.7/gettext.py", line 541, in translation
    mofiles = find(domain, localedir, languages, all=1)
  File "/usr/lib/python2.7/gettext.py", line 513, in find
    for nelang in _expand_lang(lang):
  File "/usr/lib/python2.7/gettext.py", line 206, in _expand_lang
    from locale import normalize
ImportError: cannot import name normalize
```

Fedora 26 build misses `find`:
```
make -C plugins clean
make[1]: Entering directory '/app/plugins'
find . -name "*.py[co]" -exec rm -f {} \;
/bin/sh: find: command not found
```

Ubuntu 17.04 build seems to be the one which got most far but fails due to permissions:
```
make -C doc html
make[1]: Entering directory '/app/doc'
Running Sphinx v1.5.3
making output directory...
Exception occurred:
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/app/doc/_build'
The full traceback has been saved in /tmp/sphinx-err-q4oJfT.log, if you want to report the issue to the developers.
```

genodeftest:
All the distros ship pretty outdated versions of sphinx. Using pip to install sphinx (instead of the native package manager) we'll get a recent version, which could fix most issues here except for the permissions.

genodeftest:
With exaile/exaile-testimg#5 fixed, let's try again.

* Fedora 26 build is still broken because of https://bugzilla.redhat.com/show_bug.cgi?id=1498084 , but a fix is on the way and will land in a few days.
* All the other builds fail because `/app/` is not user-writable in the docker VM and I don't know how to fix that. Is there any writable directory on this VM? Should I copy `/app` recursively to the user home or `/tmp/`?

virtuald:
Oh. The reason /app isn't writeable is because the default entrypoint we use in gst-python image creates a custom user and tries to run as an unprivileged user (which is necessary for using X/pulse, but not necessary to run tests). I'll update the entrypoints on all the images, and if we retrigger this it will get further.

virtuald:
Ok, so I fixed that bug, but now the builds are failing for a different reason... it's not 100% clear to me why, since they pass for me locally. It's probably related to the last bug.

virtuald:
Haha, the 'no permission' tests are failing now because the tests are running as root...

virtuald:
Ok, so I've reverted the test images to run as a normal user, and also installed sphinx-rtd-theme on them (as I seem to remember there was some failure related to that).

I think what we need to add is an environment variable to override the sphinx build output directory, that way it writes inside the container to an allowed space. I'm going to test it locally once docker hub finishes building a new image.

virtuald:
I tried it locally, and it works... so I pushed my changes to this PR (feel free to rebase them away). Of course, this build will fail because docker hub is unfortunately *really* slow at building images. However, once it's done, I expect the build will pass again.

virtuald:
Looks like building the docs passed except for the fedora image, because of that bug you pointed out above.

The link-check failed... I'm not totally convinced that we need that? I don't want builds failing because some random site isn't up at the moment.

genodeftest:
Ok, I've removed the link check for now.

PR #453, opened 2017-11-09
Travis: Also build all of Exaile's code

genodeftest:
This will allow us to find more coding bugs prior to committing code to the repository.


genodeftest:
I'd like to merge this in case you agree with the test.

Issue #437, opened 2017-10-02
During scanning disk for files, exaile's GUI is very slow


genodeftest:
### Steps to Reproduce
1. Open Exaile
2. Open collection manager
3. Add a new large directory full of many music files
4. Press "Apply"
5. Try to interact with Exaile

B. alternative steps to reproduce:
1. Open Exaile
2. use files browser to navigate to a large directory containing many music files
3. drag and drop folder(s) into playlist

### Expected Behavior
Exaile GUI should not be slow

### Current Behavior
Exaile GUI is very slow

### Possible Solution
We need to have a look through async stuff

### Environment
* Operating System and version: Fedora 26, Linux kernel 4.12.x, 4.13.4
* Exaile Version: 4.0.0beta2+2bc4bbe or many previous versions

sbrubes:
Since [commit: "playlist: massive speedup"](https://github.com/exaile/exaile/commit/d4495182c38a6b1efafa296b14c0fdecc5d3a59d) by @virtuald, the "_B. alternative steps to reproduce_" isn't fixed (for [more than 500 tracks](https://github.com/exaile/exaile/blob/e677d2fc6a58417c8b254387929ac853b5e18981/xlgui/widgets/playlist.py#L1765))?

sbrubes:
related to #94

sbrubes:
It’s probably good to note here that the “restore of opened tabs” with too much items, will delay the start of exaile window.

genodeftest:
> It’s probably good to note here that the “restore of opened tabs” with too much items, will delay the start of exaile window.

Why is that? Do we access all track files when restoring opened tabs? If yes, that should not happen.

PR #457, opened 2017-11-11
Batch track_tags_changed events

### LINKED ISSUE #334: Change track_tags_changed to be per-track, instead of per-tag
One complaint I've often heard is that scanning collections takes a really long time. There are a lot of reasons for this, but I think one change that could make a huge difference is changing `track_tags_changed` to be only emitted once per change to a track, instead of 8 different events per track. Lots of stuff happens as a result of events, so fewer of them occurring would be a positive thing.

There's only maybe a dozen or so places where the track tags event is used, and most of them don't actually care that the specific tag changed, only that the track itself changed.

I think if this change were to be made, then it should be done for Exaile 4. If there are no objections, I think I should be able to do it fairly quickly.

virtuald:
* Fixes #334
* Helps with #437

### DIFF
--- a/doc/dev/plugin_guide.rst
+++ b/doc/dev/plugin_guide.rst
@@ -207,9 +207,9 @@ You can set the track information like this:
 
 .. code-block:: python
 
-    myTrack.set_tag_raw('title', "Cool Track")
-    myTrack.set_tag_raw('artist', "Cool Artist")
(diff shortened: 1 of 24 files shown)


genodeftest:
I've fixed a small mistake, now it builds again.

Testing works fine for me. Without objective testing I think scanning the library got faster. Changes look fine to me.

PR #458, opened 2017-11-11
Don't emit track_tags_changed when adding new files to the collection

### LINKED ISSUE #357: During adding files to playlist, `track-tags-changed` is sent very often
Steps to reproduce:
1. watch the handler for the `track-tags-changed` signal in info.TracklistInfoPane
2. start exaile
3. add many files to a playlist

What happens:
The `track-tags-changed` event is emitted on every single track.

What should happen:
I don't know whether that is necessary, but I doubt it. This may be one reason why it takes so long to add just a few files to a playlist.

Version:
Exaile from git master

Additional info:
This issue probably also affects other stuff. I stumbled upon it while fixing the osd plugin for #346.

virtuald:
Note: This contains #457 , so only look at the last commit for this PR when reviewing it.

I'm less certain about this change... it does what is asked for by #357, but now that I'm thinking about it... should all notifications be suppressed when reading in a new file? Which means any reading done in `__new__` or `__init__` would not generate track_tags_changed notifications.

I'm fairly certain that suppressing them in `__init__` when url=something is desired behavior -- the only time `__init__` is going to be used with a URI object is to create a new track (otherwise `__new__` would return the old track). And if it's a brand new track, probably nobody cares about such notifications, right?

However, I think in `__new__`, the set_tags there should emit notifications, because the track will have existed previously.

* Fixes #357

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -235,9 +235,10 @@ class Track(object):
             self._unpickles(_unpickles)
             self.__register()
         elif uri:
-            self.set_loc(uri)
+            # notify isn't needed here because this is a new track
+            self.set_loc(uri, notify_changed=False)
             if scan:
-                self.read_tags()
+                self.read_tags(notify_changed=False)
         else:
             raise ValueError("Cannot create a Track from nothing")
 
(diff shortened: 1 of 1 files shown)


genodeftest:
I took the liberty to rebase your commit to latest git master so I could check exaile/exaile-testimg#1. Works fine now.

PR #465, opened 2017-11-19
Massive speedup

virtuald:
Since all the data we need is stored in the track object in the PlaylistModel, there's no reason to extract the information and store it in separate columns.    It turns out this is quite expensive. Instead, change the column renderers to    pull the text directly from the track, which has the added advantage of only
    extracting data as needed, instead of rendering the entire playlist.

Contains other misc playlist fixes too.

### DIFF
--- a/plugins/minimode/controls.py
+++ b/plugins/minimode/controls.py
@@ -798,7 +798,7 @@ class PlaylistButtonControl(Gtk.ToggleButton, BaseControl, QueueAdapter):
         """
             Updates the internally stored playlist
         """
-        columns = self.view.get_model().columns
+        columns = self.view.get_model().column_names
         model = PlaylistModel(playlist, columns, player.PLAYER, self.view)
         self.view.set_model(model)
 
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -927,6 +927,7 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
             columns = self.get_columns()
             for col in columns:
                 self.remove_column(col)
+                col.destroyed = True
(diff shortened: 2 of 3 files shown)


virtuald:
I suspect if I cache certain information instead of computing it over and over again, the scrolling will get better.

Of course, that doesn't seem much better than creating a giant liststore... but I think I expect using a single python object is probably more efficient than managing the various columns? I'll try it out and see what happens.

virtuald:
I switched the play icon stuff to use the playback events instead of listening to row events. Row events happen a lot more often anyways, so this should be a net win, and fix the bug you mentioned.

virtuald:
There's also a problem with the columns not being saved, but I think the bug was introduced in a prior commit.

genodeftest:
This change makes the GUI a bit more responsive during scanning the library I think.

Utilizing `GObject.handler_block()` looks like a very good improvement to me which should probably be added on different other places over the code. How do we make sure it does not get lost here?

virtuald:
There are changes I want to make, but I don't remember what they are. On Thanksgiving travel, so I'll revisit this next week.

virtuald:
@sjohannes try this out, the scrolling should be better now

sjohannes:
Yeah, feels better now.

virtuald:
FYI, using the random collection generator, I generated a 50000 song library, and Exaile can handle it now. However, it has issues at 200000 still... but I have ideas on how to address that.

sjohannes:
@virtuald I think this still has the [SPAT issue](https://github.com/exaile/exaile/pull/465#discussion_r151860702) (when you set SPAT on a non-playing tracklist).

PR #471, opened 2018-01-01
Add Catalan translation in data/exaile.desktop

rbuj:


genodeftest:
Looks fine. Should go into localization via [weblate](https://hosted.weblate.org/projects/exaile/master/), though.

genodeftest:
To be more specific: We should modify Exaile's builds to localize the ".desktop" file. This is out of scope for such a simple pull request, though.

sjohannes:
Thanks for the patch!

sjohannes:
Wait, I didn't actually merge it in....

sjohannes:
> To be more specific: We should modify Exaile's builds to localize the ".desktop" file.

[For reference.](https://blog.du-a.org/2014/03/17/gettext-desktop/) Not sure if we're supposed to use the same .pot/.po files; might need a separate set so the strings don't end up in the .mo files.

Issue #369, opened 2017-05-13
Drag/drop crash on OSX


virtuald:
### Steps to Reproduce (for bugs)

1. Drag anything from the playlist

### Current Behavior

Segfault. :(
### Environment

* Operating System and version: OSX 10.12

* Exaile Version: git master

genodeftest:
Any backtrace?

virtuald:
Not yet, planning on investigating it tonight.

virtuald:
This is a GTK issue, as I can reproduce it on gtk3-demo-application by just dragging some text around. Will file a bug upstream.

virtuald:
Upstream @ https://bugzilla.gnome.org/show_bug.cgi?id=782914

sbrubes:
http://python-gtk-3-tutorial.readthedocs.io/en/latest/drag_and_drop.html

I have tested this and works.

virtuald:
Fixed in #472

PR #472, opened 2018-01-11
Fix for issue #369 'Drag/drop crash on OSX'.

sbrubes:
Passing taget_list=None to Gtk.Widget.drag_dest_find_target, will make it use internally Gtk.Widget.drag_dest_get_target_list.

The bug is related to the lifetime of Gtk.TargetList returned by Gtk.Widget.drag_dest_get_target_list (When it gets destroyed, it crashes).

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -1333,7 +1333,7 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
 
         action = Gdk.DragAction.MOVE
         _, _, _, modifier = self.get_window().get_pointer()
-        target = self.drag_dest_find_target(context, self.drag_dest_get_target_list()).name()
+        target = self.drag_dest_find_target(context, None).name()
 
         if target == 'text/uri-list' or \
            (self._hack_is_osx and self._hack_osx_control_mask) or \


sjohannes:
Thanks for the patch! I don't have an OS X machine to test this so will have to wait for Dustin.

> The bug is related to the lifetime of Gtk.TargetList returned by Gtk.Widget.drag_dest_get_target_list (When it gets destroyed, it crashes).

Interesting. Would it be better to somehow make sure it's not destroyed (e.g. by copying it)?

sbrubes:
Well, for this usage, we are fine.

But it's a good idea looking for uses of _drag_dest_get_target_list_ and _drag_source_get_target_list_.

Also reporting the bug to Gtk, since it is not an expected behavior.

virtuald:
That would be fantastic if this fixed the problem. I just returned from vacation, so I have access to my OSX machine again... I'll try to test this soon.

virtuald:
I've got it on my agenda to test this PR this weekend. Need to rebuild my OSX gst environment first.

sbrubes:
It could be merged already, since using [None as parameter](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Widget.html#Gtk.Widget.drag_dest_find_target), it’s clearly an improvement, even keeping the bug open.

Thank you all.

virtuald:
You're right that it could be merged, but by keeping it open it reminds me to check to see if the bug is fixed or not. :) 

I spent a really long time yesterday getting my OSX environment reinitialized using the python-gtk-gst-sdk , and it's finally working. I'll try to test this (and other things) tonight.

sbrubes:
No problem :D

virtuald:
I don't know how you figured this one out, but it definitely works. There's another place where the same error exists, and I was able to demonstrate this fix working there also.

Thanks a ton!

sbrubes:
Debugging could be even faster.

I'm now looking for the problem (probably OS X only) when you drop multiple items on `Playlist` from `Collection` or `Files`panel.

It seems not to be working, getting only the first added.

I'm gonna update to last _Gtk_ version (*26 I believe), and test it.

If still not working, I'll create a fix.

P.S. If you could confirm it at yours, we could even create an issue.

Tks.

PR #474, opened 2018-01-27
DragTreeView Improvements

sbrubes:
Patch with some improvements for DragTreeView

### Highlights

- Fix of many mouse button selection issues (https://github.com/exaile/exaile/commit/6088681ed43a44beec65752733c2269e2d6a17ec)

- Menus are more tight to DragTreeView(https://github.com/exaile/exaile/pull/474/commits/eea7c8afaacb40891487b11315933fd2aae2f308)

- Use drag cover icon (https://github.com/exaile/exaile/commit/44d8219b64ba05084147ffd689e9f635e530018c)

### DIFF
--- a/xl/common.py
+++ b/xl/common.py
@@ -1051,4 +1051,54 @@ def subscribe_for_settings(section, options, self):
     return event.add_callback(_on_option_set, '%s_option_set' % section.replace('/', '_'))
 
 
+class AsyncLoader:
+    """
+        Async loader based on a generator
+        Threaded, load it and put it in `result_list`
+    """
+    def __init__(self, item_generator):
+        """
+            Constructs and already start processing (starts thread)
+            :param item_generator: iterable
+        """
+        self.__end = False
+        self.__result_list = []
+        self.__thread = threading.Thread(target=self.run, args=(item_generator,))
+        self.__thread.start()
+
+    def run(self, item_generator):
+        """
+            Process items putting in `result_list`
+            :param item_generator: iterable
+            :return: None
+        """
+        for i in item_generator:
+            if self.__end: break;
+            self.__result_list.append(i)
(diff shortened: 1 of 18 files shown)


xl/common.py
```diff
@@ -1051,4 +1051,54 @@ def _on_option_set(unused_name, unused_object, data):
...
+        """
+        return not self.__thread.is_alive()
+
+    @property
+    def result(self):
+        """
+            Gets the result
+            :return: list
+        """
+        return self.__result_list[:]
```
virtuald: Why copy the list? I have to imagine this class is fairly single purpose.

virtuald:
Though, don't we already have a type of async loading somewhere already? Progress thread? Sorta. Maybe it's worth integrating them together.

xl/common.py
sbrubes: Two things made me retrieving the actual result to the list.

First, you could use it outside getting results while the thread still running (adding items).

The second, it's even when you request the end, you don't really know when it ends.

Copying I'm getting a state of the list, so it's more what we want, for any uses.

That avoid list changes while I'm iterating it, for example.

Don't worry, this doesn't really affects performance.

Otherwise you must do something like:
```
end(0.333)
while not has_ended(): wait
then read result
```

Maybe it's also a good idea putting it on methods doc (like, "returns current state, copying it").

sbrubes:
Maybe you was referring to `copy_async` (https://github.com/exaile/exaile/issues/198).
I remember seeing something related years ago, I believe was it (but maybe wasn't even on Exaile).

Anyway, I've choose the model of this imps because it's a class, creating an object that could be stored while processing/and request it to end.

I saw that the old imp was using`@common.threaded`, but I couldn't use it, because I need to interact with the created thread (joining it).

PR #477, opened 2018-01-27
Handling keyboard interruption to kill graciously

sbrubes:
Moves xlgui.mainloop function to inside passing self.quit to be called on exit


sjohannes:
Good point. Thanks.

Issue #479, opened 2018-02-04
Drag'n'Drop from Panel to Playlist, only loads first track (OSX)


sbrubes:
Dragging and dropping from Panel to Playlist on OS X, only loads first track because the [setted uris](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/SelectionData.html#Gtk.SelectionData.set_uris) are not received. 

Between [`source.drag_data_get`](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/Widget.html#Gtk.Widget.signals.drag_data_get) and [`drop_dest.drag_data_received`](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/Widget.html#Gtk.Widget.signals.drag_data_received) it gets truncated keeping only the first uri. 

https://github.com/exaile/exaile/blob/197a93ada858e05caa824ff78f86b33634267617/xlgui/panel/files.py#L448 

https://github.com/exaile/exaile/blob/197a93ada858e05caa824ff78f86b33634267617/xlgui/widgets/playlist.py#L1288

I have this simply demonstrated at [attached code](https://github.com/exaile/exaile/files/1693089/gtk_dnd_uris.zip).

It's a version of the code from http://python-gtk-3-tutorial.readthedocs.io/en/latest/drag_and_drop.html using uris instead. (just execute, drag'n'drop an icon and it will print result/the last printted info must be 'True', the received uri list it's the same setted)

I have not tested extensively others target types, but it should be double checked.

This bug it's a so unacceptable behavior for an internal widget, that I suggest entirely taking this off the [Gtk.SelectionData](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/SelectionData.html#Gtk.SelectionData) creating a tracks cache to switch between internals dnd.

Probably code has already an effort on this direction with the class attribute [xlgui.widgets.common.DragTreeView.dragged_data](https://github.com/exaile/exaile/blob/197a93ada858e05caa824ff78f86b33634267617/xlgui/widgets/common.py#L220). (I'm not sure but it seems to be a memory leak today, since it seems not to be consumed)

As this implementation will take much time, I've created a workaround using `xlgui.widgets.Playlist._hack_is_osx` and `get_selected_tracks` ("xlgui.widgets.common.DragTreeView").

### Steps to Reproduce (for bugs)
1. Pick some tracks (more than one) from Files/Collection Panel and drop on Playlist

### Expected Behavior
Drop items

### Current Behavior
Only first

[...]

PR #480, opened 2018-02-04
Workaround to #479

sbrubes:
Workaround to internal Drag'n'Drop on OSX, use get_selected_tracks from source widget, if available

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -1249,6 +1249,21 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
         else:
             insert_position = -1
 
+        if self._hack_is_osx:
+            source_widget = Gtk.drag_get_source_widget(context)
+            if (not (source_widget is None or source_widget is self) and
+                    hasattr(source_widget, 'get_selected_tracks')):
+                try:
+                    tracks = source_widget.get_selected_tracks()
+                    if insert_position >= 0:
+                        self.playlist[insert_position:insert_position] = tracks
+                    else:
+                        self.playlist.extend(tracks)
+
+                    return
+                except:
+                    pass  # Any exception on hack is ignored
+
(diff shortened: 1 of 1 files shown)


xlgui/widgets/playlist.py
```diff
@@ -1249,6 +1249,21 @@ def on_drag_data_received(self, widget, context, x, y, selection, info, etime):
...
+                    hasattr(source_widget, 'get_selected_tracks')):
+                try:
+                    tracks = source_widget.get_selected_tracks()
+                    if insert_position >= 0:
+                        self.playlist[insert_position:insert_position] = tracks
+                    else:
+                        self.playlist.extend(tracks)
+
+                    return
+                except:
```
virtuald: Maybe do a `logger.debug("error processing DnD hack", exc_info=True)`. Also, `except Exception:` instead.

xlgui/widgets/playlist.py
```diff
@@ -1249,6 +1249,21 @@ def on_drag_data_received(self, widget, context, x, y, selection, info, etime):
         else:
             insert_position = -1
 
+        if self._hack_is_osx:
```
virtuald: Let's add a link to the original bug so someone in the future knows why this is here.

sbrubes: I thought that git blame do that.

xlgui/widgets/playlist.py
sbrubes: It's only a precaution, like a get_selected_tracks for an object with a different method signature.

We don't actually have any error, and if has, it will be treated as the actual state.

I don't see this as impediment.

sbrubes:
Couldn't you change it after or before it?

sbrubes:
Anyway it’s just a workaround, #479 should be kept open until Gnome says it’s fixed (I think 🤔).

PR #481, opened 2018-02-09
Update de.po

jschwender:
some corrections to number and units presentation, some completions.


virtuald:
Thanks for the contribution! In the future, translations should be done via our hosted weblate site at https://hosted.weblate.org/projects/exaile/master/

Issue #197, opened 2015-12-14
Reorder issue with Exaile 4


dangmai:
In Exaile 4 on Windows 8, I can't reorder songs in my playlist anymore by dragging and dropping, here's a GIF of the behavior:
http://i.imgur.com/XJHEhEf.gifv
I can see where the song should be ending up, but it just doesn't go there

genodeftest:
This issue is present on Linux (Fedora, Gtk+ 3.18.x) too.

virtuald:
I knew there was a drag and drop bug somewhere that I had forgotten about.

tfree87:
I found a few changes allow for some reordering, however, after about 4 'drag and drop' operations playlist reordering no longer works. I found that the playlist was ignoring drag and drop operations because of an error in string comparison and because of some name changes in methods associated with Gtk. This commit shows the changes that allow for this _partial restore_ of drag and drop function: https://github.com/tfree87/exaile/commit/58d20db0ba00628e541473b064c7a55ad4ce7c71 . I have yet to figure out why this only works for a limited number of operations.

tfree87:
I seem to have found why the 'drag-and-drop' operation doesn't always work. It appears that get_selected_rows() in xlgui/widgets/playlist.py sometimes returns and empty list for `paths` when `selection.get_selected_rows()` is called. This only happens when I do multiple 'drag-and-drop' operations on a single track. When I select multiple tracks, reordering is restored.

`def get_selected_paths(self):
        """
            Returns a list of pairs of treepaths which are currently
            selected in the playlist.
            
            The treepaths are returned for the base model, so they are
            indices that can be used with the playlist currently
            associated with this view.
        """
        selection = self.get_selection()
        model, paths = selection.get_selected_rows()
        
        if isinstance(model, Gtk.TreeModelFilter):
            paths = [model.convert_path_to_child_path(path) for path in paths]
        
        return paths`

Not quite sure how to address this issue.

javiercantero:
The commit d3819627973a893ced2886727c5910fe01a108e4 supersedes @tfree87's patch. I've tested it and it works, now you can reorder songs drag-and-dropping them.

However, there's another problem: after dropping the song, you need to click twice to select the next song to drag. If you don't do the second click, a d&d operation doesn't work.

Another problem (that probably should be considered a separate issue) is if the dragged song is the song being played: when dropped it loses any graphical sign of being played including the play icon.

 so this issue could be closed

sbrubes:
[...]

PR #483, opened 2018-03-02
Fix drag & drop issues

virtuald:
The playlist drag and drop issues have gotten to me recently, and instead of inventing a completely new set of code to fix it, I've decided to steal quodlibet's high quality code instead (which incidentally, it appears our code was derived from at some point). It's also GPLv2, and I've retained the original license notices, so we should be good.

The good news is that the drag/drop issues in #197 seems to have mostly gone away with this. The bad news is it's still slightly wonky on my laptop (but works great on my desktop)... there were instances where I lost focus and couldn't seem to get it back except via a double-click (though maybe this is due to unhandled touch events?).

I'm going to look at migrating some of our other treeviews over to use this to simplify more stuff, but I'll keep this PR open for a few days in case there are any objections to this theft (cc @lazka?).

### DIFF
--- a/xl/externals/qltk_treeview.py
+++ b/xl/externals/qltk_treeview.py
@@ -530,7 +530,9 @@ class MultiDragTreeView(BaseView):
         try:
             path, col, cellx, celly = self.get_path_at_pos(x, y)
         except TypeError:
-            return True
+            if not event.get_state() & get_primary_accel_mod():
+                self.__pending_action = (None, None, False)
+            return
         selection = self.get_selection()
         is_selected = selection.path_is_selected(path)
         mod_active = event.get_state() & (
@@ -550,6 +552,8 @@ class MultiDragTreeView(BaseView):
             selection.set_select_function(lambda *args: True, None)
             if single_unselect:
                 selection.unselect_path(path)
+            elif path is None:
+                selection.unselect_all()
             else:
                 self.set_cursor(path, col, 0)
             self.__pending_action = None
@@ -571,7 +575,7 @@ class RCMTreeView(BaseView):
         try:
             path, col, cellx, celly = self.get_path_at_pos(x, y)
         except TypeError:
-            return True
+            return
         self.grab_focus()
         selection = self.get_selection()
         if not selection.path_is_selected(path):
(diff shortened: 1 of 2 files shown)


lazka:
Sure, no need to ask. All of QL and mutagen is gplv2+ now (also true for 99% of the code history)

dangmai:
I tried the patch and DnD is mostly great now! Appreciate the effort @virtuald. However, as you already pointed out, there are instances where it loses focus after a DnD operation, I've narrowed it down to this case (there might be more edge cases that I'm missing): If I drag and drop the same item twice then it'll lose focus, if I choose a different item after the first DnD then it's fine.

sbrubes:
# Button selection test scope
Maps the old behavior.
[V] working [X] not working [P] partiallly working
## On Empty Area
With Ctrl or Shift doesn't change selection [V]
Unselect all for any primary or secondary button press without Ctrl/Shift pressed [X]
Shows playlist menu on secondary button click [X]
## Primary Button
With Ctrl toogle that row (only) selection [P] [if you select many you can't unselect] 
With Shift selects items in that range (from the last row to this new) [P] [when you select and re-select with Shift, like choosing one line above, it acts like Ctrl and unselect that row (if you continue selecting you'll see many distinct treatments for this event]
For clicks without Ctrl or Shift pressed, sets cursor (only) at that row [V]
## Secondary Button
*Indistinct treatment with or without Ctrl/Shift pressed
On a selected row, shows menu, keeping (any) selected row [V] 
On a not selected row, shows menu, setting the cursor on that row (only) [V]

virtuald:
@sbrubes thanks for the list of tests. I fixed the empty area clicking... but the SELECT bug is a bit more annoying and I don't think it'll be easily fixable, so I'm going to leave it for now.

The selection behavior after a drag is still really weird and I'm not quite sure what's going on there. Sometimes it selects, sometimes it doesn't. I wasn't able to duplicate @dangmai 's issue... but perhaps it's a Windows/Linux difference.

dangmai:
Hmm that's the behavior on Linux (I'm running Arch). I just ran it again using your latest commit on the `steal-dnd` branch and it's still there.

virtuald:
Closing in favor of #484 ... but I think this branch will eventually come back.

PR #486, opened 2018-03-10
smart playlists: add 'contains word' operation

virtuald:
- Useful for users that aren't familiar with regexes

Any objections to this silly addition? I'm willing to accept suggestions for a different search string instead of `!c=` and `c=`...  :)

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -37,6 +37,7 @@ from datetime import datetime, timedelta
 import logging
 import os
 import random
+import re
 import time
 import urlparse
 import urllib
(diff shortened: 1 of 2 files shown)


xl/playlist.py
```diff
@@ -1947,6 +1948,18 @@ def _create_search_data(self, collection):
                     {
                         'field': field
                     }
+            elif op == 'c=': # contains word
+                s += '%(field)s~"\\b%(value)s\\b"' % \
+                    {
+                        'field': field,
+                        'value': re.escape(value),
+                    }
+            elif op == '!c=': # does not contains word
```
sjohannes: does not contain

Issue #485, opened 2018-03-09
Deleting lots of tracks is slow if the cursor is at the top


sjohannes:
1. Open the Random 500 playlist.
2. Set focus on the tracklist.
3. Press Home, Shift+End, Del.  (CASE 1)
4. Do steps 1-2.
5. Press End, Shift+Home, Del.  (CASE 2)

Result: In case 1, there is a noticeable delay but it's under 1 s. In case 2, the delay is 2-3 seconds.

virtuald:
It's going to be because the cursor keeps changing (I think as a result of the rows below being inserted). It's yet another weird synchronization error from our model/playlist combination.

virtuald:
Found a fix, I'll push it in to #488 as the solutions are related

PR #488, opened 2018-03-11
playlist: restore view when filtering

virtuald:
Previously, when filtering it would just scroll to the top... which is fairly annoying if you're dealing with a large playlist.

This PR restores the normal behavior if you were just using a normal treeview + model without removing the model.

Also includes a fix to #485, as that fix can use the without_model helper function that was introduced here.

### DIFF
--- a/xlgui/guiutil.py
+++ b/xlgui/guiutil.py
@@ -626,8 +626,8 @@ def _get_closest_visible(model, child_path):
 @contextlib.contextmanager
 def without_model(tv):
     '''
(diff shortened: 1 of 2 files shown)


sjohannes:
Another regression: If "Resume playback on start" and "Jump to current song on track change" are true, the resumed track should be selected and visible.

sjohannes:
Huh, never mind about the second one. It wasn't working before, now it does....

virtuald:
@sjohannes give this a shot

PR #492, opened 2018-03-16
Dnd fixes

sbrubes:
That's it.

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -1378,8 +1378,9 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
                 insert_position = len(self.playlist) - len(tracks)
 
         # Select inserted items
-        if insert_position >= 0 and \
-            insert_position < len(self.playlist) and len(tracks) > 0:
+        if (0 <= insert_position < len(self.playlist) and
+            0 < len(tracks) <= 500):
+            # More than 500 songs are loaded threaded, so ignore it
             self.get_selection().select_range(
                 self.model.get_path(
                     self.model.iter_nth_child(None, insert_position)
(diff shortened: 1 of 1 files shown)


sbrubes:
@virtuald I did not test the `text/uri-list` since I'm on mac, and will be handle by #480

xlgui/widgets/playlist.py
```diff
@@ -1295,17 +1299,38 @@ def on_drag_data_received(self, widget, context, x, y, selection, info, etime):
...
                         self.playlist.extend(tracks)
-
-                    return
+                        insert_position = len(self.playlist) - len(tracks)
                 except Exception:
                     pass  # Any exception on hack is ignored
+                else:
+                    # Select inserted items
+                    if tracks:
+                        self.get_selection().select_range(
```
virtuald: Seems like you should unselect prior selected tracks?

sbrubes:
@virtuald I could not reproduce that.

Are you sure shuffle isn't on?

xlgui/widgets/playlist.py
sbrubes: You are probably right, but it seems to be working (maybe it unselects or moves cursor when model changes?).

virtuald: No, the old tracks are selected along with the newly dragged tracks, which isn't very intuitive.

virtuald:
Nevermind, I had things in the queue... it was playing from the queue instead. Weird.

xlgui/widgets/playlist.py
virtuald: I think if you unselect tracks before you select new tracks, this PR is ready to go.

sbrubes: > No, the old tracks are selected along with the newly dragged tracks, which isn't very intuitive.

Are you saying that you actually have this behavior?

sbrubes:
Sorry, I’ve forgot to test/check Queue.

PR #497, opened 2018-03-23
Fix plugins' teardown() method signature

rokm:
It seems not all plugins have been updated to have their teardown() method include the second argument. 

This causes the following errors being displayed on terminal when shutting down the program:
```
...
INFO    : Exaile is shutting down...
INFO    : Tearing down plugins...
ERROR   : Unable to tear down plugin notify
  File "/home/rok/Projects/exaile/xl/plugins.py", line 296, in teardown
    plugin.teardown(main)
TypeError: teardown() takes exactly 1 argument (2 given)
ERROR   : Unable to tear down plugin keybinder
  File "/home/rok/Projects/exaile/xl/plugins.py", line 296, in teardown
    plugin.teardown(main)
TypeError: teardown() takes exactly 1 argument (2 given)
INFO    : Saving state...
INFO    : Bye!
```

This PR fixes the signature of teardown() method of the offending plugins.

### DIFF
--- a/plugins/notify/__init__.py
+++ b/plugins/notify/__init__.py
@@ -333,14 +333,14 @@ class NotifyPlugin(object):
         return GLib.SOURCE_REMOVE
 
     def disable(self, _exaile):
+        self.teardown(exaile)
+
+        self.__exaile = None
+
+    def teardown(self, exaile):
         if self.__notifier:
             self.__notifier.destroy()
             self.__notifier = None
-        self.__exaile = None
-
-    def teardown(self):
-        if self.__notifier:
-            self.__notifier.destroy()
 
     def get_preferences_pane(self):
         return notifyprefs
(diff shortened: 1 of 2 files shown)


virtuald:
Thanks!

plugins/notify/__init__.py
```diff
@@ -333,14 +333,14 @@ def __init_notifier(self, caps):
         return GLib.SOURCE_REMOVE
 
     def disable(self, _exaile):
-        if self.__notifier:
-            self.__notifier.destroy()
-            self.__notifier = None
+        self.teardown(exaile)
```
sbrubes: Must change the methods signature, probably will fail because "exaile" isn't defined (not tested).

rokm: Ooops, you're right. I missed that underscore in method's signature.

Issue #246, opened 2016-03-13
Migration to Python3


genodeftest:
Python 2 is in serios-bugfixes-only mode, python 3 is stable for a few years now. We need to run on version 3 eventually. This task will probably be the next big task after Exaile 4.0 is released with #3 being fixed.

# 1. Preparation:
Make sure all required libraries are available for Python 3.

- [x] migrate remaining dependencies. See [this wiki page](https://github.com/exaile/exaile/wiki/Python3-dependencies-migration-tasklist)
  - [x] most dependencies are available for python 3
  - [x] port to mox3 #244 
  - [x] [port to beautifulsoup4](http://www.crummy.com/software/BeautifulSoup/bs4/doc/#porting-code-to-bs4)
  - [x] make sure musicbrainzngs is available (on arch and other distros?)
  - [x] Clean up python code (see #371)
  - [ ] CDDB. This is an optional dependency for the plugin `cd`, so nothing important enough to block migration. See #608 for a nice way out.
  - [ ] spydaap. This is a dependency of the daapserver/daapclient plugin, see [egh/spydaap](https://github.com/egh/spydaap) or – in case they don't accept our patches – [exaile/spydaap](https://github.com/exaile/spydaap)
  - [x] pyHook

# 2. Convert Exaile source code

- [ ] See the official howto guide: https://docs.python.org/3/howto/pyporting.html
  - [x] #243 
  - [ ] ~#245~
  - [x] include changes from #289
  - [ ] include changes from #171
  - [x] replace sgmllib by HTMLParser or htmllib
  - [ ] use all remaining `__future__` statements in every source file
    - [ ] `absolute_import` (mostly in `plugins/` dir)
    - [ ] `division` (partially done as of 2017-08-01)
    - [ ] `print_function` (mostly done as of 2017-08-01)
    - [ ] `unicode_literals` (this is the hardest part)
  - [ ] `2to3` and other tools
  - [ ] after running `2to3`, start exaile with `python -3` to enable warnings and fix all the remaining bugs
  - [ ] Many, many small fixes
  - [x] update bsddb bindings (bsddb has been removed from python3), otherwise collection database can't be loaded. We need to have proper migration code for this case or use bsddb3 and open the shelf with `shelve.BsdDbShelf(bsddb3.hashopen(...))`.
    - See #399
  - …

[...]

PR #500, opened 2018-05-10
Convert to Python 3

remcohaszing:
First the code was converted using 2to3. Then many errors were fixed
using quick fixes.

The only goal was to make Exaile run using Python 3. No cleanup was done
whatsoever.

Most things seem to work. Most notably skipping a song doesn’t.

Tests are still untouched.

Refs #246

### DIFF
--- a/xl/settings.py
+++ b/xl/settings.py
@@ -28,7 +28,7 @@
     Central storage of application and user settings
 """
 
-from ConfigParser import (
+from configparser import (
     RawConfigParser,
     NoSectionError,
     NoOptionError
@@ -50,7 +50,7 @@ TYPE_MAPPING = {
     'B': bool,
     'L': list,
     'D': dict,
-    'U': unicode
+    'U': str
 }
 
 MANAGER = None
@@ -240,12 +240,16 @@ class SettingsManager(RawConfigParser):
             Turns a value of some type into a string so it
             can be a configuration value.
         """
-        for k, v in TYPE_MAPPING.iteritems():
+        # Special treatment is needed for boolean values, as bool is a
+        # subclass of int.
+        if isinstance(value, bool):
+            return 'B: ' + str(value)
+        for k, v in TYPE_MAPPING.items():
             if isinstance(value, v):
                 if v is list:
                     return k + ": " + repr(value)
-                elif v is unicode:
-                    return k + ": " + value.encode('utf-8')
+                elif v is str:
+                    return k + ": " + value
                 else:
                     return k + ": " + str(value)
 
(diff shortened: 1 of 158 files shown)


genodeftest:
Personally, I'm not opposed to this change, but we are not far from releasing Exaile 4.0 and I'd prefer to have a 4.0 release asap. In other words: I think this change should happen on a separate branch as it is changing many files and probably introducing a bunch of new bugs making the code far from a state where it can be released to a stable version.

Anyway, have you had a look at the smaller steps using `__future__` statements as proposed in #246? This way we could do incremental changes which are easier to review and will not put us so far from a working stable code base.

virtuald:
I'm not willing to consider this until Exaile 4 is released. There are only a few issues remaining: https://github.com/exaile/exaile/milestone/2

remcohaszing:
It makes sense to delay this until after Exaile 4 is released.

Most changes were autogenerated using `2to3`. I believe some of these changes could even be reverted.

Most remaining bugs were just about the removal of unicode encoding and decoding previously needed in Python 3.

I did not use `__future__` anywhere in this process. I didn’t need it.

I am willing to give this another try after the Exaile 4 release if you want. (This change didn’t take too long actually.) Just let me know. :smile:

xl/settings.py
```diff
@@ -240,12 +240,16 @@ def _val_to_str(self, value):
             Turns a value of some type into a string so it
             can be a configuration value.
         """
-        for k, v in TYPE_MAPPING.iteritems():
+        # Special treatment is needed for boolean values, as bool is a
+        # subclass of int.
+        if isinstance(value, bool):
+            return 'B: ' + str(value)
```
remcohaszing: This change actually fixes a bug which is consistent in Python 3, but still may happen randomly in Python 2.

sjohannes: Thanks; I've committed this fix as e677d2fc.

genodeftest:
Is it ok if I close this PR in favor of merging the [py3](https://github.com/exaile/exaile/tree/py3) branch into master (hopefully, some day)?

sjohannes:
Sorry that this PR didn't end up getting used. For our Python 3 port I wanted to base it on `futurize` rather than `2to3` as that seems to be the more developed converter, so that's what we used in the py3 branch. Rok and I have been testing the branch and fixing a lot of the subtle issues so we'll be going with that.

raveit65:
Looks like `Exaile 4.0.2 from 14 Jan` doesn't include the python3 port.
Can we expect a new release?
I'd like to include Exaile again for Mate-Compiz spin from fedora. We're are already in beta freeze for fedora 32, and it take some time to rebuild and get a new exaile version in repos before we reach release freeze.
Thank you

sjohannes:
If you're okay with the current state of the master branch, we can tag an alpha release so you can start working on the packaging. We can then do another alpha/beta release when you need to submit the final version.

I can't guarantee everything will be working by that time, but AFAIK only a few plugins are broken right now (@genodeftest would know more about this) so it shouldn't be too bad I think.

raveit65:
Np, when there are a few minimal issues. Main problem is the python2 is dropped from fedora 32, that means older versions can't be build anymore. An alpha release should make sure than no one has the idea to retire Exaile from fedora ;)
I will notice fedoras Exaile maintainer about a new release when it is done.
Can you ping me here?
Thanks.

genodeftest:
I'm fine with pushing the python3 stuff out soon. I have an open PR for the `cd` plugin which needs to be rebased to the current master but otherwise the plugins should still work. IMHO we can do an alpha.

@raveit65: apparently, I'm the new exaile maintainer for Fedora but if you want to help, please feel free to do so.

raveit65:
@genodeftest 
I am happy to test an alpha release at bodhi, please ping me.

Issue #306, opened 2016-11-12
Cover images are associated only with Album tag and not Artist+Album


sjohannes:
1. Have 2 files tagged with the same Album name but different Artist names.
2. Attach a cover image to the first file (through e.g. right click on cover &rarr; Fetch Cover).
3. Play the second file.

Expected: Exaile shows a different cover image, or no cover.

Actual: Exaile shows the first file's cover image.

I'm pretty sure this used to work correctly; I wonder what broke it.

sjohannes:
While we're at it, maybe also look at using albumartist if available.

sbrubes:
I've found this patch from July, 05.

Could you see if it's what you are looking for?

[306.txt](https://github.com/exaile/exaile/files/1295871/306.txt)

sbrubes:
Just **disregard what I've sent**.

I'll check it together with this one: https://github.com/exaile/exaile/issues/215

sbrubes:
[A fixed version.](https://github.com/exaile/exaile/files/1297172/306-fixed.txt)

sbrubes:
[Possible solution to #306.](https://github.com/sbrubes/exaile/commit/986370ccc25fce8b8ad043722b43c66bb41c94c9)

This time uses album as a switch off[*] and the keys 'albumartist', '__compilation' or 'date'.

 [*] This way an album with no album information will not have a key, and I'm assuming that.

sbrubes:
Changing this index will request "migration". #188

virtuald:
@sbrubes can you make a PR with whatever happens to be the right solution to this? Thanks!

Thinking about it, we should add a test for this too.

sbrubes:
@virtuald yes, I can

@sjohannes fine to you? https://github.com/exaile/exaile/issues/306#issuecomment-329291172

sjohannes:
Ok, I dug into our current code to see what it's doing, and it's quite surprising. Our covers.db key used to be the tuple `(artist, album)` in the simplest case. Then during the 0.3 rewrite back in 2009 it was changed to `('album', album)`, but that's bad because it removes the artist information — I completely missed this change throughout these years.

I like the patch but I want to change things around a bit. I'll post a PR later today and see what everyone thinks.

Issue #10, opened 2014-11-07
no cover-art for unknown albums in exaile v3.4.0


FiSHYsolutions:
a lot of my mp3-files are dj-sets which are not part of a dedicated album. my dir-tree usually looks like `<interpret>/<album>/<filename>.mp3` but in cases of files without an "album"-title the hierarchy switches to `<interpret>/<filename>.mp3`. all of my files are tagged by id3v2.4 and the tags always include a cover.jpg wich is placed in the same folder of the mp3, but of course it's embedded in the id3-tag, too. so far so good.

what is the problem? if i play a mp3-file without a album-tag (e.g. "unknown album"), exaile v3.4.0 doesn't show the cover.jpg (neither tag nor local). so i can fetch the cover by rightclicking the area, where the cover should be and then i choose (options-dialog) one of both covers (tag or local file). but if i press "pause", "next", "prev" or double click the track to play it from the very beginning, exaile losts the coverart of the specific file. there is no way to save the state, so every file without album-tag has spuriously no cover-art.

OS: xubuntu 14.04.1 lts

sjohannes:
This should be much better after #501 (which will be in Exaile 4.0). In the worst case, it's possible to specify the `musicbrainz_albumid` tag to override all of Exaile's heuristics for album covers.

PR #501, opened 2018-05-11
Covers: Support multiple albums having the same name

sjohannes:
See #306 for the motivation for this patch.

One of the exciting changes here is that it adds support for the `musicbrainz_albumid` tag. When this tag is available, we don't need to check anything else; it uniquely identifies the album.

This patch also incorporates @sbrubes's idea to use the `date` tag, because [one artist can release multiple albums with the same name](https://forums.macrumors.com/threads/itunes-two-albums-with-same-name-and-album-artist.1505017/).

So the possible track keys are, in order of preference (`[]` means the tag will be used iff it's available):

* musicbrainz_albumid
* album albumartist [date]
* __compilation [date]  # __compilation is a (directory, album) tuple
* album [artist] [date]

I've also changed the format of the output of this function. The function can return anything we like as long as it's hashable. Currently we return a tuple containing another tuple containing strings. My patch changes this to just a unicode string separated by null characters, e.g. `u'tag1\0value1\0tag2\0value2'`. It should make the dict lookup (hashing) marginally faster.

Here's what the new covers.db looks like, unpickled:

    {
        u"__compilation\x00/my/dir\x01my album\x00date\x002000":
            'localfile:file:///my/dir/cover.jpg',
        u'album\x00My Album\x00albumartist\x00My Albumartist\x00date\x002000':
            'tags:cover:0:file:///my/dir/file.mp3',
        u'musicbrainz_albumid\x00aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee':
            'localfile:file:///my/dir/cover.png',
    }

Any thoughts on all this?

Open questions regarding the migration:

1. What should we do with the existing cache? We could leave it as is, but that means having the existing entries (which are all stale) take up useless space. I vote [trashing](https://developer.gnome.org/gio/stable/GFile.html#g-file-trash) `$cachedir/covers`.
2. Should we reuse the directory/file name or use a new one, e.g. `$cachedir/covers2`, in case the user wants to use it in parallel with Exaile 3?
3. Do we do the migration in `covers.py` itself, or should we use a global 3→4 migration mechanism?

### DIFF
--- a/xl/covers.py
+++ b/xl/covers.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2008-2010 Adam Olsen
+# Copyright (C) 2018 Johannes Sasongko <sasongko@gmail.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -40,6 +41,7 @@ except ImportError:
     import pickle
 
 from xl.nls import gettext as _
+import xl.unicode
 from xl import (
     common,
     event,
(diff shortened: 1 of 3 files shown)


sbrubes:
The easier solution it's using the same directory and emptying it on Exaile's 4 first run.
It will have the cost of extra space when you run Exaile 3 and 4, but who wanna do that, could be aware about it.

I don't see how a migration 3->4 could be done, unless you do it progressive (as covers are requested).
But it should keep the old code, what it's quite unnecessary.

sbrubes:
As we are discussing that, maybe the issue #10, could be considered to be fixed. 

Using the dirname as db_key, when no album tag it’s found?

sjohannes:
I don't fully understand #10, but I think it's possible to do now with musicbrainz_albumid. It needs manual tagging but it's very precise and doesn't need Exaile to guess things.

sjohannes:
In the latest push I've added the migration code.

It's pretty tricky because there's not enough info in the old covers.db to create the new one. So the best I can do is migrate just the covers from tracks that exist in the user's Collection. Covers from non-Collection tracks are not migrated, and the image files associated with these tracks are also deleted from the cache.

PR #505, opened 2018-05-18
MPRIS plugin: improve conformity to spec

rokm:
This PR improves the conformity to MPRIS2 specification:
- PlayPause now starts the playback if it is stopped (instead of just toggling pause) - ```If playback is stopped, starts playback.```
- Next and Previous have effect only if player is in playing state - ```If playback is paused or stopped, it remains that way.```


sjohannes:
This is great; thanks!

Issue #495, opened 2018-03-22
What should double click do?


sbrubes:
At beta3, when you double click a playing row, it pauses.

Related to #441

sbrubes:
We probably should have two behaviors:

If paused: continue
If playing: restart

Need opinions.

genodeftest:
> If paused: continue
> If playing: restart

Sounds reasonable.

sbrubes:
That’s awesome @rokm! Thanks.

Closed by #507.

PR #507, opened 2018-05-18
Playlist: restart currently-playing track on activation (enter, double-click)

rokm:
This commit implements the behavior from issue #495. When pressing
enter key or double clicking on the track (assuming selection does
not change):
- if track is paused, it will continue
- if track is playing, it will restart from the beginning

The behavior of space key remains unchanged (i.e., it always toggles
pause, unless selection has changed), although its handling has
been moved to on_key_press_event() handler where we prevent it
from causing the 'row-activated' signal to be emitted.


virtuald:
Works for me.

PR #510, opened 2018-05-29
Playlist widget: implement sorting across multiple columns

rokm:
One of the functions I miss in exaile is the ability to sort the playlist by multiple colums, for example first by artist name and then by song title. Currently, sorting seems to be done only by the last-selected column, even though the track sorting functions appear to already support sorting by multiple fields.

This PR attemps to address this by implementing tracking of the sort column changes, and sorting the playlist using the obtained list of columns. This way, the last selected column is applied first, then the one that was selected before it, and so on. 

Note that the change of sort order applies to all columns and not just the last-selected one, as applying different sort orders on different colums would require larger changes across the codebase...

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -734,7 +734,9 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
         self.selection.set_mode(Gtk.SelectionMode.MULTIPLE)
 
         self._filter_matcher = None
-        
+
+        self._sort_columns = list(common.BASE_SORT_TAGS) # Column sort order
+
         self._setup_models()
         self._setup_columns()
         self.columns_changed_id = self.connect("columns-changed",
@@ -858,11 +860,9 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
         sortcol = self.get_sort_column()
         if sortcol:
             reverse = sortcol.get_sort_order() == Gtk.SortType.DESCENDING
-            sort_by = [sortcol.name] + list(common.BASE_SORT_TAGS)
         else:
             reverse = False
-            sort_by = list(common.BASE_SORT_TAGS)
-        return (sort_by, reverse)
+        return (self._sort_columns, reverse)
 
     def play_track_at(self, position, track):
         '''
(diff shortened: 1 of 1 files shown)


xlgui/widgets/playlist.py
```diff
@@ -734,7 +734,9 @@ def __init__(self, playlist, player):
         self.selection.set_mode(Gtk.SelectionMode.MULTIPLE)
 
         self._filter_matcher = None
-        
+
+        self._sort_columns = [] # Column sort order
```
virtuald: Seems like it would be easier to just initialize this to `list(common.BASE_SORT_TAGS)`?

xlgui/widgets/playlist.py
```diff
@@ -858,10 +860,16 @@ def get_sort_by(self):
         sortcol = self.get_sort_column()
         if sortcol:
             reverse = sortcol.get_sort_order() == Gtk.SortType.DESCENDING
-            sort_by = [sortcol.name] + list(common.BASE_SORT_TAGS)
+            if not self._sort_columns:
+                sort_by = [sortcol.name] + list(common.BASE_SORT_TAGS)
+            else:
+                sort_by = self._sort_columns
```
virtuald: This seems wrong. What happens if the sort column didn't exist previously?

xlgui/widgets/playlist.py
rokm: That might indeed be a better idea.

xlgui/widgets/playlist.py
rokm: If there was no sort column, would we not be in the other branch (as sortcol would be None)?

virtuald: What I meant was what if the column itself did not previously exist? Then it couldn't be in the _sort_columns list (which means it would get skipped here?)

rokm: Is there a scenario where that happens? I originally tried to safe-guard against _sort_columns being empty in both branches because I was not sure if it could happen that get_sort_by() would be called and have a valid sortcol without on_column_clicked() happening first (which is where _sort_columns gets modified)...

I am asking because after revising the commit to initialize _sort_columns with ```list(common.BASE_SORT_TAGS)```, I also tried to simplify that part so that _sort_columns is always returned and sortcol is used only to get sorting direction. As far as I can tell, sortcol can be valid only if user clicked on a column and on_column_clicked() was executed, because that is where sort indicator gets set (and that is what get_sort_column() in get_sort_by() is checking for).

virtuald:
Thanks, works for me!

Issue #447, opened 2017-10-31
Plugin daapclient: integer `0` gets converted to `None` – Why?


genodeftest:
Under some circumstances `0` gets converted to `None` and I have no clue why. Can you please explain this to me?

Steps to reproduce:
1. start an instance of exaile as daap server
1.1. enable plugin "daapserver"
1.2. add some files to music library, preferably just a few (2 is enough)
2. start another instance of exaile as daap client
2.1. enable plugin "daapclient"
2.2. connect to the other instance (Menu → Tools → Connect to DAAP…)
2.3. drag all files to a playlist
2.4. in the playlist, show column "Location"
2.5. try to play all files

### Expected Behavior
at 2.4., all URLs should look like this: `http://<server>:<port>/databases/<dbid>/items/<id>.<type>?session-id=<sessionid>` with `<id>` being any number.
at 2.5., all files should play fine.

### Current Behavior
at 2.4., one file has `<id>` set to `None`. From a look at all the files one could guess it should have `0` instead.
at 2.5., the file with incorrect id does not play.

### Dirty hack
In [plugins/daapclient/__init__.py:518ff in `DaapConnection.convert_list()`](https://github.com/exaile/exaile/blob/master/plugins/daapclient/__init__.py#L518), replace 

```
                # http://<server>:<port>/databases/<dbid>/items/<id>.<type>?session-id=<sessionid>

                uri = "http://%s:%s/databases/%s/items/%s.%s?session-id=%s" % \
                    (self.server, self.port, self.database.id, tr.id,
                    tr.type, self.session.sessionid)
```

by
```
                track_id = tr.id
                if track_id is None:
                    track_id = 0
                # http://<server>:<port>/databases/<dbid>/items/<id>.<type>?session-id=<sessionid>
                uri = "http://%s:%s/databases/%s/items/%i.%s?session-id=%s" % \
                    (self.server, self.port, self.database.id, track_id,
                     tr.type, self.session.sessionid)
```
This is a quite dirty hack though, so I'd like to find a better solution.

### Additional info:
By debugging I found where the actual issue happens:
When `DaapConnection.convert_list()` accesses `tr.id`, it calls `DAAPObject.getAtom()`, which correctly returns the `int` `0`. When checking the value in  `DaapConnection.convert_list()` though, it is `None`. I don't know how that can happen.

__[Edit]__
I am 100% sure that the correct value is transmitted to the client because I have wireshark output.
__[/Edit]__

### Environment
[...]

PR #511, opened 2018-06-02
DAAP plugins: fix the retrieval of atoms with integer value 0

rokm:
When retrieving the atom value with DAAPObject.getAtom(), None will
be incorrectly returned when the code corresponds to a child
atom with integer value 0.

This is because the validity of the returned value is checked via
boolean coercion instead of explicitly checking for None (and thus
0 evaluates to False, same as a None would have).

Fixes issue #447.


virtuald:
Thanks!

PR #512, opened 2018-06-09
Reformat code with black

virtuald:
I'm not necessarily advocating that we have to continue on this path, but I'm opening this PR to start the conversation about whether we should do this or not. It is nice to not have to worry about style. From my initial look at this, for the most part it seems to generate better looking code with only a few exceptions (and in those places I've added `# fmt: on/off` directives to disable it). I also modified the travis configuration to enforce usage of the formatter.

I ran black like so:

```
black -S *.py plugins/ xl/ xlgui/ tests/
```

One thing it does reformat is our custom of doing imports like so:

```
from xl import (
  common
  nls
)
```

to 

```
from xl import common, nls
```

... honestly, I don't really care much about that particular style point anyways, so seems fine to me.

One thing I do care about is single vs double quoted strings. I used the `-S` option of black to tell it to not force everything to double quotes.

Thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 PYTHON2_CMD   ?= python2
-PYTEST         = py.test
+PYTEST        ?= py.test
+BLACK         ?= black
 
 PREFIX         = /usr/local
 EPREFIX        = $(PREFIX)
@@ -32,7 +33,7 @@ EXAILEMANDIR   = $(DESTDIR)$(MANPREFIX)/man
 .PHONY: all all_no_locale builddir compile make-install-dirs uninstall \
 	install install_no_locale install-target locale install-locale \
 	plugins-dist manpage completion clean pot potball dist check-doc test \
-	test_coverage lint_errors sanitycheck
+	test_coverage lint_errors sanitycheck format
 
 all: compile completion locale manpage
 	@echo "Ready to install..."
@@ -260,3 +261,9 @@ lint_errors:
 	-pylint -e --rcfile tools/pylint.cfg xl xlgui 2> /dev/null
 
 sanitycheck: lint_errors test
+
+format:
+	$(BLACK) -S *.py plugins/ xl/ xlgui/ tests/
+
+check_format:
+	$(BLACK) --check -S *.py plugins/ xl/ xlgui/ tests/
--- a/plugins/cd/__init__.py
+++ b/plugins/cd/__init__.py
@@ -43,6 +43,7 @@ import cdprefs
 try:
     import DiscID
     import CDDB
+
     CDDB_AVAIL = True
 except Exception:
     CDDB_AVAIL = False
@@ -61,7 +62,6 @@ CDROM_DATA_TRACK = 0x04
 
 
 class CdPlugin(object):
-
     def enable(self, exaile):
         self.exaile = exaile
         self.hal = None
@@ -94,6 +94,7 @@ class CdPlugin(object):
 
     def get_preferences_pane(self):
         return cdprefs
+
 
 plugin_class = CdPlugin
 
@@ -151,7 +152,6 @@ class CDTocParser(object):
 
 
 class CDPlaylist(playlist.Playlist):
-
     def __init__(self, name=_("Audio Disc"), device=None):
         playlist.Playlist.__init__(self, name=name)
 
@@ -172,14 +172,15 @@ class CDPlaylist(playlist.Playlist):
         for count, length in enumerate(lengths):
             count += 1
             song = trax.Track("cdda://%d/#%s" % (count, self.device))
-            song.set_tags(title="Track %d" % count,
-                          tracknumber=str(count),
-                          __length=length)
+            song.set_tags(
+                title="Track %d" % count, tracknumber=str(count), __length=length
+            )
             songs[song.get_loc_for_io()] = song
 
         # FIXME: this can probably be cleaner
-        sort_tups = sorted([(int(s.get_tag_raw('tracknumber')[0]), s)
-                            for s in songs.values()])
+        sort_tups = sorted(
+            [(int(s.get_tag_raw('tracknumber')[0]), s) for s in songs.values()]
+        )
         sorted_elements = [s[1] for s in sort_tups]
 
         self.extend(sorted_elements)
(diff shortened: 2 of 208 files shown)


plugins/cd/__init__.py
```diff
@@ -151,7 +152,6 @@ def get_track_lengths(self):
 
 
 class CDPlaylist(playlist.Playlist):
-
```
genodeftest: This type of change (remove empty line directly after the class definition) is not good imho. It is inconsistent with any other method. As far as I remember it also conflicts with the PEP8 rules.

genodeftest:
I would skip everything in `plugins/daapclient/spydaap/` as that's going away with #445.

It looks nice overall, improving code readability a bit. I have no strong opinion on this though.

genodeftest:
The changes do not seem to change any logic (as sometimes happens with autopep8) so that's good.

plugins/cd/__init__.py
virtuald: It appears that it only does this if the class doesn't have a docstring (which is generally recommended anyways).

For examples of the types of reformatting it does on various class bodies, see the tests at https://github.com/ambv/black/blob/master/tests/data/class_methods_new_line.py

genodeftest:
I like the matching parenthesis ([Egyptian Brackets](https://blog.codinghorror.com/new-programming-jargon/)). It's better than the old style of having code after the opening bracket even if the content spans multiple lines.

plugins/mpris2/mprisobject.py
```diff
@@ -217,13 +225,7 @@ def Identity(self):
...
     def SupportedMimeTypes(self):
         # Taken from exaile.desktop
-        mimetypes = 'audio/musepack;application/musepack;application/x-ape;' \
-            'audio/ape;audio/x-ape;audio/x-musepack;application/x-musepack;' \
-            'audio/x-mp3;application/x-id3;audio/mpeg;audio/x-mpeg;' \
-            'audio/x-mpeg-3;audio/mpeg3;audio/mp3;audio/x-m4a;audio/mpc;' \
-            'audio/x-mpc;audio/mp;audio/x-mp;application/ogg;' \
-            'application/x-ogg;audio/vorbis;audio/x-vorbis;audio/ogg;' \
-            'audio/x-ogg;audio/x-flac;application/x-flac;audio/flac'
+        mimetypes = 'audio/musepack;application/musepack;application/x-ape;' 'audio/ape;audio/x-ape;audio/x-musepack;application/x-musepack;' 'audio/x-mp3;application/x-id3;audio/mpeg;audio/x-mpeg;' 'audio/x-mpeg-3;audio/mpeg3;audio/mp3;audio/x-m4a;audio/mpc;' 'audio/x-mpc;audio/mp;audio/x-mp;application/ogg;' 'application/x-ogg;audio/vorbis;audio/x-vorbis;audio/ogg;' 'audio/x-ogg;audio/x-flac;application/x-flac;audio/flac'
```
genodeftest: This line is way too long which does not make any sense to me. Can we prevent that from happening?

genodeftest:
In some cases it seems to do weird stuff about the line length.

xl/metadata/asf.py
```diff
@@ -33,14 +33,14 @@ class AsfFormat(BaseFormat):
...
     tag_mapping = {
-        "artist"        : "Author",
-        "album"         : "WM/AlbumTitle",
-        "title"         : "Title",
-        "genre"         : "WM/Genre",
-        "tracknumber"   : "WM/TrackNumber",
-        "date"          : "WM/Year",
-        "albumartist"   : "WM/AlbumArtist",
-        "grouping"      : "WM/ContentGroupDescription"
+        "artist": "Author",
```
genodeftest: This breaks the indentation. I won't mind but as far as I remember one of you (@sjohannes or @virtuald) did.

genodeftest: This repeats multiple times.

genodeftest:
The only thing I don't like at all is that it creates a few very long lines like in https://github.com/exaile/exaile/pull/512/files/d7392472a8f9b8db480bdf9d273caf82becad90f#diff-8e3534fc9fb5a2f7818196dbb0314156 or when assigning many variables during construction of `Gtk.Template`. Everything else I've seen so far (I've stopped at about 25% of the changes) looks very good, good or acceptable to me.

xl/metadata/asf.py
virtuald: Good catch, would have to add fmt off directives around this.

plugins/mpris2/mprisobject.py
virtuald: Looks like a bug. Filed upstream at https://github.com/ambv/black/issues/330

xl/metadata/asf.py
genodeftest: There are a few examples of this all over the code base.

plugins/mpris2/mprisobject.py
genodeftest: Also affects `Gtk.Template` stuff as noted above. See https://github.com/ambv/black/issues/330#issuecomment-395995308 for an example.

virtuald:
Reflowed those sections. I don't mind the 88 line length, but I would be ok with 99. Would not want 79.

virtuald:
Heh, encountered another bug with black: https://github.com/ambv/black/issues/334

... well, it _is_ still marked as beta :man_shrugging:

genodeftest:
In https://github.com/ambv/black/issues/330#issuecomment-396004666, I'd go with suggestion 1.

sjohannes:
For GtkTemplate, wouldn't this be better?

```py
(
    found_label,
    playlists,
    replace,
    replace_entry,
    search_entry,
    tracks_list,
) = GtkTemplate.Child.widgets(6)
```

It wastes vertical space if you have lots of things in the list, but that's a problem with Black in general.

virtuald:
Your solution does work and I agree that it's very Blackish. I've filed a separate bug asking that he reconsider his stance. https://github.com/ambv/black/issues/336

Makefile
```diff
@@ -1,5 +1,6 @@
 PYTHON2_CMD   ?= python2
 PYTEST         = py.test
+BLACK          = black
```
sjohannes: I think you either forgot to remove this line or forgot to use the variable below.

virtuald:
Boom.

PR #513, opened 2018-06-10
Add option to use a 'dark theme' if available

virtuald:
Tried this on Windows 7... it doesn't look that terrible. I suspect it would look better on Windows 10 if you were using a dark Windows theme already, but I don't have such a system available to me.

On Linux it doesn't seem that I can specify the non-dark theme anymore... and when I pick a theme that doesn't have a corresponding dark theme, this doesn't seem to do anything.


virtuald:
Tried it on OSX, looks pretty ok there.

dangmai:
Here's what it looks like on Win 10, I think it's pretty decent:
exaile-win-10-dark

dangmai:
With music:

exaile-win-10-dark-2

PR #515, opened 2018-06-11
daapclient: explicitly convert panel's name to string

rokm:
Otherwise, it will be stored to the list of opened panels as
dbus.String() type, causing the following error:
NameError: name 'dbus' is not defined
when restoring the GUI on the next program start.


sjohannes:
Thanks!

Issue #76, opened 2015-04-27
Stutter problems with Preview Plugin on Windows


virtuald:
Various users I know have reported this issue. I can't currently reproduce it on my machines. A friend's computer showed the issue when on battery power, but not when plugged in.

virtuald:
I had originally thought that there may be some problem with the patched version of the directsoundsink that we're using, but I just reviewed the patch and find it unlikely that we introduced the bug.

virtuald:
I tried a bunch of things on Linux (mostly involving adding sleeps to various places) and wasn't able to get the stutter to occur there. 

I popped up a Windows 7 VM, and it stutters even without the preview device. Using Windows Media Player to play the same file, no stutter. This would seem to point to an inefficiency on our part somewhere... or it's a gstreamer problem. Going to try playing the files via gst-launch to verify that it's our pipelines.

virtuald:
Ok, so I ran `gst-launch playbin2 uri=file:///C:/path/to/mp3/file` and I get skipping there too when the VM is doing things. Windows Media Player still doesn't skip at all. Might be a gstreamer problem, in which case there might not be much we can do about it.

virtuald:
Ok, so this is interesting, using Process Explorer I looked at the priorities of the threads that Windows Media Player and gst-launch were using internally, and here's what I've got:

Windows Media Player:
* Audio threads: 1.5% CPU, base priority 5, dynamic priority 21

gst-launch:
* Audio threads: 0.6% CPU... but there are a lot of them
* One that originates in dsound.dll is base prio 5, dynamic prio 22
* Another that originates in dsound:DirectSoundCreate is base prio 15, dynamic 15
* Everything else seems to be base prio 8, dynamic prio 8. 

As a test, I set the priority of the entire gst-launch process to 'realtime' and the stutter problem went away, even when triggering system duress. It also seems to help Exaile's stutter also... but I haven't tried this on bare metal yet.

A possible solution might be trying to detect when gstreamer launches threads, and setting their priorities higher than what GST might do. It might get a bit hacky though... presumably this has been fixed in later versions of GST, but we're stuck with 0.10 for now.

virtuald:
It appears that trying to enhance the priority of the entire process on bare metal does not fix the problem on bare metal. However, I was also to reproduce some stutter issues on bare metal using gst-launch, whereas it was very difficult to get VLC to stutter. Sigh.

[...]

PR #518, opened 2018-06-24
Use GstDeviceMonitor for all the devices!

virtuald:
I know, we want to do a 4.0 release, but this gets rid of a lot of nasty code! And as an added bonus, users can try out the WASAPI sinks if they're suffering from #76

### DIFF
--- a/xl/player/gst/sink.py
+++ b/xl/player/gst/sink.py
@@ -108,6 +108,13 @@ def _gst_device_autodetect():
             if not device_id:
                 continue
             
+            display_name = device.get_display_name()
+            if hasattr(device.props, 'properties'):
+                api = device.props.properties.get_string('device.api')
+                if api:
+                    display_name = '%s [%s]' % (display_name, api)
+
+            yield (display_name,
                    device_id,
                    device.create_element)
     finally:
--- a/xl/player/gst/sink.py
+++ b/xl/player/gst/sink.py
@@ -65,10 +65,11 @@ SINK_PRESETS = {
         'name': "DirectSound",
         'pipe': 'directsoundsink'
     },
-    "wasapi": {
-        'name': "WASAPI",
-        'pipe': 'wasapisink'
-    },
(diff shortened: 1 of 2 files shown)


xl/player/gst/sink.py
```diff
@@ -61,35 +61,65 @@
         "name": "JACK",
         "pipe": "jackaudiosink"
     },
+    "directsoundsink": {
```
sjohannes: I think just `directsound` would be better here, to match the other names.

Also, a WASAPI entry would make sense as well.

virtuald:
Yes, it only works on the automatic sink, as you can't explicitly which sink you want to use when you use the `Gst.DeviceMonitor` stuff (which is why I added the device.api display stuff).

xl/player/gst/sink.py
virtuald: The key is stored in settings, so it would break backwards compatibility to change it.

virtuald:
Hm. Looks like Exaile freezes when I start+stop+start a song when using a WASAPI sink (doesn't seem to occur with a directsound sink).

virtuald:
Looks like it's something internal to the WASAPI sink, I get the following error when I set `GST_DEBUG=2`

```
0:00:07.205249000  1456    32bb880 ERROR             wasapisink gstwasapisink.c:687:gst_wasapi_sink_reset:<player-audiosink> IAudioClient::Stop failed (1): Incorrect function.
```
Then when I start playback promptly freezes:

```
0:00:12.997217000  1456    46b43f0 ERROR             wasapisink gstwasapisink.c:613:gst_wasapi_sink_write:<player-audiosink> IAudioClient::Start failed (88890005): AUDCLNT_E_NOT_STOPPED
```
This bug has been reported at https://bugzilla.gnome.org/show_bug.cgi?id=796581 by someone else. I suspect the appropriate fix for it is covered by the patches in https://bugzilla.gnome.org/show_bug.cgi?id=796354

PR #522, opened 2018-06-28
DAAP server plugin: cleanup

rokm:
This PR removes couple of files that appear to have been left over by the spydaap de-bundling process, and accounts for the renaming of the spydaap.zeroconf module to spydaap.zeroconfimpl in upstream spydaap.


virtuald:
Thanks!

Issue #530, opened 2018-07-26
Mini mode: issues with switching playlist


rokm:
Switching a playlist after enabling mini mode results in an error, which prevents the update of playlist in mini mode's view.

### Steps to reproduce

1. Create two playlists
2. Play a track from first playlist
3. Enable Mini mode plugin and switch to mini mode; the playlist shown in mini mode's view should be the first playlist
4. Switch back from mini mode, and switch to second playlist by playing a track from it

### Outcome

```
Traceback (most recent call last):
  File "/home/rok/Projects/exaile/plugins/minimode/controls.py", line 812, in update_playlist
    columns = self.view.get_model().column_names
AttributeError: 'TreeModelFilter' object has no attribute 'column_names'
```
After switching back to mini mode, the first playlist is still shown there.

Stopping the playing track results in the same backtrace as in Issue #529.

Issue #529, opened 2018-07-26
Mini mode: tracebacks when updating music collection after enabling mini mode


rokm:
Initializing or updating music collection after enabling the mini mode results in a slew of tracebacks.

### Steps to reproduce

1. Enable Mini mode plugin
2. Switch to Mini mode and back
3. Initialize the Collection or add more music to it

### Outcome

The above steps result in  a slew of tracebacks (where ```TRACK NAME```, ```ARTIST```, and ```ALBUM``` correspond to actual fields from the newly-added tracks):

```
ERROR   : Event callback exception caught!
Traceback (most recent call last):
  File "/home/rok/Projects/exaile/xl/event.py", line 361, in _emit
    event.type, event.object, event.data, *cb.args, **cb.kwargs
  File "/home/rok/Projects/exaile/plugins/minimode/controls.py", line 955, in on_track_tags_changed
    track_position = playlist.index(track)
  File "/home/rok/Projects/exaile/xl/playlist.py", line 1689, in index
    return self.__tracks.index(item, start)
  File "/home/rok/Projects/exaile/xl/common.py", line 767, in index
    return self.__list.index(i, start)
ValueError: <Track u'TRACK NAME' by u'ARTIST' from u'ALBUM'> is not in list
```

Issue #528, opened 2018-07-26
Issues with removal of tracks from playlist once Mini mode is enabled


rokm:
After enabling mini mode, removal of multiple tracks from a playlist results in an error.

### Steps to reproduce

Assuming a completely fresh Exaile setup, i.e., with separate all-data-dir:

1. Initialize Music collection
2. Create a playlist with three tracks
3. Play a track (this seems to be necessary to avoid another issue with Mini mode)
4. Enable Mini mode plugin
5. Switch to Mini Mode and back (to create Mini Mode's PlayList view)
6. Remove the last two tracks from the player's playlist

### Outcome

```
ERROR   : Event callback exception caught!
Traceback (most recent call last):
  File "/home/rok/Projects/exaile/xl/event.py", line 361, in _emit
    event.type, event.object, event.data, *cb.args, **cb.kwargs
  File "/home/rok/Projects/exaile/xlgui/widgets/playlist.py", line 1924, in on_tracks_removed
    self.remove(self.iter_nth_child(None, position))
TypeError: Argument 1 does not allow None as a value
```

As the result of the error, one of the playlist views (either in main player or in Mini mode) will retain the last track, which was supposed to be removed.

In the above scenario, we have one playlist object, and two PlayListView objects with their corresponding PlayListModel objects. When we remove the tracks, the tracks_removed signal is emitted by the playlist object, and on_tracks_removed() handler in both PlayListModels is called.

However, because the handler reverses the received track list, the second handler is effectively called with the modified (reversed) list, and as such, its subsequent attempt to reverse the list again results in the original ordering. Which then causes the error because we are removing elements from its internal list in forward instead of reverse order.

Using reverse iterator instead of reversing the list solves the issue.

PR #531, opened 2018-07-26
Mini mode fixes

rokm:
This PR provides fixes for Issues #528, #529, and #530.

### DIFF
--- a/plugins/minimode/controls.py
+++ b/plugins/minimode/controls.py
@@ -952,9 +952,12 @@ class PlaylistButtonControl(Gtk.ToggleButton, BaseControl, QueueAdapter):
             Updates the button on tag changes
         """
         playlist = self.view.playlist
-        track_position = playlist.index(track)
 
-        if track in playlist and track_position == playlist.current_position:
+        if track not in playlist:
+            return
+
+        track_position = playlist.index(track)
+        if track_position == playlist.current_position:
             self.label.set_text(self.formatter.format(track))
 
     def on_option_set(self, event, settings, option):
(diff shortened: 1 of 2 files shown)


rokm:
There is no iteration error because iterating itself is not problematic - what we do inside it is.

```tracks``` is the list (```MetadataList```) of removed tracks received from the underlying playlist object, i.e., a list of tuples that contain position (index) in the playlist and reference to track object. This list seems to be sorted in the ascending order (e.g., ```[ (1, trackObj1), (2, trackObj2) ]``` if we are removing last two of three tracks).

In the ```on_tracks_removed``` handler, ```PlaylistModel``` uses this list to prune its internal list of tracks:
```python
def on_tracks_removed(self, event_type, playlist, tracks):
    tracks.reverse()
    for position, track in tracks:
        self.remove(self.iter_nth_child(None, position))
```

Because the ```PlaylistModel```'s internal list of tracks is modified on-the-fly, the removal needs to be done in backwards direction (i.e., we are removing tracks from the back to the beginning of the playlist), otherwise we would be invalidating the indices as we go. That is the purpose of the ```tracks.reverse()```.

If we have two ```PlaylistModel```s connected to the playlist, the first handler reverses the list, and prunes its internal list without issues. However, because list object is effectively passed by reference, the second handler then receives the list that has already been reversed. Its reversal attempt prior to iteration therefore restores the original ascending order.

So, if the playlist had three tracks, and our list of removed tracks was ```[ (1, trackObj1), (2, trackObj2) ]```, we remove track with index 1 first, and index 2 becomes invalid because our internal list now has length of two. So when trying to remove track with index 2, ```self.iter_nth_child()``` returns ```None```, which then raises exception in ```self.remove()```.

By using reverse iterator, we avoid modifying the list itself, and both ```on_tracks_removed``` handlers prune their internal lists with correct order of tracks positions.

PR #536, opened 2018-08-24
daapclient: provide a user agent

rokm:
Prevents the client from crashing forked-daapd server
(ejurgensen/forked-daapd#571). Even if it did not crash the
server, providing the user agent will allow forked-daapd to cache the
request/query results.


plugins/daapclient/client.py
```diff
@@ -42,7 +42,7 @@ def _get_response(self, r, params={}, gzip=1):
 
         log.debug('getting %s', r)
 
-        headers = {'Client-DAAP-Version': '3.0', 'Client-DAAP-Access-Index': '2'}
+        headers = {'User-Agent': 'Exaile/4.x', 'Client-DAAP-Version': '3.0', 'Client-DAAP-Access-Index': '2'}
```
virtuald: Use exaile.get_user_agent_string() instead. 

https://github.com/exaile/exaile/blob/8fb0ebb4ea108792d38e2b8ad355b27d70e1d246/xl/main.py#L880

sjohannes:
The CI failure is due to long lines, I think. We use [Black](https://github.com/ambv/black) to do the formatting, which has 88 character limit per line. To satisfy it you'll have to reformat the code like so:

```py
headers = {
    'User Agent': ...,
    ...: ...,  # note: there must be a comma here
}
```

If you have Black installed you can run `make format` to do it automatically on the whole Exaile tree.

plugins/daapclient/client.py
rokm: Ah, that's indeed much better - thanks!

PR #537, opened 2018-08-26
Added random playback

0mark:
A few words to explain:

1st) Random is not the same as shuffle
This is often stated, but actually wrong. While "random" play selects the next track at random out of all tracks available (in playlist, queue or whatever), "shuffle" linearly picks the next track out of a randomly ordered list. The main difference is that shuffle never plays the same song again until the shuffled list is completely played through or the list is reshuffled. In addition, usually the shuffled list is also displayed to the user, while random simply shows the default order (whatever that may be).

2nd) Random is neither "better" than shuffle nor "what you really want if you give it a tray"
First and foremost, this a question of taste. And secondly, while shuffle certainly has advantages on short playlists (where random can easily lead to a perceived preference for one or a few songs due to local clustering in the distribution as well as psychological reasons), it has disadvantages on really big playlists (many thousands of songs). Each song played is not played again for days worth of playing. Granted, on the long run songs are played equally often and on equally distance in both ways. But locally, a song just played is played again in (on average) n/2 songs (n is the total amount of songs), while it is played again in (n-m)+(n/2) (where m is the amount of songs out of the current shuffle already played) if the list is re-shuffled after playing through and (n-m)+m if not. The probability of a particular song to be played next for random play is 1/n, but for shuffle its 0, if the song was already played or 1/(n-m) if not (actually: its 1 if its the next song in the shuffled list and 0 otherwise. But assumed the user does not know the shuffled list...).

3rd) The shuffe mde in exile is not shuffle
Its random with consume, but that effectively the same. Except that its not, at least in the sense that the playlist is always displayed in the default order. I consider that an advantage over real shuffle, but others might not.

So, in short:
Exaile lacked "normal" random play mode. Here is one :)

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -1134,7 +1134,7 @@ class Playlist(object):
                 t = [x for x in self if x.get_tag_raw('album') == album]
                 t = trax.sort_tracks(['tracknumber'], t)
                 return self.__tracks.index(t[0]), t[0]
-        if mode == 'random':
+        elif mode == 'random':
             try:
                 return random.choice(
                     [
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -1134,17 +1134,17 @@ class Playlist(object):
                 t = [x for x in self if x.get_tag_raw('album') == album]
                 t = trax.sort_tracks(['tracknumber'], t)
                 return self.__tracks.index(t[0]), t[0]
-	if mode == 'random':
-	    try:
+    if mode == 'random':
+        try:
                 return random.choice(
                     [
                         (i, self.__tracks[i])
                         for i, tr in enumerate(self.__tracks)
                     ]
                 )
-	    except IndexError:
-		return -1, None
-	else:
+        except IndexError:
+        return -1, None
+    else:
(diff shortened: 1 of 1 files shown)


0mark:
Apparently i mixed up tabs and spaces. I will see to fix that, although i have no idea how to alter a pull request.

genodeftest:
Thanks for the  pull request!

> Apparently i mixed up tabs and spaces. I will see to fix that, although i have no idea how to alter a pull request.

If you're using git, you can create a new git commit and push it with `git push --force`.

virtuald:
This patch seems fine to me, but the formatting still needs to be fixed.

Since formatting is syntax in python, it's important to be consistent. You can run `make check` to use the black autoformatter to format the code.

0mark:
The Last Test fails with, but i cant see why.

virtuald:
We use the black autoformatter, and your code didn't match. In the future, with black installed, you can fix that via 'make format'.

I merged your changes manually. Thanks!

0mark:
Awesome, thank you :)

PR #541, opened 2018-09-11
Improve metadata retrieval for remote tracks (DAAP)

rokm:
This PR improves the metadata retrieval for tracks in remote locations, with the main aim of enhancing the collection views for remote (DAAP) shares.

The first commit adds parsing of track year/date from the remote stream's metadata when the track is played. Like all metadata set at the playback time, this information serves as a fall-back for the case when it is not retrieved otherwise (e.g., at the moment, daapclient plugin does not retrieve track year).

The second commit modifies daapclient plugin to parse track year (which is already in the list of retrieved DAAP atoms), with the goal of enabling "Date - Artist", "Date - Album", and "Artist - (Date - Album)" views in collection pane.

The third commit similarly adds retrieval of disc number, which improves track sorting within collection views (and playlists) for multi-disc albums.

As noted in the source code, retrieving more metadata means greater communication overhead; however I would argue that we should be retrieving all metadata that we use for filtering and sorting. Otherwise, this functionality becomes fully available only after all tracks are played at least once (and metadata is retrieved from the stream itself), which is very impractical.

### DIFF
--- a/plugins/daapclient/__init__.py
+++ b/plugins/daapclient/__init__.py
@@ -537,6 +537,7 @@ class DaapConnection(object):
             'artist': 'asar',
             'album': 'asal',
             'tracknumber': 'astn',
+            'date': 'asyr',
         }
         #            'genre':'asgn','enc':'asfm','bitrate':'asbr'}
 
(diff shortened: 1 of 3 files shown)


virtuald:
Will merge this after a 4.0 release. Should be this weekend.

virtuald:
Nevermind, we'll need to do another RC due to the disk thing and the other thing, so pulling this in too.

PR #542, opened 2018-09-14
lastfmlove: various fixes to make it work again

eri-trabiccolo:


eri-trabiccolo:
Hi @rokm yeah I see that's about the black formatter, but I'm not sure what should I do then, since, as you noted, the code guidelines say the opposite. Is then the one-line import the new official standard?

rokm:
The code was auto-formatted using black in the beginning of this summer (728463362), and the code guidelines have very likely not been updated to reflect that yet. In that commit, the multi-line imports have been replaced with one-liners (including the threading one that you changed back to old style). 

If you have black installed, you can run ```make format``` before committing your changes and avoid worrying about the standard altogether. At the moment, I would say that black's preferences are the official standard (and if they are to be changed, they will be automatically (re-)applied to the whole codebase again).

eri-trabiccolo:
Ok thanks @rokm 
will push another commit with the required fixes.

virtuald:
In the future @eri-trabiccolo running `make format` will fix the code for you if you have black installed. The latest version of the code guidelines have the correct advice, it's the 'stable' version that have old advice.

eri-trabiccolo:
thanks @virtuald 
and yeah, sorry, I looked at the code guidelines of the 'stable' version.

PR #547, opened 2018-09-19
metadata: lock computation of reverse mappings

### LINKED ISSUE #539: MP3 tags not read
MP3 tags not read

Add folder to collection.

### Expected Behavior
Tags should be properly imported for all supported media files.

### Current Behavior
Tags for ogg and flac files work just fine, but all MP3 files show up without any tag information.

### Further information
* Probably not a problem with mutagen per se since other programs that use it (I tested puddletag) work without problems.
* Interestingly, if I edit the tag info within exaile, it shows up correctly afterwards.

### Environment
* Operating System and version: Debian testing (buster), x86_64
* Exaile Version: 4.0.0-rc2
* mutagen-1.41.1

virtuald:
- Fixes #539

### DIFF
--- a/xl/metadata/_base.py
+++ b/xl/metadata/_base.py
@@ -26,6 +26,7 @@
 
 from collections import namedtuple
 import copy
+import threading
 
 import logging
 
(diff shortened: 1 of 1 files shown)


virtuald:
I believe this is needed in addition to #549, just in case the race happens.

PR #549, opened 2018-09-22
Metadata: avoid directly instantiating BaseClass for unsupported formats

rokm:
Direct instantiation of BaseClass will set '_reverse_mapping' field
in the BaseClass, and thus prevent format-specific initialization
of '_reverse_mapping' for child classes that may get instantiated
afterwards.

Instead, use a DummyFormat class. Fixes issue #539.


virtuald:
Thanks, your explanation makes sense!

PR #556, opened 2018-09-28
Whitespace removed

comradekingu:


sjohannes:
Thanks for the patch. We'll probably wait until 4.0 is out before applying this so it doesn't break current translations.

sjohannes:
I'm doing some manual editing on the translations so I've decided to make this change now as well.

Thanks again for the patch.

PR #559, opened 2018-09-30
Automatically generate .desktop and AppData files

Carmina16:
Allows the strings from these data files to be extracted into the template file, and the translations to be put back during the install.

Requires intl-tool.


sjohannes:
I think xgettext now supports dealing with .desktop and .appdata.xml files, so there should be a way to do this without intltool and its intermediary .h files.

Carmina16:
I tried to use solely `gettext` but couldn't make it work properly on my system. Also, `msgfmt` contains a bug that prevents .desktop to be localized completely.

sjohannes:
I see. Thanks; I'll try the patch after 4.0 is released, though I might try the xgettext-only route first.

luzip665:
Finally done in #826 
Thanks to @Carmina16

PR #563, opened 2018-10-07
Show GTK+ theme name in About dialog

sjohannes:
For when we get theme-specific bugs.


virtuald:
Seems like it should be logged when the GTK version is logged also?

sjohannes:
How about this one? We call `version.register` to log and store the initial value, then just before showing the about dialog we update it.

virtuald:
I don't really see why you'd need to cache it. What if the theme changes at runtime?

virtuald:
Oh I'm an idiot, I misread it earlier.

PR #570, opened 2018-11-12
Fix Files panel failing to browse to C:\

sjohannes:
Currently the Files panel chokes on `C:\` due to the `*.sys` files. In this loop:

    infos = directory.enumerate_children(...)
    for info in infos:
        ...

some of the internal `next()` calls throw errors. We can't simply put the whole thing in a try-except because we just want to skip the problematic files, not break out of the loop entirely.

So this patch wraps the iteration, putting each `next()` call in a try-except. This assumes that the iterator is still in a valid state, and that it still raises `StopIteration` after going through each file (this works well in my testing).

### DIFF
--- a/xlgui/panel/files.py
+++ b/xlgui/panel/files.py
@@ -43,6 +43,26 @@ from xlgui.widgets.common import DragTreeView
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


sjohannes:
Merged manually.

Issue #575, opened 2018-12-29
Better Album Artist support


rokm:
This year I finally got around to reorganizing and re-tagging my music library, where the following two were made (among the others):
* making use of the "Album Artist" tag (in addition to the "Artist")
* move of featured author attribution from track's Title to the Artist, as per [the following recommendation](https://wiki.musicbrainz.org/Style/Artist_Credits#Featured_artists)

Unfortunately, this sort of organization seems to have a somewhat lacking support in Exaile. In particular, because Artist tag now also contains feautured artist, the "Artist" view in Collection is full of entries in form of:
- Some Artist feat. Some Other Artist A
- Some Artist feat. Some Other Artist B
- Some Artist feat. Some Other Artist C
- etc.

This, on one hand, makes it harder to browse the "Artist" view in Collection, and on the other hand also makes it impossible to select a whole Album at once (from the Artist > Album view).

In fact, with the above organization, pretty much all sorting and grouping that is being based on Artist tag should likely be done based on the Album artist instead. On the other hand, the original Artist-based view might still be useful, for example when trying to find a specific combination of artist and featured artist.

Therefore, I propose that for each Artist-based view in the Collection, we add a corresponding Album-Artist-based view: Album Artist; Genre > Album Artist; Date > Album Artist; Album Artist > (Date - Album), and an additional Album Artist column to the playlist. This should enable better support for album-artist-based library, while preserving functionality that we have now.

I will open a pull request to implement this functionality.

I would appreciate if someone familiar with that part of code and functionality takes a look at it, because there already seems to be some support for album artist and file-location-based compilation detection, but this never seemed to work for me with the old Artist-based library (i.e., movie soundtracks with different artists were always scattered among individual artist entries). With the Album-artist-based library, compilations are explicitly tagged with Album Artist being set to "Various Artists", so there is no need to really detect them (if album artist is available)...

sbrubes:
I propose you to test the bcollection plugin (https://github.com/exaile/exaile/pull/593).

This have a syntax that allows user to customize tree view patterns the way he wants.

Unfortunatelly, it request that you have a Beets database, or, at least a SQLite database with a table called `items` and the fields described [here](https://github.com/exaile/exaile/pull/593/files#diff-83d39449c1252bbe667aa516b9c42793R35). Anything else is defined by what you put on pattern.

Maybe someday we could add some of those features to Standard Exaile's Collection.

rokm:
Closing this issue as the added support from #576 more or less covers my use case.

PR #576, opened 2018-12-29
Album artist support

rokm:
This PR implements improved support for Album Artist based library, as outlined in #575.

Changes:
* addition of "Album Artist" views to Collection pane
* always display Album Artist field in Track properties dialog
* added "Album artist" column to playlist
* DAAP client: retrieve album artist information from the server


virtuald:
I'm inclined to merge this even though I don't understand the code. It would be nice if some of this track stuff had unit tests, but there aren't right now. @sjohannes thoughts?

sjohannes:
All `__compilation`-related code makes my head spin, but at a glance this looks alright to me.

If artist != albumartist, can we search for either one?

Issue #551, opened 2018-09-25
Cannot set rating by clicking on the 'Rating' column


phalgunlolur:
Hi,

I would like to directly be able to rate my songs with a rating. Currently if I click on the stars nothing happens. I have to right-click it and select a rating.

Also, it would be helpful, if half star ratings are introduced as well.

sjohannes:
Hm, the Rating colum on the tracklist is meant to be clickable, so this is actually a bug.

PR #577, opened 2019-01-03
Playlist rating fix

rokm:
This PR should address #551, which is caused by click_area being effectively (0, 0, 0, 0) due to its members not being initialized from the values passed the constructor. The second commit allows user to set the rating even if clicked above or below the stars in the corresponding cell.


virtuald:
Fantastic, thank you.

Issue #568, opened 2018-11-12
Ctrl+arrow on entry boxes triggers seek instead of word navigation


sjohannes:
1. Play a track.
2. Press Ctrl+F, type "hello world".
3. Press Ctrl+Left.

Expected: Cursor moves to left of "w".

Actual: Playback seeked backward.

PR #578, opened 2019-01-06
Temporarily disable main window's accelerators in search entry widgets

rokm:
This PR is an attempt to address #568. The issue is that the accelerators installed to main window are triggered even when editing a text entry. As far as I am aware, the only way around this is to temporarily remove the accelerators when text entry gains keyboard focus, and reinstall them once the focus is lost.

To do so, the first commit extends AcceleratorManager with two new methods that allow enabling/disabling all registered accelerators (in addition to keeping track of registered/unregistered providers/accelerators).

The second commit adds support for disabling and re-enabling accelerators when guiutil.SearchEntry gains and loses keyboard focus, respectively. This allows skipping between the words with ctrl+arrow not only in Playlist's Search entry (#568), but also in all other search entries that use this utility class (for example, the search entry in Collection panel).

### DIFF
--- a/xlgui/guiutil.py
+++ b/xlgui/guiutil.py
@@ -37,6 +37,9 @@ from gi.repository import Pango
 
 from xl import settings, xdg
 
+# Required for xlgui.get_controller()
+import xlgui
+
 # Import from external namespace
 from xl.externals.gi_composites import GtkTemplate as _GtkTemplate
 
@@ -309,6 +312,17 @@ class SearchEntry(object):
         entry.connect('changed', self.on_entry_changed)
         entry.connect('icon-press', self.on_entry_icon_press)
         entry.connect('activate', self.on_entry_activated)
+
+        # Disable global accelerators while editing, re-enable when
+        # done
+        entry.connect(
+            'focus-in-event',
+            lambda w, e: xlgui.get_controller().main.accel_manager.disable_accelerators(),
+        )
+        entry.connect(
+            'focus-out-event',
+            lambda w, e: xlgui.get_controller().main.accel_manager.enable_accelerators(),
+        )
 
     def on_entry_changed(self, entry):
         """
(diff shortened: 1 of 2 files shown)


xlgui/guiutil.py
```diff
@@ -310,6 +313,17 @@ def __init__(self, entry=None, timeout=500):
         entry.connect('icon-press', self.on_entry_icon_press)
         entry.connect('activate', self.on_entry_activated)
 
+        # Disable global accelerators while editing, re-enable when
+        # done
+        entry.connect(
+            'focus-in-event',
+            lambda w, e: xlgui.get_controller().main.accel_manager.disable_accelerators(),
```
virtuald: This seems fine to me. The only question I have is whether there's a guarantee that the `focus-out-event` will always be emitted? Though, now that I've typed it, that seems silly that it wouldn't be guaranteed.

virtuald:
I tried this, works great.

I don't use keyboard shortcuts all that much, but I opened our useful shortcuts dialog and I note that I would expect accelerators that don't conflict with the normal operation of the entry box to continue working. I wonder if there's a way we can just disable the conflicting keys, via a lookup or something similar?

rokm:
Hmm, my argument would go the opposite way: that while one is editing a text entry, there is no reason for any accelerators, conflicting or not, to be enabled. Because the current mode of operation becomes "entering text" instead of controlling other parts of application (either intentionally or accidentally).

But if desired, I guess we could devise a support for disabling specific accelerators, like the ones that are expected to conflict with text entry operation.

sjohannes:
I think some accels still make sense even during editing, for example for quitting and adjusting volume. I'm not too familiar with how GTK accellerators work; is there a way to put our accels at lower priority than the toolkit defaults?

rokm:
I am not aware of any priority support in Gtk accelerators (aside from locking them from being changed during application's runtime, apparently). Nor does there seem to be any built-in support for temporarily disabling them.

virtuald:
Seems good enough to me.

Issue #64, opened 2015-03-16
Regression: errors loading Smart Playlists that filter based on a playlist


virtuald:
Errors occur similar to: `'NoneType' object has no attribute 'type'`

PR #581, opened 2019-01-11
Playlist: _search_playlist: fixed detection of timestamp-based fields

### LINKED ISSUE #19: Smart playlists with criteria involving timestamps do not work
The search string for criteria like "Last time played" isn't constructed correctly. Also, playlists like "all tracks not played in the last X weeks/days/..." don't contain any songs for which the `__last_played` field isn't set yet, although that is probably the expected behaviour. 

[Here is a patch](https://gist.githubusercontent.com/slosd/22b7888d47b4d533d78a/raw/45c5d69d5cf773976e9d77f5eae7fa71a2ad73dd/0001-Fix-smart-playlists-involving-timestamps.patch) with changes I made to fix it for me personally. Possibly that's not how it should be properly fixed. In particular, I'm not sure if the localization I removed is actually desired.

rokm:
Due to incorrect comparison of "timestamp" string against _TD object
returned by tag_data.get(field), we failed to detect timestamp-based
fields, which in turn broke the search/filtering. The comparison
should be done against "type" member of the _TD object.

Should fix #19 (which was reopened for 4.0.0).

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -1952,6 +1952,9 @@ class SmartPlaylist(object):
                 params += [param]
                 continue
             (field, op, value) = param
+            fieldtype = tag_data.get(field)
+            if fieldtype is not None:
+                fieldtype = fieldtype.type
 
             s = ""
 
@@ -1975,7 +1978,7 @@ class SmartPlaylist(object):
                 else:
                     matchers.append(trax.TracksNotInList(pl))
                 continue
-            elif tag_data.get(field) == 'timestamp':
+            elif fieldtype == 'timestamp':
                 duration, unit = value
                 delta = durations[unit](duration)
                 point = datetime.now() - delta


virtuald:
This was changed to fix an older bug... #64 . I think you need to ensure that the result of `.get` is not None. See [this commit](https://github.com/exaile/exaile/commit/aef3340d89002ffc15f05bbc8bdb465f319fee67) which seems to have addressed that issue incorrectly.

rokm:
> This was changed to fix an older bug... #64 . I think you need to ensure that the result of `.get` is not None. See [this commit](https://github.com/exaile/exaile/commit/aef3340d89002ffc15f05bbc8bdb465f319fee67) which seems to have addressed that issue incorrectly.

Ah, I see. In that case, I think it would be best to bring back the ```fieldtype``` variable. I've pushed the modified commit.

sjohannes:
Confirmed that this fixes the issue.

PR #582, opened 2019-01-11
widgets: SpinButtonAndComboField: fix state restore in non-C locale

rokm:
When running localized version of Exaile, the combo boxes in
SpinDateField (e.g., for "Last played") keep resetting to seconds
every time the dialog is re-run.

The issue is that set_state() translates the input unit string
before trying to look it up in internal list of (non-translated)
strings.


virtuald:
Thanks!

Issue #580, opened 2019-01-07
Playlist column widths forgotten on restart


4ndyD:
### Steps to Reproduce (for bugs)
Quit and re-open Exaile.
This was observed after an upgrade to LM18.3 (based on U16.04) that broke some GStreamer codec plugin dependency for mp3 in the repository-provided version 3.3.2, causing [this bug](https://github.com/exaile/exaile/issues/35) despite all qualities of codec plugins being installed, resulting in me testing the latest version.

### Expected Behavior
The playlist columns should be at the same width as they were when the application last ran.

### Current Behavior
When Exaile starts after quitting, all columns are at their minimum width (nothing in the column can be read), except for the right-most column, which takes up the rest of the screen width.

Also, sometimes when dragging out the collapsed columns to appropriate widths, as soon as the edge of the last (left-most) column "\#" for track number is clicked on, the entire application crashes. This last part has been unreliable - e.g. it happened on the first 2 of 3 times this morning, but I only observed it so far on 4.0.0beta2, currently available as .deb installer via PPA.

There is also slightly different starting visual behaviour - in 4.0.0beta2, all the columns begin collapsed onto the same pixel-width, whereas in 4.0.0rc3, they begin with a few pixels regularly spaced between each column edge.

### Environment
Linux Mint 18.3 XFCE 32bit edition
Exaile 4.0.0rc3 from source
GStreamer: 1.8.3
GTK+: 3.18.9
GTK+ theme: Mint-X
Locale: en_GB UTF-8
Mutagen: 1.41.1
PyGObject: 3.20.0
Python: 2.7.12

Peculiarity - I also have Python 3.5 installed, but it seems to be defaulting to 2.7 when running.

I looked at another system already running LM18.3 64bit, and Exaile was not shown in the repository at all. Executing from the github release would not start, probably lacking dependencies.

sjohannes:
Hm, I can't reproduce the issue on my system. What I did to test this:

1. Resize the Artist colum to be very wide.
2. Restart Exaile.

And the column is still the same width. I also don't encounter the other weirdness that you get.

Could you run Exaile from a terminal and see if there are any error messages displayed?

4ndyD:
Console output, two errors displayed:
```
INFO    : Loading Exaile 4.0.0beta2...
INFO    : Using Python 2.7.12
INFO    : Using PyGObject 3.20.0
INFO    : Loading settings...
INFO    : Using Locale en_GB UTF-8
[...]

PR #585, opened 2019-01-16
playlist_columns: Fix columns getting 0 width on older GTK+

sjohannes:
On some older GTK+ versions (e.g. on Linux Mint 18.3 with GTK+ 3.18.9), on startup GTK+ emits the `notify::width` signal on every column while the column width is 0. (Or maybe this is timing-related instead of version-related? I'm not entirely sure.)

This causes us to save the value 0 to our settings, and then that gets applied, causing [#580](https://github.com/exaile/exaile/issues/580).

This patch ignores any `notify::width` events with width==0. This assumes that it doesn't make sense for the user to shrink columns until they're invisible.


sjohannes:
Withdrawn; will post a better one based on discussion in #580.

PR #587, opened 2019-01-27
Playlist columns: filter out activation events for non-active size items

### LINKED ISSUE #586: Right-clicking on column header disables column autosize
1. Right click on column header, click Autosize.
2. Right click on column header, then dismiss the popup (e.g. with Escape or by clicking elsewhere).
3. Right click on column header and look at the bottom two entries.

Expected:

☐ Resizable  
☒ Autosize

Actual:

☒ Resizable  
☐ Autosize

---

Tested on GTK+ 3.24.4 on MSYS.

#201 actually touched on this, but the main focus of that bug was on the wild sizing behaviour which doesn't occur anymore.

rokm:
In certain scenarios (e.g., during menu (re)population), we may
receive an "activate" event for a radio menu item that is not
active anymore. Therefore, on_sizing_item_activate() needs to
verify the menu_item's actual state before taking action.

Fixes #586, where such spurious activation is emitted for the
first radio item ("Resizable"), and thus we always end up marking
it as selected.

### DIFF
--- a/xlgui/widgets/playlist_columns.py
+++ b/xlgui/widgets/playlist_columns.py
@@ -675,6 +675,11 @@ def __register_playlist_columns_menuitems():
         """
             Updates column sizing setting
         """
+
+        # Ignore the activation if the menu item is not actually active
+        if not menu_item.get_active():
+            return
+
         if name == 'resizable':
             settings.set_option('gui/resizable_cols', True)
         elif name == 'autosize':


rokm:
Perhaps it does from semantic viewpoint, but its behaviour is identical to the one of ```activate```, i.e., it gets emitted both for the the item that lost active status and the one that gained it (because they were both, in fact, toggled). So we would get the event for un-selected item during menu's construction as well.

PR #594, opened 2019-02-06
Editable comment playlist column

### LINKED ISSUE #579: Add comments to a song on the main player with a single/double click
It would be beneficial for me, if I can double click on the comments column for each song and add a comment.

Currently, I have to right-click each song, click on track properties, go to tag properties, scroll to the comments section and then add my comment. Which is okay for one or two songs, but if I'm adding multiple comments, it becomes a pain.

rokm:
This PR implements editable comment playlist column (enhancement #579).

The first commit makes ```CommentTagFormatter``` use ```get_tag_raw()``` instead of ```get_tag_disk()``` to retrieve the comment. Without this change, the comment column always displays the comment stored in the music file, and fails to reflect the changes made to the comment via Exaile's Track properties dialog.

The second commit addresses interference with selection of an editable column, which turns out to be caused by code-path that aims to preserve multiple-item selection for DnD. Since the latter is relevant only for multiple-item selection, it should be activated only when multiple items are actually selected.

The third commit marks the comment column as editable and implements storage of the edited value.

The fourth commit expands the preceding one by adding support for newlines; when the column is edited, we now display newlines '\n' instead of stripping them, which allows both addition of new lines directly in the column's edit entry, and prevents the loss of existing newlines in the comment.

Closes #579.

### DIFF
--- a/xl/formatter.py
+++ b/xl/formatter.py
@@ -850,7 +850,7 @@ class CommentTagFormatter(TagFormatter):
             :returns: the formatted value
             :rtype: string
         """
-        value = track.get_tag_disk(self.name)
+        value = track.get_tag_raw(self.name)
 
         if not value:
             return ''
--- a/xlgui/widgets/playlist_columns.py
+++ b/xlgui/widgets/playlist_columns.py
@@ -543,6 +543,18 @@ class CommentColumn(Column):
     autoexpand = True
     # Remove the newlines to fit into the vertical space of rows
     formatter = TrackFormatter('${comment:newlines=strip}')
+    cellproperties = {'editable': True}
+
+    def __init__(self, *args):
+        Column.__init__(self, *args)
+        self.cellrenderer.connect('edited', self.on_edited)
+
+    def on_edited(self, cellrenderer, path, new_text):
+        model = self.get_tree_view().get_model()
+        iter = model.get_iter(path)
+        track = model.get_value(iter, 0)
+
+        track.set_tag_raw("comment", new_text)
 
 
 providers.register('playlist-columns', CommentColumn)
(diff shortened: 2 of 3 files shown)


xl/formatter.py
```diff
@@ -850,7 +850,7 @@ def format(self, track, parameters):
             :returns: the formatted value
             :rtype: string
         """
-        value = track.get_tag_disk(self.name)
+        value = track.get_tag_raw(self.name)
```
virtuald: I don't believe comments can use raw tag, otherwise things break... I don't quite remember why this distinction exists, but it does exist. I believe it's because comments aren't currently stored in the database.

virtuald: Hm, actually, that might just be covers. Comment isn't marked as use_disk in tags.py.

virtuald:
I tried it out, seems to work great, except that it doesn't save the tag to disk, and all other edits in Exaile save the tag to disk.

I have some other thoughts, but I'll post them on the original request. Depending on how that goes, I think it could be useful to make the editable column thing a mixin that can be used by other columns.

xlgui/widgets/playlist_columns.py
```diff
@@ -539,6 +539,39 @@ class CommentColumn(Column):
...
+        # Undo newline escaping
+        new_text = new_text.decode('unicode-escape')
+
+        # Set comment
+        model = self.get_tree_view().get_model()
+        iter = model.get_iter(path)
+        track = model.get_value(iter, 0)
+
+        track.set_tag_raw("comment", new_text)
+        event.log_event('comment_changed', self, new_text)
```
virtuald: Why is this extra event added? When tags are changed an event already is emitted, no need to add a special event for this.

... sorry for not using the review feature and doing a bunch of comments instead.

rokm: Ah, I probably added that event logging for debug purposes while developing the patchset (although as you said, it is redundant), and forgot to remove it. Will push a revised patchset without it.

Issue #588, opened 2019-02-02
Running exaile.exe opens git.exe window


scootergrisen:
### Steps to Reproduce (for bugs)
Run exaile.exe

### Expected Behavior
Clean start up of the Exaile window without uncessasry stuff like command prompt windows flashing and icons changing.

### Current Behavior
A quick flash is seen when i run exaile.exe because a window with the title "C:\Program Files\Git\cmd\git.exe" opens and closes very quickly.
The window is empty. No commands can be seen.
This window have an generic icon on the process bar.
This is very short only like 1-2 frames.

Then an icon appear on the process bar with the same icon as used by the exaile.exe file (not the green play logo) and the Exaile splash image is visible for about 1-2 seconds.

Then the icon is closed and at the same time a new icon (the green play logo) opens with the Exaile window.

### Environment
* Operating System and version:
Windows 7 6.1.7601

* Exaile Version: 
4.0.0rc3

GStreamer: 1.14.2
GTK+: 3.24.0
Locale: Danish_Denmark 1252
Mutagen: 1.40.0
PyGObject: 3.30.0
Python: 2.7.15

genodeftest:
On the releases (not checked out from git) we could just skip the git code in [xl/version.py](https://github.com/exaile/exaile/blob/fc9648106456dabfb88c2e7ca49b6f23d1dbe184/xl/version.py#L39). Any objections?

sjohannes:
> On the releases (not checked out from git) we could just skip the git code in xl/version.py. Any objections?

I think it's already meant to only run git if `xdg.local_hack` is true. So in this case it seems that the check is wrong.

sjohannes:
Looks like `xdg.local_hack` is the wrong thing to use here. It checks for the presence of the `data` dir in the install location, but this directory is present on Windows and OS X installations, which don't use FHS-style layout.

Maybe we should just check for the `.git` directory.

rokm:
If release archives are generated using ```git archive```, it is also possible to use ```.gitattributes```and ```export-subst``` mechanism to hard-code (at export time) an indicator variable in the exported code (which can either store the revision identifier for the exported code, or just serve as an indicator that the code is exported archive).

PR #600, opened 2019-03-12
Get current git revision only if .git directory exists

X0rg:
Related to #588.

> Maybe we should just check for the .git directory.

So that's what I did.


sjohannes:
Merged, thanks.

Issue #599, opened 2019-03-12
Use {}/format instead of % for ngettext strings


sjohannes:
In some languages it can make sense to translate some numbered strings without including the number itself.

For example, currently our Arabic localization does this for ["%d second" / "%d seconds"](https://hosted.weblate.org/translate/exaile/master/ar/?checksum=5f827310961c6764) when `%d` is 0, 1, or 2. Luckily this particular string is unused, because otherwise it would cause an error:

```pycon
>>> 'abc' % 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
```

These strings should be changed from using percent-based formatting to {}-based formatting (`str.format`) because the latter ignores extraneous args:

```pycon
>>> 'abc'.format(1)
'abc'
```

As part of this, I'd want the existing translations to be updated as well, so that we don't invalidate them all.

dkstm95:
Hi :)
Could you tell me one example file that include this issue ?

sjohannes:
The `%d seconds` example comes from [`xl/formatter.py`](https://github.com/exaile/exaile/blob/50f9337ff3749929d0d69a81ce82a85a6507f42c/xl/formatter.py#L601) and the code looks like this:

    text += ngettext('%d second', '%d seconds', span.seconds) % span.seconds

You can find all affected strings by searching for `ngettext` in the source code. There are only a few of them (4 lines in `xl/formatter.py` and 1 line in `xlgui/preferences/plugin.py`), but the bigger task is updating our existing translations (`po/*.po`). For example, the entry in `ar.po` should change from

```
#: ../xl/formatter.py:601
#, python-format
msgid "%d second"
msgid_plural "%d seconds"
msgstr[0] "%d seconde"
msgstr[1] "%d secondes"
```

to something like

```
#: ../xl/formatter.py:601
msgid "{} second"
msgid_plural "{} seconds"
msgstr[0] "{} seconde"
msgstr[1] "{} secondes"
```

dkstm95:
I am now changing all %d to {}, is this right?
(except for %dd or % dk etc)

sjohannes:
No, this task is just for numbers that are translated with `ngettext`. The ones translated with `gettext` are fine and don't need to be changed.

dkstm95:
Oh, what I mean is %d in .po files.

sjohannes:
Only for the strings that are also changed in the `.py` files.

PR #602, opened 2019-03-15
Change format(%d -> {})

dkstm95:


sjohannes:
Thanks for the merge request, but as I mentioned in https://github.com/exaile/exaile/issues/599#issuecomment-472685061, you should only modify these strings in `*.po`:

https://github.com/exaile/exaile/blob/e5462e3160acbb7f5b7735da353e16bccb414df1/po/fr.po#L31-L57

PR #605, opened 2019-03-21
Change volume up accellerator from Ctrl++ to Ctrl+=.

sjohannes:
Ctrl+= makes more sense than Ctrl++ on QWERTY, but are there layouts where it's bad? If this is an issue on other layouts we can always allow both (like Firefox does).

### DIFF
--- a/xlgui/main.py
+++ b/xlgui/main.py
@@ -185,7 +185,7 @@ class MainWindow(GObject.GObject):
                 lambda *_e: self._on_seek_key(False),
             ),
             (
-                '<Primary>plus',
+                '<Primary>equal',
                 _('Increase the volume'),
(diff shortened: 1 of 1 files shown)


genodeftest:
On German Qwertz, Ctrl++ is quite easy, Ctrl+= is Ctrl+Shift+0, which is quite hard. If we're starting to optimize accelerators for keyboard layout, we should probably make them dependent on the locale (as the keyboard layout is probably not available to the application). I doubt we want to add this amount of complexity though.
I suggest either allowing both, Ctrl++ and Ctrl+= at the same time.

sjohannes:
Let's go with allowing both for now. I don't know what the standard practice is for this sort of thing, whether apps usually change accels based on locale.

Issue #610, opened 2019-04-05
plugin massstorage and Device Manager is broken: rework or remove


genodeftest:
### Steps to Reproduce (for bugs)
1. Start Exaile
2. enable plugin massstorage
3. attach any multimedia device
4. Open "Tools" → "Device Manager" 

### Expected Behavior
The code should react to the device being attached. Device manager should list the device if appropriate

### Current Behavior
Nothing happens.

### Reason
The massstorage plugin uses hal which is not available on most modern desktop systems. It has been replaced by UDisks(1) which has been replaced by Udisks2 years ago.

### Possible Solution
Two options:
* Port massstorage to UDisks2 and make "Device Manager" work again
* Drop massstorage plugin and "Device Manager" feature

If the massstorage plugin is not available (on Windows, because we have no UDisks2 on Windows), what should happen then? Disable the "Device Manager" feature? See also #46. Contrary to #46, the iPod plugin has also been removed and the CD plugin provides its own panel so there is no use for DeviceManager without massstorage plugin.

### Environment
* Operating System and version: Fedora 29
  * kernel-5.0.5-200.fc29.x86_64
  * gstreamer1-1.14.4-1.fc29.x86_64
  * gtk3-3.24.1-2.fc29.x86_64
  * glib2-2.58.3-1.fc29.x86_64
  * python2-2.7.15-11.fc29.x86_64
* Exaile Version: 
  * 4.0.0rc3+ec0b515

PR #611, opened 2019-04-05
Drop hal and UDisks(1) support

genodeftest:
HAL has been deprecated for years and has been removed from most Linux distributions . Even its successor, UDisks version 1 has been superseded by UDisks2 about 4 years ago.

FreeBSD has its own implementation of UDisks2: https://svnweb.freebsd.org/ports/head/sysutils/bsdisks/
@rokm: Do I remember correctly that you are using FreeBSD? If yes, can you please check whether this works for you too?

The massstorage plugin is still broken, see #610 for that.

### DIFF
--- a/plugins/cd/PLUGININFO
+++ b/plugins/cd/PLUGININFO
@@ -1,7 +1,7 @@
 Version='4.0.0'
 Authors=['Aren Olson <reacocard@gmail.com>']
 Name=_('CD Playback')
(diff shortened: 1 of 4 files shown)


rokm:
I'm using Fedora, not FreeBSD. But I gave it a try in a VM, and can confirm that Exaile under FreeBSD 12 also enters the UDisks2 codepath (I could not test it any further, though, because my test installation seems to have problems reading/accessing Audio CDs in general...).

So Exaile on FreeBSD should not be affected by this PR, and should continue working as it did before the change.

PR #614, opened 2019-04-20
Editable column updates

virtuald:
* Extract to EditableColumn
* Write on edit
* Update edited value immediately instead of waiting for the cache to catch up
  * Makes a difference on large playlists
* Make many columns editable

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -1407,6 +1407,12 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
 
             return True  # Prevent 'row-activated'
 
+        # Columns with editable rows will take row-activated and start editing
+        # but we want to start playback, so simulate it instead
+        elif event.keyval == Gdk.KEY_Return:
+            self.on_row_activated()
+            return True  # Prevent 'row-activated'
+
     def on_header_key_press_event(self, widget, event):
         if event.keyval == Gdk.KEY_Menu:
             # Open context menu for selecting visible columns
--- a/xlgui/widgets/playlist_columns.py
+++ b/xlgui/widgets/playlist_columns.py
@@ -180,6 +180,13 @@ class EditableColumn(Column):
         # Undo newline escaping
         new_text = new_text.decode('unicode-escape')
 
+        validate = getattr(self, 'validate', None)
+        if validate:
+            try:
+                new_text = validate(new_text)
+            except ValueError:
+                return
+
         # Update the track
         model = self.get_tree_view().get_model()
         iter = model.get_iter(path)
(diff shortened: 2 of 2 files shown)


rokm:
I gave this a quick test, and overall it seems to work quite nicely.

One thing I noted is that, for Comment column, the preservation of newlines was lost in the transition. 

The Comment column uses ```TrackFormatter('${comment:newlines=strip}')``` as formatter, which strips out the newlines. And the new generic ```EditableColumn``` uses column's default formatter to populate the editable text entry. So if a multi-line comment is created (e.g., via Track Properties dialog), and is then directly edited via editable column, the newlines will be lost.

In old editable Comment column this was solved by explicitly using a different formatter for filling editable text entry. Perhaps a column should have an additional formatter specified for editing purposes?

virtuald:
Good catch, fixed it!

phalgunlolur:
This is great, thanks. It made me realize something else though. Now, I can no longer double-click on the song to play it. Or even hit enter. It just takes it to edit mode.

I have to click on the tiny portion just before the track number column to play the song. I wonder if something can be done about it

virtuald:
I suspect that other rendering engines will only activate the 'editing' functionality if the click occurs on the text itself, and not the empty area between the cells? Unfortunately, it seems that GTK doesn't have this option available as far as I can tell.

virtuald:
It is possible to create a custom cell renderer that would have that behavior. It's not clear to me whether the effort is worth it -- but I admit that not being able to double click things now is pretty annoying.

virtuald:
Went digging in the GTK source code, it looks like it's extensible enough to allow this sort of modification by overriding the start-editing function. Conveniently, there's also a function that seems to tell you where the text will be rendered to ([Gtk.CellRenderer.get_aligned_area](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/CellRenderer.html#Gtk.CellRenderer.get_aligned_area)), so this was easier than I thought it might be.

Try it out and let me know how it works for you.

genodeftest:
As discussed in #626: Can you please rebase your changes to the latest git master? Then it will pass the Fedora build too.

rokm:
Works for me. There's one remaining (potential) regression with respect to original behavior; Enter key now triggers editing instead of starting playback, which may be undesirable.

I suppose an easy way to work around this is to modify ```on_key_press_event``` in ```playlist.py``` to handle the Enter key and trigger the playback while preventing further event processing, similarly to what we already do for the space key (except that ```_play_track_at``` needs to be called with both ```on_activated``` and ```restart_if_playing``` arguments set to ```True``` to match the old behavior).

virtuald:
Fixed that too.

PR #619, opened 2019-05-02
Travis: Update to match docker image versions (2nd try)

genodeftest:
Retry because I merged #615 too fast.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,11 @@ services:
 language: python
 
 env:
-  - IMGTAG=debian8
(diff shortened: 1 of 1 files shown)


genodeftest:
the travis job for exaile-testimg seems to have a hickup, thus is missing both the Fedora 27 and Fedora 29 builds. I will try to get new gst-docker-python with the latest Fedora 30 instead.

As Travis does not have Python 3.7 yet, I'll skip that too, to be updated later.

PR #621, opened 2019-05-04
Plugin musicbrainzcovers: Small improvements

genodeftest:
Changes:
* Use exaile's API to generate user agent string
* Switch to HTTPS for coverartarchive.org

### DIFF
--- a/plugins/musicbrainzcovers/__init__.py
+++ b/plugins/musicbrainzcovers/__init__.py
@@ -52,7 +52,7 @@ class MusicBrainzCoverSearch(covers.CoverSearchMethod):
 
     name = 'musicbrainz'
     title = 'MusicBrainz'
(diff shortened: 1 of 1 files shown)


plugins/musicbrainzcovers/__init__.py
```diff
@@ -16,15 +16,19 @@
...
 import musicbrainzngs
 
 logger = logging.getLogger(__name__)
 
-musicbrainzngs.set_useragent(
-    'Exaile_MusicBrainz_Covers', '1.0.0', 'https://exaile.org/'
-)
+
+def __init_musicbrainzngs():
+    version = main.exaile().get_user_agent_for_musicbrainz()
```
sjohannes: I don't think this exists yet? Also, I think the function should stay inside the plugin.

genodeftest: You're right about it not existing yet. Looks like I am confused ;)

PR #624, opened 2019-05-04
Makefile: Also clean docs if possible

genodeftest:


genodeftest:
This ought to be simple, but it looks like it isn't. The sphinx module on Fedora 28 is very outdated, this may be the cause for the failure. Will not merge because it will break all the other builds.

PR #626, opened 2019-05-05
Travis: Update Fedora versions to test against

genodeftest:


genodeftest:
I can reproduce the test fail on my Fedora 29 installation.

genodeftest:
Note: Both Fedora 29 and Fedora 30 are failing the test. This is caused by a bug in gdbm, which has been [reported to debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910911) and [fixed upstream](https://puszcza.gnu.org.ua/forum/forum.php?forum_id=1194), but the fix has not yet been released to Fedora. I've [reported a bug](https://bugzilla.redhat.com/show_bug.cgi?id=1706639) hoping that it gets fixed soon.

The library `gdbm` is being pulled in as dependency of `python2-libs`, which is being installed during generation of the [exaile/gst-python-docker](https://github.com/exaile/gst-python-docker) containers. 
Once Fedora's package maintainer updates the version and thus fixes the bug, we need to rebuild the docker containers of [exaile/gst-python-docker](https://github.com/exaile/gst-python-docker) and of [exaile/exaile-testimg](https://github.com/exaile/exaile-testimg) before triggering this build again.

genodeftest:
For reference, this is the test's output:
```
_____________________________ test_migration[gdbm] _____________________________
data = ({'_dbversion': 2.0, '_key': 10, '_serial_libraries': [{'location': 'file:///test', 'monitored': False, 'realtime': Fa...n_interval': 0, ...}], 'name': 'Collection', ...}, '/tmp/pytest-of-exaile/pytest-0/test_migration_gdbm_0/gdbm', 'gdbm')
    def test_migration(data):
        truth, loc, dbtype = data
    
        print(os.listdir(loc))
    
        try:
>           db = open_shelf(join(loc, 'music.db'))
tests/xl/trax/test_migration.py:55: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
xl/common.py:431: in open_shelf
    migrate(path)
xl/migrations/database/to_bsddb.py:41: in migrate
    old_shelf = shelve.open(path, 'r', protocol=common.PICKLE_PROTOCOL)
/usr/lib64/python2.7/shelve.py:243: in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
/usr/lib64/python2.7/shelve.py:227: in __init__
    Shelf.__init__(self, anydbm.open(filename, flag), protocol, writeback)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
file = '/tmp/pytest-of-exaile/pytest-0/test_migration_gdbm_0/gdbm/music.db'
flag = 'r', mode = 438
    def open(file, flag='r', mode=0666):
        """Open or create database at path given by *file*.
    
        Optional argument *flag* can be 'r' (default) for read-only access, 'w'
        for read-write access of an existing database, 'c' for read-write access
        to a new or existing database, and 'n' for read-write access to a new
        database.
    
        Note: 'r' and 'w' fail if the database doesn't exist; 'c' creates it
        only if it doesn't exist; and 'n' always creates a new database.
        """
    
        # guess the type of an existing database
        from whichdb import whichdb
        result=whichdb(file)
        if result is None:
            # db doesn't exist
            if 'c' in flag or 'n' in flag:
                # file doesn't exist and the new
                # flag was used so use default type
                mod = _defaultmod
            else:
                raise error, "need 'c' or 'n' flag to open new db"
        elif result == "":
            # db type cannot be determined
            raise error, "db type could not be determined"
        else:
            mod = __import__(result)
>       return mod.open(file, flag, mode)
E       error: Malformed database file header
/usr/lib64/python2.7/anydbm.py:85: error
```

virtuald:
Did you delete the fedora27 image? It doesn't seem to be present on docker hub anymore.

genodeftest:
> Did you delete the fedora27 image? It doesn't seem to be present on docker hub anymore.

Yes, I did. Was there anything wrong with that? I thought because Fedora 27 has been out of support for 5 months we would not need it any more. I'd also replace Fedora 28 asap, but that does not work yet due to the bug described a few comments up.

genodeftest:
Oh, I see. It looks like I broke your #614. I'm sorry for that. Although I don't really understand why: `.travis.yml` on git master has it removed (see #619). Somehow GitHub/Travis picks up the old travis config (also with debian 8 and the old ubuntu image) instead of the current one. I'll try to rebase your commits to the latest git master, that should work around the issue.

genodeftest:
Looks like I'm not permitted to do so.

virtuald:
Well, if anyone ever wanted to test with the old images for some reason, I would have liked to keep them around. But too late. :-/

genodeftest:
> Well, if anyone ever wanted to test with the old images for some reason, I would have liked to keep them around. But too late. :-/

I'm sorry, I should have asked before doing so…

genodeftest:
This pull request is currently blocked by the following Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1706639

genodeftest:
I dropped the Fedora 30 image (because the bug will probably not be fixed there anyways) and updated directly to Fedora 31.

PR #628, opened 2019-05-11
Explicit python2 in all shebangs

X0rg:
I'm packaging Exaile 4.0.0.rc4 for Arch Linux, and I see `namcap` complains about the built package:
```
$ namcap exaile-4.0.0.rc4-1-any.pkg.tar.xz
exaile E: Dependency python detected and not included (programs ['python'] needed in scripts ['usr/share/exaile/plugins/lastfmlove/cellrenderertoggleimage.py'])
```
It is because `namcap` thinks Python 3 is required (due to Arch's default), but it is not the case.

I prefer to set an explicit Python version to avoid this.

### DIFF
--- a/plugins/lastfmlove/cellrenderertoggleimage.py
+++ b/plugins/lastfmlove/cellrenderertoggleimage.py
@@ -1,4 +1,3 @@
-#! /usr/bin/env python
 # -*- coding: utf-8 -*-
 
 from gi.repository import GdkPixbuf, GObject, Gtk, Gdk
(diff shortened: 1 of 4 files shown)


sjohannes:
I think just remove the one in `cellrenderertoggleimage.py`, since it's not executable.

X0rg:
What about the others under `tools`?

sjohannes:
The IPython ones look like they are executable, but the one in `.bpythonrc` should probably be removed.

X0rg:
Ok, fixed in d94565a.

sjohannes:
Thanks!

genodeftest:
Nice, thank you @X0rg!

PR #631, opened 2019-05-12
README: Point to exaile.org instead of readthedocs

sjohannes:
Potential users often expect all the info about a project in the GitHub readme. We're not doing this, but we should at least point them to our website, which has more info (e.g. screenshots).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,4 +15,4 @@ tagging, last.fm scrobbling, support for portable media players, podcasts,
 internet radio such as icecast and Soma.FM, ReplayGain, output via a secondary
 output device (great for DJs!), and much more.
 
-For more information see https://exaile.readthedocs.io/
+For more information see https://exaile.org/


genodeftest:
Good idea!

virtuald:
https doesn't actually work correctly.

virtuald:
Well. It works with www, not without www. Not quite sure why that is?

genodeftest:
https://www.exaile.org ships its own certificate while https://exaile.org ships github's certificate for [*.]github.com, [*.]github.io. Looking at this and the IP addresses, they are running on different servers. Do you have access (including DNS) to any of these servers?

This may help:
https://help.github.com/en/articles/securing-your-github-pages-site-with-https
https://help.github.com/en/articles/using-a-custom-domain-with-github-pages
https://help.github.com/en/articles/troubleshooting-custom-domains

virtuald:
We use github's hosting for the website, so we only have access through the exaile.github.io repo. I don't own the DNS for exaile.org (@synic does), but from my reading it seems that you can only do one of https://x or https://www.x, not both. Our jekyll CNAME is set to www.

virtuald:
Actually nevermind, that doesn't seem to be the case. It seems we need to update the DNS records. @synic can you update the DNS A records for exaile.org to 

    185.199.108.153
    185.199.109.153
    185.199.110.153
    185.199.111.153

synic:
Just saw this. Does this still need to be done?

virtuald:
I just checked the https urls, they all seem to have valid certificates, so I think we're good?

sjohannes:
https://exaile.org/ still has invalid cert, but I don't know where that needs to be fixed.

PR #636, opened 2019-05-20
Fix wrong gettext call on shortcut keys

sjohannes:
I don't think the key definitions (e.g. `<Primary>F`) are meant to be translated here. Anyway the `_(keys)` call is definitely wrong because the keys themselves are not currently marked for translation.


sjohannes:
Possibly, but in that case the `_` should have been on the actual strings (e.g. `_('<Primary>F')`) so that xgettext can extract them and put them in the .pot file.

PR #638, opened 2019-05-22
Tracklist edit fix double click

### LINKED ISSUE #633: Double-clicking on editable field doesn't trigger playback
Double-click on a non-selected track. This has to be done on an editable field, e.g. title, *on the text itself* (not on an empty region).

This will just select the field and not play the track.

virtuald:
Also incorporates @sjohannes rename tag on F2 because it seems cool too.

* Fixes #633

### DIFF
--- a/xlgui/widgets/playlist_columns.py
+++ b/xlgui/widgets/playlist_columns.py
@@ -168,44 +168,28 @@ class Column(Gtk.TreeViewColumn):
         )
 
 
-class CustomizedCellRendererText(Gtk.CellRendererText):
-    '''
-        The default GTK cell renderer triggers editing when any part of the column is
-        clicked. This is undesirable, as tracks can never be played if we allow columns
-        to be edited!
-
-        Instead, this only triggers editing if the mouse clicks the text of the
-        column.
-    '''
-
-    def do_start_editing(self, event, widget, path, background_area, cell_area, flags):
-        if event:
-            has_coords, x, y = event.get_coords()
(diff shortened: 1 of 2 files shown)


sjohannes:
Issue 1

1. Select a track.
2. Drag the track starting from its title.

Expected: Dragging starts, editing doesn't.
Actual: Both dragging and editing start.

---

Issue 2

exaile-inline-edit-1

1. Select track 1.
2. Hover on track 2.
3. Click track 1 on the empty area (red box).

Expected: Nothing happens.
Actual: Editing starts.

4. Select track 2.
5. Hover on track 3.
6. Click track 2 on the area above track 3's empty area (green box).

Expected: Editing starts.
Actual: Nothing happens.

virtuald:
Good catch! This should fix it.

xlgui/widgets/playlist_columns.py
```diff
@@ -168,44 +168,51 @@ def __repr__(self):
...
+        iter = model.get_iter(path)
+        self.data_func(self, self.cellrenderer, model, iter, None)
+
+        aa = self.cellrenderer.get_aligned_area(treeview, 0, cell_area)
+        x = event.x
+
+        if x >= aa.x and x <= aa.x + aa.width:
+            return True
+
+        return False
```
sjohannes: This makes me itch to change the last few lines to to `return ... <= ... <= ...`, but whichever you think is more readable.

PR #642, opened 2019-06-02
panel/playlists: Fix some dnd stuff.

sjohannes:
I barely understand all this dnd stuff, but these are some low-hanging fruits that I found.


virtuald:
Fixes issue for me.

Issue #643, opened 2019-06-07
UnicodeEncodeError


eXtreme-by:
> Traceback (most recent call last):
  File "/usr/local/lib/exaile/xlgui/playlist_container.py", line 83, in do_clicked
    self.notebook.create_new_playlist()
  File "/usr/local/lib/exaile/xlgui/playlist_container.py", line 206, in create_new_playlist
    name = page.get_page_name().decode('utf-8', errors='replace')
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-5: ordinal not in range(128)

Distributor ID:	LinuxMint
Description:	Linux Mint 19.1 Tessa x64

rokm:
I can reproduce this with following steps (Fedora 30, exaile git master):
1. Create a playlist
2. Rename the playlist to a name that contains a non-ASCII character (e.g., "Playlist with Š")
3. Save the playlist (using the name from Step 2)
4. Close & restart
5. Restore the saved playlist from Playlists -> Custom Playlists
6. Try to add a new playlist

This results in the same traceback as in the original report.

I suspect the saving and restoring part may be necessary in order for the playlist name to be set as page name.

rokm:
A related issue:
1. Create a playlist
2. Rename the playlist to a name that contains a non-ASCII character (e.g., "Playlist with Š")
3. Close exaile

This results in the following error:
```
INFO    : Saving state...
ERROR   : Error saving tab 'order1.current.Playlist with \xc5\xa0'
Traceback (most recent call last):
  File "/home/rok/Projects/exaile/xlgui/playlist_container.py", line 314, in save_current_tabs
    self.tab_manager.save_playlist(page.playlist, True)
  File "/home/rok/Projects/exaile/xl/playlist.py", line 2119, in save_playlist
    pl.save_to_location(os.path.join(self.playlist_dir, encode_filename(name)))
  File "/usr/lib64/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc5 in position 30: ordinal not in range(128)
INFO    : Bye!
```

virtuald:
Fixed by #644

PR #644, opened 2019-06-07
Fix unicode handling in playlist names

rokm:
A fix for issues with non-ASCII characters in playlist name (#643 ).

The original issue is caused by the decode() call on page name, which is already a unicode object. 

(Unless I am mistaken, the decode() call would make sense if the name was a str object containing UTF-8 characters. To get from unicode back to str, encode() should be used instead. But since default_playlist_name and consequently default_name_parts are also unicode objects, no conversion is actually required).

The second issue (error upon closing exaile) is fixed by ensuring that the name which we pass to encode_filename is a unicode object instead of str.

### DIFF
--- a/xlgui/playlist_container.py
+++ b/xlgui/playlist_container.py
@@ -203,7 +203,7 @@ class PlaylistNotebook(SmartNotebook):
 
         for n in range(self.get_n_pages()):
             page = self.get_nth_page(n)
-            name = page.get_page_name().decode('utf-8', errors='replace')
+            name = page.get_page_name()
             name_parts = [
                 # 'Playlist 99' => 'Playlist '
                 name[0 : len(default_name_parts[0])],
--- a/xlgui/widgets/notebook.py
+++ b/xlgui/widgets/notebook.py
@@ -377,6 +377,9 @@ class NotebookTab(Gtk.EventBox):
         """
         name = self.entry.get_text()
 
+        if not isinstance(name, unicode):
+            name = name.decode('utf-8', errors='replace')
+
         if name.strip() != "" and not self.entry.props.editing_canceled:
             self.page.set_page_name(name)
             self.label.set_text(name)


xl/playlist.py
```diff
@@ -2115,6 +2115,11 @@ def save_playlist(self, pl, overwrite=False):
                 playlist should it happen to already exist
         """
         name = pl.name
+
+        # Make sure name is unicode
+        if not isinstance(name, unicode):
```
sjohannes: Is it possible for `pl.name` to be an str? If it's always unicode we can just `assert(isinstance(name, unicode))` here, which will help us clarify the types in the future when moving to Python 3.

xl/playlist.py
```diff
@@ -2115,6 +2115,11 @@ def save_playlist(self, pl, overwrite=False):
                 playlist should it happen to already exist
         """
         name = pl.name
+
+        # Make sure name is unicode
```
sjohannes: Is it possible for `pl.name` to be an str? If it's always unicode we can just `assert(isinstance(name, unicode))` here, which will help us clarify the types in the future when moving to Python 3.

xl/playlist.py
```diff
@@ -2170,7 +2175,13 @@ def load_names(self):
                 except Exception:
                     logger.exception("Failed loading playlist: %r", path)
                 else:
-                    existing.append(pl.name)
+                    name = pl.name
+
+                    # Make sure name is unicode
```
sjohannes: (As above.)

xl/playlist.py
rokm: At the moment, it does seem to happen that ```pl.name``` is a str object:
1) Open a new playlist
2) Rename a playlist
3) Close exaile - the conversion code path will be triggered

Although on subsequent open & close, this does not seem to be the case anymore, so it might also be a bug in the name-setting codepath?

xl/playlist.py
rokm: Same as above; the conversion does seem to be necessary when re-starting exaile after renaming a playlist. But not on subsequent restarts...

rokm:
Hmm, it seems that the last two commits can be replaced by forcing conversion to unicode object in 
[end_rename](https://github.com/exaile/exaile/blob/358871fe3725b19a1ea5a652d915e666dda9b36d/xlgui/widgets/notebook.py#L378) handler for the notebook.

The question is, do we want to deal with this now, or do we take care of this issue while converting to python3 (where this and similar issues are likely to show up and will need to be consistently fixed anyway)?

sjohannes:
Usually I directly `.decode('utf-8')` whenever I'm grabbing values from GTK. That way we minimise the number of bytes/str objects getting passed around within our codebase, and hopefully that makes the Python 3 porting easier.

rokm:
OK, pushed a revised version. We now ensure that name is a unicode object by converting it in end_rename, if necessary.

Issue #608, opened 2019-04-05
Update CDDB bindings (used for plugin "cd", part of python 3 migration) or port to musicbrainz


genodeftest:
We're currently using CDDB-py (known as python-cddb on some linux distributions) for retreiving CD metadata. This is the only remaining dependency except spydaap which is not available for python 3. This issue is related to #246.

CDDB-py:
* uses freedb.org servers
* http://cddb-py.sourceforge.net/
* https://sourceforge.net/projects/cddb-py/
* pro: very stable
* cons: last update in 2003, python 1/2 code plus native C library
* License: GPLv2

PyCDDB: 
* uses freedb.org servers
* http://pycddb.sourceforge.net/
* https://sourceforge.net/projects/pycddb/
* License: GPLv2
* requires discid (same author)
  * https://sourceforge.net/projects/discid/
  * http://discid.sourceforge.net/
  * License: GPLv2
  * cons: python code plus native C library
* untested by myself

libcddb:
* uses freedb.org servers
* http://libcddb.sourceforge.net/
* https://sourceforge.net/projects/libcddb/
* pro: still used in many distros
* con: native library, no bindings
⇒ no option

libdiscid:
* uses musicbrainz (most compatible with the rest of Exaile)
* official library by musicbrainz: https://musicbrainz.org/doc/libdiscid
* multi-platform
* still partially maintained
* bindings:
  * https://pypi.org/project/python-libdiscid/
  * https://pypi.org/project/discid/

cd-discid command line tool:
* no database connection
* http://linukz.org/cd-discid.shtml
* probably not ported to windows
⇒ no option

python-musicbrainzngs:
* pure python implementation
* https://pypi.org/project/musicbrainzngs/
* already used by Exaile

Our options:
1. port CDDB-py and/or PyCDDB to python 3. I would try to port the native code to python code too. Maintain it ourselves. In this case we may think about putting it into our own source tree as the license is compatible.
2. use bindings to libcddb. This solution is less flexible (uses native code) but probably the easier way to go.
3. use python bindings to libdiscid, reuse musicbrainzngs which we're using in the musicbrainzcovers plugin anyway.

genodeftest:
I'd go for the last solution if that works out.

sjohannes:
> 2. use bindings to libdiscid. This solution is less flexible (uses native code) but probably the easier way to go.
> 3. reuse musicbrainzngs which we're using in the musicbrainzcovers plugin anyway.

[...]

PR #645, opened 2019-06-22
Update dependency descriptions and spydaap dependency

genodeftest:


DEPS
```diff
@@ -39,42 +39,62 @@ Device detection:
 
 CD info:
 
-* python-cddb
+* cddb (python2)
```
sjohannes: We're using [this one](http://cddb-py.sourceforge.net/), right? I recommend keeping it as "python-cddb" as that's the package name on most distros, maybe add the URL as well for clarity. Just "cddb" I think is too confusing because there are so many projects with similar names.

genodeftest: I hope to drop the cddb dependency soon anyway, see #608.

genodeftest:
> We don't need to mention Python 2, I think; it's already implied by the Python 2.7 requirement.

On some packages it is not clear whether they are python packages or "native" packages. This is why I added `(python2)`. I'd like to keep it that way.

genodeftest:
> We don't need to mention Python 2, I think; it's already implied by the Python 2.7 requirement.

On some packages it is not clear whether they are python packages or "native" packages. This is why I added `(python2)`. I'd like to keep it that way. Is that ok for you?

sjohannes:
Yeah, it's fine. Thanks.

genodeftest:
Thank you for your review!

PR #650, opened 2019-07-14
Replace cddb by libdiscid

### LINKED ISSUE #608: Update CDDB bindings (used for plugin "cd", part of python 3 migration) or port to musicbrainz

genodeftest:
This is an almost complete rework of the cd plugin to replace the CDDB dependency by libdiscid. CDDB is the last external dependency not available on python 3 so this is a blocker to python 3.

If you wish, I may squash the changes into less or one single commit.

I already did test this code, but you can probably still find issues when testing it too – especially because you will test it differently.

I introduced quite some `TODO:`s in the code for places where the code could be improved. The most important step would be communicating some progress to the user.

Fixes #608, which is part of #246.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -42,7 +42,7 @@ Device detection:
 
 CD info:
 
-* python-libdiscid (optional)
+* python-libdiscid or python-discid (optional on linux, required to use musicbrainz)
 * python-musicbrainzngs (optional)
 
 DAAP plugins (daapserver and daapclient):
--- a/DEPS
+++ b/DEPS
@@ -42,7 +42,8 @@ Device detection:
 
 CD info:
 
-* cddb (python2), from http://cddb-py.sourceforge.net/
+* python-libdiscid (optional)
+* python-musicbrainzngs (optional)
 
 DAAP plugins (daapserver and daapclient):
 
--- a/plugins/cd/PLUGININFO
+++ b/plugins/cd/PLUGININFO
@@ -1,7 +1,7 @@
 Version='4.0.0'
 Authors=['Aren Olson <reacocard@gmail.com>']
 Name=_('CD Playback')
-Description=_('Adds support for playing audio CDs.\n\nRequires UDisks2 to autodetect CDs\nRequires cddb-py (%s) to look up tags.') % 'http://cddb-py.sourceforge.net/'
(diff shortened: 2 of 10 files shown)


genodeftest:
@sjohannes , @virtuald , @rokm : Could one of you please review these changes a bit?

genodeftest:
Hm, I should have tested it after rebasing, the panel looks broken now. Will close for now and fix it.

PR #654, opened 2019-07-28
Use GLib.set_prgname() to set program name to 'exaile'

### LINKED ISSUE #653: Exaile icon is incorrect on KDE Plasma Wayland session

1. start KDE Plasma Wayland session
2. open exaile
3.
4.

### Expected Behavior

Both the title bar and the task switcher (alt+tab menu) show Exaile icon.
### Current Behavior

We can see a Wayland generic icon in the title bar and when we press alt+tab.
### Possible Solution

maybe this link can help you
https://community.kde.org/Guidelines_and_HOWTOs/Wayland_Porting_Notes#Application_Icon
### Environment

* Operating System and version:
Arch Linux running KDE Plasma 5.16.3 desktop environment on Wayland

* Exaile Version: 4.0.0

exaile

rokm:
This is necessary to change the app_id under Plasma wayland
session from default 'exaile.py' to 'exaile', which in turn
allows matching with the exaile.desktop file. Fixes #653.

### DIFF
--- a/xl/main.py
+++ b/xl/main.py
@@ -42,11 +42,11 @@ from xl.externals.sigint import InterruptibleLoopContext
 from xl.nls import gettext as _
 
 # Imported later to avoid PyGObject imports just for --help.
-Gio = Gtk = common = xdg = None
+GLib = Gio = Gtk = common = xdg = None
 
 
 def _do_heavy_imports():
-    global Gio, Gtk, common, xdg
+    global GLib, Gio, Gtk, common, xdg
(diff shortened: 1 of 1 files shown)


genodeftest:
Looks like a nice solution until we're using GApplication (which may be in the far future or never).

genodeftest:
This does not break the current behavior in gnome-shell. Thank you!

Would we want to do a separate point release (4.0.1) for this fix?

PR #659, opened 2019-10-10
Use “translator-credits” for more clarity.

### LINKED ISSUE #590: Use "translator-credits" instead of "Unknown translator"
Instead of using "Unknown translator" for translator names in the po files i suggest using "translator-credits":

Now:

    #. TRANSLATORS: Add your own name here. Separate translator names with newlines.
    #: ../data/ui/about_dialog.ui:25
    msgctxt "About dialog translator credits"
    msgid "Unknown translator"

The future (maybe):

    #. TRANSLATORS: Add your own name here. Separate translator names with newlines.
    #: ../data/ui/about_dialog.ui:25
    msgctxt "About dialog translator credits"
    msgid "translator-credits"

I see "translator-credits" being used in multiple other software.
Not sure if it is some kind of standard but translators can properly easy recognize it.

Also instead of writing Unknown translator in the about/contributor dialog  then maybe just dont write anything.

fitojb:
Fixes #590

Also ran `make pot`

### DIFF
--- a/data/ui/about_dialog.ui
+++ b/data/ui/about_dialog.ui
@@ -23,7 +23,7 @@ Fabiano Fidêncio
 Joseph S. Atkinson
 Christian Stadelmann
 Rok Mandeljc</property>
-    <property name="translator_credits" translatable="yes" context="About dialog translator credits" comments="TRANSLATORS: Add your own name here. Separate translator names with newlines.">Unknown translator</property>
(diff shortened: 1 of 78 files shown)


fitojb:
Ping? @exaile

sjohannes:
Thanks, I've merged this in. My only worry was that we'd have a "Translated by: translator-credits" line in the About dialog, but after testing I found that GTK automatically removes the line when it matches that. So now we get the best of both worlds. Thanks!

PR #663, opened 2019-11-11
Use {}/format instead of % for numerical ngettext strings

Drudenhaus:
Addresses #599.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ random/
 
 # VS Code settings
 .vscode/
+
+# Virtual environment
+.venv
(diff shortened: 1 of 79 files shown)


Drudenhaus:
I can reformat per the suggested changes in the travis output, but is `make pot` the correct way to update the line numbers in the translation files? It made a ton of line reference changes to the translation files that are unrelated to the changes I made. e.g. 
```
sean@drudenhaus ~/workspace/exaile $ git diff --shortstat
 77 files changed, 8893 insertions(+), 10214 deletions(-)
```

PR #671, opened 2020-01-16
[py3] Some python 3.x meta stuff

genodeftest:


DEPS
```diff
@@ -11,17 +11,25 @@ Core dependencies
 
 Essential:
 
-* python (== 2.7.x)
+* python3 (>= 3.6)
 * bsddb (distributed with stdlib) or bsddb3
```
sjohannes: bsddb was removed from the stdlib, so this now requires the external bsddb3 module.

DEPS
```diff
@@ -11,17 +11,25 @@ Core dependencies
...
 * gtk+ >= 3.10
 * gstreamer >= 1.4
 * gstreamer-plugins-base >= 1.6
 * gstreamer-plugins-good >= 1.4
-* mutagen (>= 1.10)
-* dbus-python
-* pygobject (>= 3.13.2)
```
sjohannes: We probably still need to mention pygobject and its version.

DEPS
```diff
@@ -11,17 +11,25 @@ Core dependencies
...
 * GI typelib files for GTK+, GStreamer (including gstreamer-plugins-base) and cairo (python-gi-cairo on Ubuntu)
+  * Packages on Debian and Ubuntu:
+    * python3-gi
+    * python3-gi-cairo
+    * gir1.2-gtk-3.0
+    * gir1.2-gstreamer-1.0
+    * gir1.2-gst-plugins-base-1.0
+  * Packages on Fedora:
+    * python3-cairo
+    * python3-dbus
```
sjohannes: dbus-python is not a GI library, so it shouldn't be under this section.

DEPS
```diff
@@ -11,17 +11,25 @@ Core dependencies
...
 * gtk+ >= 3.10
 * gstreamer >= 1.4
 * gstreamer-plugins-base >= 1.6
 * gstreamer-plugins-good >= 1.4
-* mutagen (>= 1.10)
-* dbus-python
-* pygobject (>= 3.13.2)
-* python-cairo
+* python3-mutagen (>= 1.10)
 * GI typelib files for GTK+, GStreamer (including gstreamer-plugins-base) and cairo (python-gi-cairo on Ubuntu)
```
sjohannes: You can remove the part about python-gi-cairo.

DEPS
genodeftest: Thanks!

DEPS
genodeftest: You are right, thanks!

DEPS
genodeftest: Done.

genodeftest:
Thanks for the review, @sjohannes !

DEPS
genodeftest: @sjohannes : Any objections?

sjohannes: Hm, in that case maybe we should attach the version number to python3-gi (Debian) and python3-gobject (Fedora)? It's not a big deal, though, and I'm not sure that number is still accurate anyway.

genodeftest: In practice, nobody ever tests exaile on old python3-gi versions. So I'd just fill in the oldest version our CI is testing against, that's ubuntu18.04 with python3-gi 3.22.0.

genodeftest: I also updated some other dependencies accordingly (gtk3, gstreamer, …)

PR #672, opened 2020-01-16
[py3] Several fixes for the plugins module

genodeftest:


xl/plugins.py
```diff
@@ -76,11 +76,14 @@ def load_plugin(self, pluginname, reload_plugin=False):
         path = self.__findplugin(pluginname)
         if path is None:
             return False
-        sys.path.insert(0, path)
-        plugin = imp.load_source(pluginname, os.path.join(path, '__init__.py'))
+
+        spec = importlib.util.spec_from_file_location(pluginname, os.path.join(path, '__init__.py'))
+        plugin = importlib.util.module_from_spec(spec)
+        sys.modules[spec.name] = plugin
```
sjohannes: Hm, what does spec.name contain here? Would it conflict with other core or plugin modules?

genodeftest: `spec.name` is always the same as `pluginname`, both if you follow `importlib.util.spec_from_file_location`'s source code and also while debugging. To clarify that, I'm using `pluginname` instead of `spec.name` now. Also, I made sure that the plugin cannot break exaile no matter how its name is. And I also remove the plugin now immediately after loading it.

genodeftest: @sjohannes : Better now?

PR #673, opened 2020-01-16
[py3] SyntaxWarning and removed API fixes

nicfit:
This PR is based on py3 branch and testing with Python 3.8:

- Python3 SyntaxWarning fix: is not "" (use !=)
- cgi.parse_qs deprecated in python3 and removed in python3.8; use urllib.parse


sjohannes:
Thanks!

PR #675, opened 2020-01-16
py3: fix documentation generation on python 3.6+ (try 2)

rokm:
Second attempt at fixing the documentation generation issue, tested both with python 3.6.x on Ubuntu 18.04 and 3.7.x on Fedora 31.

This time, the ```__mro_entries__``` issue is addressed by raising the ```AttributeError```.

However, this still leaves the ```TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases``` error, which is caused by the mocked ```GObject.GObject``` being involved in multiple inheritance in ```xl.common.ProgressThread```. For this to work, we need our ```Mock``` object for the ```GObject``` module return a type instead of an instance for ```GObject``` (class).

Due to the name repetition (```gi.repository.GObject.GObject```, with ```GObject``` referring both to the module and the class) and ```__getattr__``` being a class method, distinction between the two is handled by introduction of a new ```MockGiModule``` class, which now also collects all mock functionality of GI (sub)modules.

### DIFF
--- a/doc/mocks.py
+++ b/doc/mocks.py
@@ -8,29 +8,32 @@
 import sys
 
 
-class Mock:
-
-    __all__ = []
-
+class MockGiModule:
     def __init__(self, *args, **kwargs):
         pass
 
     def __call__(self, *args, **kwargs):
-        return Mock()
-
-    # Fix for python 3.7+ (issue #667)
-    def __mro_entries__ (self, bases):
-        return ()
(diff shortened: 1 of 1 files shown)


genodeftest:
Documentation builds and works fine and nothing breaks so feel free to merge imho. Thanks for caring!

PR #676, opened 2020-01-17
[py3] Fix SyntaxWarnings

nicfit:
```
/home/travis/devel/exaile/xlgui/preferences/collection.py:37:
SyntaxWarning: "is not" with a literal. Did you mean "!="?
  return [v.lower() for v in default_strip_list.split(' ') if v is not '']
/home/travis/devel/exaile/xlgui/preferences/collection.py:54:
SyntaxWarning: "is not" with a literal. Did you mean "!="?
  values = [v.lower() for v in self.widget.get_text().split(' ') if v is not '']
```

### DIFF
--- a/plugins/daapclient/__init__.py
+++ b/plugins/daapclient/__init__.py
@@ -566,7 +566,7 @@ class DaapConnection:
                             temp.set_tag_raw(field, [tag], notify_changed=False)
 
                     except Exception:
-                        if field is 'tracknumber':
+                        if field == 'tracknumber':
                             temp.set_tag_raw('tracknumber', [0], notify_changed=False)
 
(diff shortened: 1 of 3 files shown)


genodeftest:
Nice catch, thank you!

Does this mean the second line of this code is also wrong?

    except Exception:
        if field is 'tracknumber':
            temp.set_tag_raw('tracknumber', [0], notify_changed=False)

Source: https://github.com/exaile/exaile/blob/py3/plugins/daapclient/__init__.py#L569

nicfit:
I did some grepping but did not catch that one in daapclient.. but that's the same and will SyntaxWarn if hit.

nicfit:
@genodeftest Fixed that one too.

PR #677, opened 2020-01-19
Mod format: bump soname version, mention in DEPS, add logging

genodeftest:
Changes:
* bump libmodplug soname version to match fedora31, debian8+, ubuntu12.10+
* mention in DEPS that you need to install libmodplug for modplug
  metadata
* add logging to give a hint if metadata fetching does not work for the
  MOD file format

### DIFF
--- a/xl/metadata/mod.py
+++ b/xl/metadata/mod.py
@@ -27,12 +27,17 @@
 
 from xl.metadata._base import BaseFormat
 
+import logging
 import os
+
+
+logger = logging.getLogger(__name__)
+
 
 try:
     import ctypes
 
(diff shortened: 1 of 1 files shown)


DEPS
```diff
@@ -68,6 +68,10 @@ Lyrics from lyrics.wikia.com (lyricwiki):
 
 * Beautifulsoup4 (python3)
 
+Modplug file support:
+
+* libmodplug
```
sjohannes: Wouldn't this already be required by the GStreamer plugin?

genodeftest: You're right, it's in `gstreamer1-plugins-bad-free-extras`. I've removed this change from the PR.

PR #678, opened 2020-01-19
[WIP] Py3 use with for open and other small fixes

genodeftest:


xl/playlist.py
```diff
@@ -2229,22 +2228,21 @@ def load_from_location(self, location):
...
-                break
+                with open(loc, 'r') as fd:
+                    return self.__load_from_location_internal(fd)
             except Exception:
                 pass
-        if f is None:
-            return []
+        return None
+
+    def __load_from_location_internal(self, filedescriptor):
```
rokm: Considering the rest of the code in this PR, perhaps ```filedescriptor``` should be named ```f``` or ```fd``` for consistency sake?

xl/settings.py
```diff
@@ -305,20 +305,19 @@ def save(self):
         logger.debug("Saving settings...")
 
         with open(self.location + ".new", 'w') as f:
-            self.write(f)
-
             try:
-                # make it readable by current user only, to protect private data
-                os.fchmod(f.fileno(), 384)
+                # make it readable by current user only
+                os.fchmod(f, S_IRUSR | S_IWUSR)
```
rokm: These should probably be ```stat.S_IRUSR``` and ```stat.S_IWUSR```, unless I missed an import somewhere. Otherwise, this will silently fail.

rokm: Using ```stat.S_IRUSR``` will require ```import stat```.

Also, ```fchmod()``` requires file descriptor, so the first argument should be ```f.fileno()``` as in original version.

xl/playlist.py
genodeftest: Personally, I don't like short identifiers like `f` because they don't help anything. I've shortened it to `fd` to match the style anyway.

xl/settings.py
genodeftest: Thanks! I've updated the code now but I haven't tested it yet so I'm marking it as [WIP].

rokm: Looks (and works) OK now.

plugins/audioscrobbler/__init__.py
```diff
@@ -259,16 +259,14 @@ def _save_cache_cb(self, a, b, c):
...
 
     def load_cache(self):
         try:
-            f = open(self.cachefile, 'rb')
-            cache = pickle.load(f)
-            f.close()
-            scrobbler.SUBMIT_CACHE = cache
+            with open(self.cachefile, 'rb') as f:
+                cache = pickle.load(f)
+                scrobbler.SUBMIT_CACHE = cache
```
sjohannes: This line can be outside the `with`.

xl/playlist.py
```diff
@@ -1471,15 +1471,16 @@ def load_from_location(self, location):
...
             try:
-                f = open(loc, 'r')
+                with open(loc, 'r') as fd:
+                    return self.__load_from_location_internal(fd)
                 break
             except Exception:
                 pass
-        if not f:
-            return
+        return
```
sjohannes: This can be removed.

xl/playlist.py
```diff
@@ -2209,18 +2209,17 @@ def save_to_location(self, location):
...
-            f = open(location, "w")
-        for playlist in self.playlists:
-            f.write(playlist)
-            f.write('\n')
+            destination = location
+        with open(destination, 'w') as f:
+            for playlist in self.playlists:
+                f.write(playlist)
+                f.write('\n')
+            f.write("EOF\n")
```
sjohannes: These can be `print(..., file=f)` without the `\n`, if you prefer. Doesn't matter though.

xl/playlist.py
```diff
@@ -2229,22 +2228,21 @@ def load_from_location(self, location):
...
             try:
-                f = open(loc, 'r')
-                break
+                with open(loc, 'r') as fd:
+                    return self.__load_from_location_internal(fd)
             except Exception:
                 pass
-        if f is None:
-            return []
+        return None
```
sjohannes: Is this okay? Previously this returned `[]`.

xl/playlist.py
```diff
@@ -1471,15 +1471,16 @@ def load_from_location(self, location):
         # note - this is not guaranteed to fire events when it sets
         # attributes. It is intended ONLY for initial setup, not for
         # reloading a playlist inline.
-        f = None
         for loc in [location, location + ".new"]:
             try:
-                f = open(loc, 'r')
+                with open(loc, 'r') as fd:
+                    return self.__load_from_location_internal(fd)
                 break
```
sjohannes: Actually this can also be removed.

PR #679, opened 2020-01-19
Python: Fix some mypy findings

genodeftest:
I ran mypy on exaile.py and this is what I could fix.

### DIFF
--- a/xl/externals/gi_composites.py
+++ b/xl/externals/gi_composites.py
@@ -19,6 +19,7 @@
 from os.path import abspath, join
 
 import inspect
+from typing import List
 import warnings
 
 from gi.repository import Gio
@@ -152,7 +153,7 @@ class _Child:
         is called
     '''
 
-    __slots__ = []
+    __slots__: List[str] = []
 
     @staticmethod
     def widgets(count):
--- a/xl/trax/util.py
+++ b/xl/trax/util.py
@@ -103,7 +103,7 @@ def get_tracks_from_uri(uri):
 def sort_tracks(
     fields: Iterable[str],
     items: Iterable[T],
-    trackfunc: Optional[Callable[[T], Track]] = None,
+    trackfunc: Optional[Callable[[Track], Track]] = None,
     reverse: bool = False,
     artist_compilations: bool = False,
 ) -> List[T]:
--- a/xlgui/widgets/playlist_columns.py
+++ b/xlgui/widgets/playlist_columns.py
@@ -32,6 +32,7 @@ from gi.repository import Pango
 
 import logging
 import time
+from typing import Any, Type
 
 from xl import common, event, player, settings, providers
 from xl.common import classproperty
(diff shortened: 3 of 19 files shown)


xl/externals/gi_composites.py
```diff
@@ -152,7 +153,7 @@ class _Child:
         is called
     '''
 
-    __slots__ = []
+    __slots__: List[Gtk.Widget] = []
```
sjohannes: Should be `List[str]`, assuming this is the usual Python `__slots__`.

xlgui/widgets/smart_playlist_editor.py
```diff
@@ -254,7 +256,7 @@ def __init__(self):
     N_('Playlist'): '__playlist',  # not a real tag
 }
 
-_REV_NMAP = {}
+_REV_NMAP: Mapping[str, str] = {}
```
sjohannes: I think `Dict` would be more suitable here.

xl/externals/gi_composites.py
genodeftest: Yep.

xlgui/widgets/smart_playlist_editor.py
genodeftest: `Mapping` is more generic than `Dict`, which could be nice sometimes, but as the code creates a `Dict`, that's probably better.

genodeftest:
Please note that the last two commits in this PR are new (added after @sjohannes' review).

xl/trax/util.py
```diff
@@ -103,7 +103,7 @@ def get_tracks_from_uri(uri):
 def sort_tracks(
     fields: Iterable[str],
     items: Iterable[T],
-    trackfunc: Optional[Callable[[T], Track]] = None,
+    trackfunc: Optional[Callable[[Track], Track]] = None,
```
sjohannes: This doesn't seem right. From my understanding, `tracfunc` is meant to transform some unknown type `T` (element of `items`) into a `Track`. That is, it's a function with this signature:

    def trackfunc(item: T) -> Track:
        pass

I'll try checking with mypy later.

xlgui/widgets/playlist_columns.py
```diff
@@ -53,7 +54,7 @@ class Column(Gtk.TreeViewColumn):
     formatter = classproperty(lambda c: TrackFormatter('$%s' % c.name))
     size = 10  # default size
     autoexpand = False  # whether to expand to fit space in Autosize mode
-    datatype = str
+    datatype: Type[Any] = str
```
sjohannes: `type` should work here.

xl/trax/util.py
sjohannes: OK, I've tested this and I think the existing annotation is correct (`Callable[[T], Track]`), but the body needs this change:

```py
if trackfunc is None:
    trackfunc = lambda tr: cast(Track, tr)
```

That is:

* If `trackfunc` is specified, `T` can be anything. `trackfunc` then transforms this `T` into a `Track`.
* If `trackfunc` is not specified, `T` must already be `Track`.

I don't know if this can be expressed easily in the type signature, but I think for now the cast is good enough.

PR #680, opened 2020-01-19
Python: Replace calls to deprecated logging.warn()

genodeftest:
logging.warning() is just the same without deprecation notice.

Note: `logging.warn()` is not only deprecated since python 3.3 but has never been documented ([Details](https://bugs.python.org/issue13235)).


genodeftest:
Thanks for the review!

PR #684, opened 2020-01-27
HACK: AutoScrollTreeView: Restore voffset on drag end

fuhry:
When dragging to reorder a playlist, note the scroll position and scroll
back to that location after a brief timeout. This does result in the
playlist temporarily scrolling to the top and then back to the dropped
position for about 50ms, but is better than always being sent to the top
of the playlist upon any reordering operation.

This only fixes the issue for reordering the playlist. It is still an issue when
new items are added.

TODO: Fix properly


virtuald:
Closed in favor of #766

Issue #683, opened 2020-01-27
Replace Current as Response for Sidebar Double-Click on a Playlist


dwbapst:
### Expected Behavior
I'd really like the ability to just double-click on a playlist in the playlists sidebar and have the playlist contents wholly replace the open 'playlist' queue/tab, instead of opening a second playlist/queue tab. I'd like to minimize situations in which I open multiple playlist tabs ever.

### Current Behavior
When I double-click one of my custom playlists in the sidebar, it opens a new playlist tab. This is irrespective of having the 'Replace Content on Side Pane Double Click' option, which suggests that this might happen, but I imagine that option might have originally just been intendedfor allowing double-clicking songs/albums/etc in the collection to replace the current content of the playlist tab, not playlists replacing the current playlist. For what its worth, the last version of Clementine has the behaviour I am looking for, but I like Exaile much better!

### Possible Solution
Well, I guess it would take to just implement another option to enable this feature. I'm an R developer, I don't know if you want me touching Python. But if its a relatively minor change, I guess I would be happy to help make it happen.

### Environment
* Operating System and version: Linux Mint 19.3
* Exaile Version: 4.0.2

sjohannes:
I think this makes sense if the "Replace content on side pane double click" config is enabled.

PR #686, opened 2020-02-04
Double-clicked playlists in playlist pane now replace current playlist if enabled in options

dwbapst:
Addresses #683 

Playlists double-clicked in the Playlists pane when 'double-click in lefthand pane replaces current' option is enabled now replace the current playlist entirely. This also applies to if a particular track under a particular playlist is double-clicked - that track will replace the existing playlist as well, just as if I had double-clicked that track in Collection.

PS: I am an R developer of ten years, and this is the first time I have ever edited Python code (even though I helped teach several Python workshops). I apologize for any striking errors, frankly I am amazed I was even able to do this.

### DIFF
--- a/xlgui/panel/playlists.py
+++ b/xlgui/panel/playlists.py
@@ -210,10 +210,9 @@ class BasePlaylistPanelMixin(GObject.GObject):
             if isinstance(item, (Playlist, SmartPlaylist)):
                 if replace:
                     # what we want to do if replace is true
-                    playlist = self.get_selected_item(raw=raw)
+                    selected_playlist = self.tree.get_selected_page()
                     tracks = [track for track in playlist]
                     self.emit('replace-items', tracks)
-        
                 else:
                     # for smart playlists
                     if hasattr(item, 'get_playlist'):
--- a/xlgui/panel/playlists.py
+++ b/xlgui/panel/playlists.py
@@ -233,9 +233,10 @@ class BasePlaylistPanelMixin(GObject.GObject):
                     self.emit('playlist-selected', item)
             else:
                 if replace:
-                    self.emit('replace-items', [item.track]))
+                    self.emit('replace-items', [item.track])
                 else:             
-                    self.emit('append-items', [item.track], True)    
+                    self.emit('append-items', [item.track], True)  
+                      
     def add_new_playlist(self, tracks=[], name=None):
         """
             Adds a new playlist to the list of playlists. If name is
--- a/xlgui/panel/playlists.py
+++ b/xlgui/panel/playlists.py
@@ -211,7 +211,7 @@ class BasePlaylistPanelMixin(GObject.GObject):
                 if replace:
                     # what we want to do if replace is true
                     selected_playlist = self.tree.get_selected_page()
-                    tracks = [track for track in playlist]
+                    tracks = [track for track in selected_playlist]
                     self.emit('replace-items', tracks)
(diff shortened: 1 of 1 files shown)


dwbapst:
My pull request with the most recent build now passes all checks - is there anything I need to do, or should do to proceed further, such as formally requesting review?

rokm:
Sorry for the delay! Functionality-wise the proposed changes look good - the contents of current tab are now properly replaced when either a smart playlist, a playlist, or a track from a playlist are double clicked and "Replace content on side pane double click" is enabled.

I have two comments, though:

1. Seeing as the PR consists of one main commit and several follow-up fixes, can you please squash them all into a single commit? (You can do an interactive rebase on your local tree, and then force push to branch to update the PR).

2. The added playlist-replacement codepath introduces an additional way for retrieving the selected playlist via `self.tree.get_selected_page()`. Is there a particular reason for doing so, and not simply re-using the `item` from the existing codepath?
In other words, leaving the smart playlist vs. playlist selection as it is, and then either emitting `playlist-selected`, or gathering tracks from `item` and emitting `replace-items`?

dwbapst:
Thanks, @rokm ! I'm glad to hear it looks good from a functional perspective. I'll squash the commits as soon as I can.

I'm not sure I understand your second point. Could you elaborate a bit more? Like I said, this is the first time I've coded in Python rather than R, so I think its just a vocabulary issue.

rokm:
> I'm not sure I understand your second point. Could you elaborate a bit more? Like I said, this is the first time I've coded in Python rather than R, so I think its just a vocabulary issue.

When replacement is enabled, the selected playlist lookup is duplicated:
```python
selected_playlist = self.tree.get_selected_page()
```
At that point, the selected playlist should already be available in the ```item``` variable.
The ```item``` might contain a smart playlist, and would thus require an additional
 ```get_playlist()``` call - which is what the existing code (in the non-replace branch)
 is doing. (The ```get_selected_page()``` also does that behind the scenes, but it [also quietly ignores errors](https://github.com/exaile/exaile/blob/2f3b5a306f36826bc33c4f4548295fb10881a500/xlgui/panel/playlists.py#L972).)

So I was wondering if it would not be better to simply reuse the existing playlist lookup and only adjust the action at the end:
```python
def open_item(self, tree, path, col):
    """
        Called when the user double clicks on a playlist,
        also called when the user double clicks on a track beneath
        a playlist.  When they active a track it opens the playlist
        and starts playing that track
    """
    iter = self.model.get_iter(path)
    item = self.model.get_value(iter, 2)

    replace = settings.get_option('playlist/replace_content', False)

    if item is not None:
        if isinstance(item, (Playlist, SmartPlaylist)):            
            # for smart playlists
            if hasattr(item, 'get_playlist'):
                try:
                    item = item.get_playlist(self.collection)
                except Exception as e:
                    logger.exception("Error loading smart playlist")
                    dialogs.error(
                        self.parent,
                        _("Error loading smart playlist: %s") % str(e),
                    )
                    return
            else:
                # Get an up to date copy
                item = self.playlist_manager.get_playlist(item.name)
                # item.set_is_custom(True)

            if replace:
                tracks = [track for track in item]
                self.emit('replace-items', tracks)
            else:
                self.emit('playlist-selected', item)
        else:
            if replace:
                self.emit('replace-items', [item.track])
            else:
                self.emit('append-items', [item.track], True)
```

PR #696, opened 2020-04-15
Fixes for ipconsole plugin

### LINKED ISSUE #694: IPython Console doesn't open
1. Enable IPython Console plugin.
2. Tools → Show IPython Console.

Expected: IPython console window shows.  
Actual: Nothing shows and we get a traceback on the actual console.

The last time I tried the plugin was with IPython 5 (based on 1e2c139347e57b143170f556b2bb6021a0ebaa28). Now it's at version 7 and things don't work. I fixed a simpler problem in https://github.com/exaile/exaile/commit/d22df4ddd12f77e9a5aec7c427724989b64c3b8c but now getting:

```pytb
  File "/exaile/plugins/ipconsole/__init__.py", line 214, in <lambda>
    callback=lambda *_args: self.__show_console(),
  File "/exaile/plugins/ipconsole/__init__.py", line 244, in __show_console
    self.__console_window = IPythonConsoleWindow(
  File "/exaile/plugins/ipconsole/__init__.py", line 159, in __init__
    self.__ipv = IPView(namespace)
  File "/exaile/plugins/ipconsole/__init__.py", line 80, in __init__
    ip.IPythonView.__init__(self)
  File "/exaile/plugins/ipconsole/ipython_view.py", line 608, in __init__
    self.execute()
  File "/exaile/plugins/ipconsole/ipython_view.py", line 209, in execute
    self.IP.input_splitter.push(line)
AttributeError: 'TransformerManager' object has no attribute 'push'
```

rokm:
This PR attempts to fix #694 by first back-porting the upstream fixes for ipython_view.py, and then applying some additional fixes for issues that were discovered during testing.

### DIFF
--- a/plugins/ipconsole/ipython_view.py
+++ b/plugins/ipconsole/ipython_view.py
@@ -216,7 +216,7 @@ class IterableIPShell:
             self.IP.showtraceback()
         else:
             if self.no_input_splitter:
-                self.lines.append(self.IP.raw_input(self.prompt))
+                self.lines.append(line)
                 self.iter_more = self.IP.check_complete('\n'.join(self.lines))[0] == 'incomplete'
             else:
                 self.IP.input_splitter.push(line)
--- a/plugins/ipconsole/ipython_view.py
+++ b/plugins/ipconsole/ipython_view.py
@@ -346,7 +346,7 @@ class IterableIPShell:
                 return str1
 
             if possibilities[1]:
-                common_prefix = reduce(_commonPrefix, possibilities[1]) or line[-1]
+                common_prefix = reduce(_commonPrefix, possibilities[1]) or split_line[-1]
                 completed = line[: -len(split_line[-1])] + common_prefix
             else:
                 completed = line
--- a/plugins/ipconsole/ipython_view.py
+++ b/plugins/ipconsole/ipython_view.py
@@ -435,9 +435,9 @@ class ConsoleView(Gtk.TextView):
             if name in ["Courier New", "Courier Mono"]:
                 chosen = name
                 break
-            if name in ["Liberation Sans"]:
-                chosen = name
-                # But prefer a monospace one if possible
+            # if name in ["Liberation Sans"]:
+            #    chosen = name
+            #    # But prefer a monospace one if possible
         if chosen == None:
             chosen = "Mono"
         self.modify_font(Pango.FontDescription(chosen))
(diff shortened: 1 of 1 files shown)


plugins/ipconsole/ipython_view.py
```diff
@@ -402,7 +427,20 @@ def __init__(self):
...
-        self.modify_font(Pango.FontDescription('Mono'))
+        pango_ctx = self.get_pango_context()
+        chosen = None
+        for f in pango_ctx.list_families():
+            name = f.get_name()
+            # These are known to show e.g U+FFFC
+            if name in ["Courier New", "Courier Mono"]:
+                chosen = name
+                break
+            if name in ["Liberation Sans"]:
```
sjohannes, line 438: The problem with this is that Liberation Sans is a quite popular font in the Linux world while the other two aren't. Liberation Sans is a proportional font, which means it will cause misalignments on stuff like the `^` indicator in tracebacks:

```pytb
  File "<string>", line 1
    test!
        ^
SyntaxError: invalid syntax
```

I wonder why Accerciser wants to show [U+FFFC](https://www.compart.com/en/unicode/U+FFFC); I've never seen it before, and we definitely don't need it for Exaile. I think it's best to comment out this `if` block.

genodeftest: Why don't you use `Liberation Mono` instead of `Liberation Sans`? Afaik, it also displays U+FFFC (`￼`).

sjohannes: No, Liberation Mono does not contain that character. You can check it with `fc-list ':charset=fffc' | sort`. As far as I can tell, the most popular font with that character is DejaVu Sans Mono (installation stats: [80% on Debian](https://qa.debian.org/popcon.php?package=fonts-dejavu), [85% on Arch](https://pkgstats.archlinux.de/packages/ttf-dejavu)).

rokm: Should I just drop the commit that introduces this change (so we always keep using `Mono`)? 
Or would we prefer to keep the commit to remain in sync with upstream, and just comment out the `Liberation Sans` if block?

rokm:
> If it's not too inconvenient, could you also contribute your fixes upstream? Maintaining this module has always been a pain, so the more we push upstream the easier it will be to sync when things start breaking again.

Yeah, will do.

plugins/ipconsole/ipython_view.py
sjohannes: Might be useful keeping it commented, so that in the future we know that we intentionally disable that part of the code.

rokm: OK, done.

PR #698, opened 2020-04-19
[RFC/WIP] Optional per-playlist column settings

rokm:
This PR implements an option to detach the playlist's column settings from the global state, and use private playlist-specific settings instead. The switch is added at the top of the columns context menu.

My primary use case is having the default columns (#, Title, Album, Artist, Length) for album-based playlists, and having only Title and Artist for custom-made playlists.

This involves reworking of how the column settings are managed; instead of direct access to the properties in the settings handle, the settings are now managed by the `PlaylistView` object in the GUI/widgets layer. However, to minimize the changes on the playlist save/restore codepath, the settings are also mirrored to the underlying `xl.playlist.Playlist` for storage purposes. I'm not entirely happy with this design, but it seems less intrusive than either trying to serialize/deserialize the column settings at the GUI layer, or trying to push the settings management all the way down to the `xl.playlist.Playlist`.

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -927,6 +927,11 @@ class Playlist:
         'dynamic_mode',
         'current_position',
         'name',
+        # GUI
+        'global_playlist_columns',
+        'playlist_columns',
+        'resizable_cols',
+        'playlist_column_widths',
     ]
     __playlist_format_version = [2, 0]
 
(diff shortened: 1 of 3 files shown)


sjohannes:
Regarding the design, I'm fine with it; seems the most reasonable option. It's probably not ideal having a GUI option leaking to what is part of "core", but it's not a big deal. But I think grouping the settings (as mentioned in my review comment) would make it better.

PR #718, opened 2020-09-05
track._MetadataCacher: Fix error when the time is the same

sjohannes:
If tracks are added to the cacher fast enough, multiple entries can have
the same time (i[2]). This then causes the min((i[2], i) ...) call to end
up comparing the track objects (i[0]), causing a spurious error preventing
the Track Properties dialog from opening:

    TypeError: '<' not supported between instances of 'Track' and 'Track'

After this change, the code will pick any arbitrary item with the lowest
time value.

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -27,6 +27,7 @@
 
 from copy import deepcopy
 import logging
+import operator
 import re
 import time
 from typing import List, Union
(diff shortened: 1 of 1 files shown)


sjohannes:
I've looked into this a bit more and am confident this is safe. The code around this needs a bigger cleanup but this will do for now.

PR #723, opened 2020-10-15
xlgui: collection manager dialog: display unescaped location URIs

### LINKED ISSUE #722: Collection manager does not show spaces and accented characters of added path as expected

1. open Exaile
2. click on "Edit" menu > choose "Collection"
3. add a folder called "Minhas músicas" to your collection
4.

### Expected Behavior

collection manager should show the added path as "/path/to/Minhas músicas"
### Current Behavior

As we can see in the screenshot below, the collection manager replaces the space with "%20" and "ú" with "%C3%BA"
in the name of the added folder 
### Environment

* Operating System and version: Arch Linux, KDE Plasma 5.20 desktop environment

* Exaile Version: I built the source code of master branch downloaded from github

exaile

rokm:
Display unescaped location URIs instead of original, escaped ones. Fixes #722.

### DIFF
--- a/xlgui/collection.py
+++ b/xlgui/collection.py
@@ -45,6 +45,7 @@ class CollectionManagerDialog(Gtk.Dialog):
     __gtype_name__ = 'CollectionManager'
 
     view, model, remove_button, content_area = GtkTemplate.Child.widgets(4)
+    location_column, location_cellrenderer = GtkTemplate.Child.widgets(2)
 
     def __init__(self, parent, collection):
         """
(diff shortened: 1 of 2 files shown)


xlgui/collection.py
```diff
@@ -64,6 +65,17 @@ def __init__(self, parent, collection):
         for location, library in collection.libraries.items():
             self.model.append([location, library.monitored, library.startup_scan])
 
+        # Override the data function for location_column, so that it
+        # displays unespaced location URIs instead of escaped ones
+        def display_unescaped_location(column, cell, model, iter, *data):
```
sjohannes, line 70: Rather than simply unescaping the URL, I think it would be nicer to use [`Gio.File.get_parse_name`](https://lazka.github.io/pgi-docs/#Gio-2.0/classes/File.html#Gio.File.get_parse_name), which I believe displays local paths when possible.

rokm: It is indeed a better choice. Thanks!

sjohannes (approved):

PR #726, opened 2020-10-28
Reminds new users to look at DEPS file.

berrytsakala:
DEPS file is overlooked. Errors due to missing dependencies aren't indicative. Python programs usually has "requirements.txt". Also, added a sample, minimal command line to install, just to get started.
Misc. formatting for readability

### DIFF
--- a/tools/installer/README.md
+++ b/tools/installer/README.md
@@ -13,12 +13,18 @@ Requirements
 
 First, download the python-gtk3-gst-sdk somewhere:
 
-  git clone https://github.com/exaile/python-gtk3-gst-sdk
+    git clone https://github.com/exaile/python-gtk3-gst-sdk
 
 Next install the SDK links by running this from inside this directory:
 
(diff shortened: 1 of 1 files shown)


rokm:
Hm, while I think adding a note about `DEPS` file and giving an example how to install dependencies on (ubuntu/debian) linux is a good idea, I don't think this file is the right place for that. This file contains instructions for building binary installers for Windows and OSX, and it is not something that new users should be looking at. 

[The user guide](https://exaile.readthedocs.io/en/stable/user/index.html) is actually assembled together from the contents of `doc` subfolder, which also pull in `INSTALL` and `DEPS` files, so the guide contains the corresponding sections from those files (do note that the user guide corresponds to the latest release version, which is a bit outdated). 

Perhaps the `INSTALL` file needs to mention the `DEPS` file for users who are reading the file directly (instead of as part of the user guide)?

virtuald:
Updated documentation, feel free to open a new PR if you think that isn't sufficient.

Issue #685, opened 2020-01-27
AutoScrollTreeView scrolls to the top when reordering or adding songs to a playlist


fuhry:
### Steps to Reproduce (for bugs)

1. Create a playlist with enough songs to have 3+ screens worth of scrolling to do.
2. Scroll down a few screens in the playlist
3. Drag to reorder songs, or drag songs from your library to the playlist.

### Expected Behavior

The playlist scroll offset should not change when list items are reordered.

### Current Behavior

The playlist list view (an `AutoScrollTreeView`) scrolls to the top. This appears to be a regression from Exaile 3.x.

### Possible Solution

See this draft PR for a horrific, hacky fix for the issue: https://github.com/exaile/exaile/pull/684

### Environment

Arch Linux, GTK3 3.24.13, MATE desktop environment

sjohannes:
I can't find the root of the problem either; I don't know what's causing it to scroll. What do you think about https://github.com/sjohannes/exaile/commit/9aacc46dca7e3a7b35b5c92d3e423a3c9382ce93 ?

luzip665:
Hello,

I got the same problem even without triggering the auto scroll. Each time I move a track on the bottom of the list to another position the list scrolls to (almost) to top.

A possible solution I found is to save the current vertical offset at the end of on_drag_data_received and re-set after the size-allocation signal.

In on_size_allocate the original value of self.get_parent().get_vadjustment() always is 7.0.

luzip665:
There where similar things while inserting track from the left sidebar to the current list and with the queue.
In the queue it never scrolled to top but lost the right selection of tracks.

luzip665:
Moved the scroll restoring to AutoScrollTreeView (as in PR #684)
Only the selection restoring happens in the playlist itself.

virtuald:
Fixed in #766

PR #729, opened 2020-11-07
widgets/playlist: On dropping tracks, always scroll to them

sjohannes:
This is a workaround for https://github.com/exaile/exaile/issues/685 ,
where DnD causes the view to scroll to the top. I can't figure out what
causes it, so let's go with a hack.

This change causes the view to scroll to the first of the dropped track,
putting it a quarter of the way from the top.

This makes the hidden `gui/scroll_when_appending_tracks` option (default:
False) obsolete because it's now the only supported behavior, so this
commit also removes that option. (The name of the option is misleading; it
only affects DnD.)

Fixes: https://github.com/exaile/exaile/issues/685
Closes: https://github.com/exaile/exaile/pull/684

---

@fuhry: Could you take a look at this and see if it addresses your issues?

### DIFF
--- a/xlgui/widgets/playlist.py
+++ b/xlgui/widgets/playlist.py
@@ -1673,12 +1673,15 @@ class PlaylistView(AutoScrollTreeView, providers.ProviderHandler):
         delete = context.get_selected_action() == Gdk.DragAction.MOVE
         context.finish(True, delete, etime)
 
-        scroll_when_appending_tracks = settings.get_option(
-            'gui/scroll_when_appending_tracks', False
-        )
-
-        if scroll_when_appending_tracks and tracks:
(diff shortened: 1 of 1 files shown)


rokm:
Nice, that scrolling to the top sure is annoying in large playlists...

However, this seems to introduce a new warning/error when a large number of tracks are added (e.g., a custom playlist with >500 tracks):
```
(exaile:41512): Gtk-CRITICAL **: 16:48:50.958: gtk_tree_view_scroll_to_cell: assertion 'tree_view->priv->model != NULL' failed
```

According to a comment couple of lines above, large additions seem to be handled differently... so perhaps this change needs to account of that as well?

virtuald:
Closed in favor of #766

PR #730, opened 2020-11-13
Fix some Python, GTK, GIO, PyGObject deprecations

sjohannes:
Python:

* imp: Obsoleted by importlib, but let's just use normal import.

GTK:

* Entry.set_inner_border: We're supposed to use CSS, but things look fine
  on Adwaita so we'll do without it for now.
* TreeView.set_rules_hint: No replacement, seems to already do nothing.
* Widget.drag_source_set_icon_stock: Use ..._set_icon_name.
* Image.set_from_stock: Use set_from_icon_name.
* Widget.reparent: Use remove+add.

GIO:

* FileInfo.get_modification_time: Use get_modification_date_time.

PyGObject:

* Gtk.Widget.stop_emission: Use stop_emission_by_name,
* Gtk.MessageDialog constructor args: Use GObject property construction.

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -395,10 +395,11 @@ class Track:
 
             # Retrieve file specific metadata
             gloc = Gio.File.new_for_uri(loc)
-            mtime = gloc.query_info(
-                "time::modified", Gio.FileQueryInfoFlags.NONE, None
-            ).get_modification_time()
-            mtime = mtime.tv_sec + (mtime.tv_usec / 100000.0)
+            mtime = (
+                gloc.query_info("time::modified", Gio.FileQueryInfoFlags.NONE, None)
+                .get_modification_date_time()
+                .to_unix()
+            )
 
             if not force and self.__tags.get('__modified', 0) >= mtime:
                 return f
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -395,11 +395,17 @@ class Track:
 
             # Retrieve file specific metadata
             gloc = Gio.File.new_for_uri(loc)
-            mtime = (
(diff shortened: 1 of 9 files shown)


sjohannes:
These are found by running Exaile with `python -X dev ...`. We should probably make that the default when running Exaile from source. There are other warnings but some of them are really tricky or are things we intentionally ignore (e.g. the GtkStatusIcon deprecation warning).

xl/trax/track.py
```diff
@@ -395,10 +395,11 @@ def read_tags(self, force=True, notify_changed=True):
 
             # Retrieve file specific metadata
             gloc = Gio.File.new_for_uri(loc)
-            mtime = gloc.query_info(
-                "time::modified", Gio.FileQueryInfoFlags.NONE, None
-            ).get_modification_time()
-            mtime = mtime.tv_sec + (mtime.tv_usec / 100000.0)
+            mtime = (
+                gloc.query_info("time::modified", Gio.FileQueryInfoFlags.NONE, None)
+                .get_modification_date_time()
```
rokm, line 400: According to docs, `get_modification_date_time` was added in GLib 2.62, which was released a year ago. I think this is going to break LTS distributions that use older versions.

rokm:
Looks reasonable - the only potential problem I see is `get_modification_date_time`, which seems to be a relatively recent addition.

sjohannes:
Thanks, fixed.

By the way, for those not subscribed to the exaile-dev mailing list, I recently made [an automatically-updating page](https://sjohannes.gitlab.io/exaile-deps-table/) to check which OS has which dependency versions. It's just for informational purposes; we don't really have to support everything there if that makes things difficult. It's mostly to know whether we can remove some old workarounds etc.

Issue #735, opened 2020-11-25
Tests fail when locale is not English


X0rg:
### Steps to Reproduce (for bugs)

1. run `make PYTEST=py.test test` when `LANG` is not en_US

### Expected Behavior
Tests should pass, regardless of the system language.

### Current Behavior
Tests fail when a translation is available (tested with German and French), as reported [here](https://aur.archlinux.org/packages/exaile/#comment-777378).
Log available: https://cpaste.org/?7c2f99699ae94732#7BTuhYRncLS5884EdgQmZVD4n75AJR1wdYTHzRNruF3g

### Possible Solution
Override LANG for tests.

### Environment

* Operating System and version: Arch Linux (rolling release) with Python 3.8.6

* Exaile Version: 4.1.0-beta1

PR #736, opened 2020-11-28
Set LC_ALL=C on `make test`

sjohannes:
This prevents having the tests use / rely on translated strings.

Fixes: https://github.com/exaile/exaile/issues/735


rokm (approved):
That's probably the best way to handle this problem.

PR #738, opened 2020-12-11
Replace Travis CI setup with GitHub Actions

sjohannes:
The travis-ci.org closure deadline is coming closer and I think at this point going with GitHub Actions is the best choice. I've considered AppVeyor as well (because our Windows build is already there), but the potential of having everything on GitHub and not having to manage other accounts is just too tempting....

Regarding the code, I've tested it a bit and it seems to work perfectly. The only exception is IRC notification, which sometimes breaks, but whatever.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+    timeout-minutes: 1
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send IRC notification
+        # This uses Bash's /dev/tcp <https://tldp.org/LDP/abs/html/devref1.html>.
+        # The char1 stuff reads one char to make sure that the server is ready
+        # before we start sending things, which hopefully fixes some flakiness
+        # that happens otherwise.
+        run: |
+          exec 3<> /dev/tcp/chat.freenode.net/6667
```
virtuald, line 87: Haha, I never knew this existed. Crazy!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+        # the Dockerfile. Exaile has tests for file permissions that fail in
+        # that setup, so we de-escalate manually here.
+        run: |
+          useradd -MN exaile
+          chown -R exaile .
+          export HOME=/tmp/home
+          su -m exaile -c "make BUILDDIR=/tmp/build test test_compile check-doc"
+
+  deploy:
+    if: startsWith(github.ref, 'refs/tags/')
```
virtuald, line 44: I've been using `if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')`... I don't remember if the push actually matters, I copied it from elsewhere.

sjohannes: It should be ok in this case (this workflow only listens to `push` and `pull_request` events, and the latter [doesn't contain `ref`](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request)), but I'll add it anyway just to be explicit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+        run: sudo apt-get install -y --no-install-recommends python3-gi
+      - id: dist
+        name: Create source dist archive
+        run: |
+          version=${GITHUB_REF#refs/*/}
+          printf '::set-output name=version::%s\n' "${version}"
+          make DIST_VERSION="$version" dist
+          checksum=$(cd dist && sha256sum --tag *)
+          printf '::set-output name=checksum::%s\n' "${checksum}"
+      - name: Create GitHub release
```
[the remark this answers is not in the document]
sjohannes: IIRC it was because it's a bit simpler than the combination of [create-release + upload-release-asset](https://github.com/actions/upload-release-asset#example-workflow---upload-a-release-asset), which has a bunch of nonsense like having to write the filename twice and manually specify the content-type.

Untested:

```
      - id: create_release
        name: Create release
        uses: actions/create-release@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          tag_name: ${{ github.ref }}
          release_name: Exaile ${{ steps.dist.outputs.version }}
          body: |
            ```
            ${{ steps.dist.outputs.checksum }}
            ```
          draft: true
      - name: Upload release asset
        uses: actions/upload-release-asset@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }}
          asset_path: dist/exaile-${{ steps.dist.outputs.version }}.tar.gz
          asset_name: exaile-${{ steps.dist.outputs.version }}.tar.gz
          asset_content_type: application/gzip
```

Which one do you prefer? I don't mind either way.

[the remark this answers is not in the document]
virtuald: Oh, for some reason I had thought that the create-release action had a mechanism for uploading, but I see that it in fact does not. Your original version is definitely simpler then.

Issue #733, opened 2020-11-16
Travis → GitHub deploy fails


sjohannes:
The source deployment stage on Travis [is failing](https://travis-ci.org/github/exaile/exaile/jobs/743915896) and the reason is completely unclear to me.

Looking at the build history, the [alpha1 deploy also failed](https://travis-ci.org/github/exaile/exaile/builds/659951506). So how the hell did it appear in the release? Did I create the tarball manually? I don't remember doing it at all.

rokm:
This looks like the issue with `virtualenv: system_site_packages` being enabled. See https://travis-ci.community/t/python-3-6-and-3-7-with-system-site-packages-enabled-fails-to-activate-on-xenial/1697

rokm:
Do we need system site packages? Or rather, do we need `python-gi` (installed via apt-get, hence the need for system site packages) to create the source distribution?

sjohannes:
No, `make dist` just uses normal Unix stuff (sed, tar, etc.) and Python with no external modules (to run `tools/plugin_tool.py`).

(If you want to trigger a Travis build for testing: `git tag -a whatever && git push origin whatever`. Note that Travis builds take something like 30 minutes now due to #732. To remove the tag: `git push origin :whatever && git tag -d whatever`.)

sjohannes:
I was wrong. `plugin_tool.py` imports Exaile stuff (`xl.xdg` and `xl.version`), which eventually imports GLib.

rokm:
Indeed. Can we rework/isolate version retrieval, so that it won't require GLib?

rokm:
Actually, it seems the only thing `version.py` is importing `xdg.py` for is `exaile_dir`, which is set to `os.environ['EXAILE_DIR']`.

So we could probably have `version.py` use its local copy of `exaile_dir = os.environ['EXAILE_DIR']`. Or, have `exaile_dir` defined in `version.py` or some other place, and have `xdg.py` pull it in from there.

genodeftest:
Can we use something more recent than Xenial?

sjohannes:
Forgot to comment on this when closing it: we've moved from Travis to GitHub Actions, so the original issue is not relevant anymore. Would still be nice not to require PyGObject during build time, though.

PR #739, opened 2020-12-28
Allow tools/plugin_tool.py run without PyGObject/GLib

rokm:
Allows `plugin_tool.py` to import `xl.xdg`, which enables creation of distribution tarball in an environment that does not have `PyGObject` installed.

### DIFF
--- a/tools/plugin_tool.py
+++ b/tools/plugin_tool.py
@@ -53,8 +53,6 @@ if __name__ == '__main__':
     os.environ['EXAILE_DIR'] = exaile_dir
     sys.path.insert(0, exaile_dir)
 
-    import xl.xdg
-    xl.xdg.local_hack = False
     import xl.version
     v = '%s' % (xl.version.__version__.split("-")[0])
 
--- a/xl/version.py
+++ b/xl/version.py
@@ -28,11 +28,15 @@ import logging
 import os
 from typing import Dict
 
-from . import xdg
-
 logger = logging.getLogger(__name__)
 
 __version__ = "devel"
+
+# We need the local hack for OSX bundled apps, so we depend on the main script
+# to set the environment variable correctly instead of trying to infer an
+# absolute path
+# exaile_dir = os.path.split(os.path.dirname(os.path.realpath(__file__)))[0]
+exaile_dir = os.environ['EXAILE_DIR']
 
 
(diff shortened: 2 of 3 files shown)


rokm:
Having fall-back for GLib-dependent directories looks like the least invasive option that allows `plugin_tool.py` to work without PyGObject. 

An alternative would be moving `exaile_dir` and `local_hack` to another module (e.g., `xdg.version`), but as `local_hack` is directly manipulated, we would then also need to propagate the change throughout the whole codebase (i.e., setting up "forward"/"alias" imports in `xl.xdg` for backward compatibility won't work). But we can also go that way, if that's preferable approach.

genodeftest:
See also: #733.

genodeftest:
@rokm: Is this removing the build-time dependency on GLib? Is this something useful for the *BSDs?

rokm:
> @rokm: Is this removing the build-time dependency on GLib? Is this something useful for the *BSDs?

Hmm, it does not fully remove build-time dependency - just enough to make `make dist` work (where `tools/plugin_tool.py` importing `xl.xdg` was causing problems due to GLib dependency in the latter, as discussed in #733). 

Running a build via `make` still does not work, because that runs `tools/generate-completion.py`, which imports `xl.main` and that one eventually comes back to a module with a GLib dependency (the first one is `xl.externals.sigint`).

So this one is meant to address only creation of distribution tarballs in an environment without PyGObject installed.

sjohannes:
Actually, do we even need the `xl.xdg` import in `plugin_tools.py`? I don't see why we're setting `local_hack = False` there.

rokm:
> Actually, do we even need the xl.xdg import in plugin_tools.py? I don't see why we're setting local_hack = False there.

Hmm, you're right. `plugin_tools.py` should not care about `xl.xdg.local_hack` at all. 

If remove that, we can move `exaile_dir` from `xl.xdg` to `xl.version` and add an import in `xl.xdg` to establish backward-compatible alias for the rest of the codebase... Which seems cleaner than the currently-implemented work-around.

rokm:
OK, I've pushed the revised version. Now `plugin_tool.py`  imports only `xl.version˙, and thus completely avoids `xl.xdg` and its `GLib` dependency.

Issue #745, opened 2021-03-15
Default Export Playlist Folder


phalgunlolur:
Every time I try to export a playlist, it takes me to the default folder Program Files (x86)/Exaile/

Is there a way to change this default folder to a specific folder? Or better yet, have the program remember the last used folder?

### Environment

* Operating System and version: Windows 10

* Exaile Version: 4.0.2
GStreamer: 1.16.0
GTK+: 3.24.8
GTK+ theme: Adwaita
Locale: English_United States 1252
Mutagen: 1.40.0
PyGObject: 3.32.1
Python: 2.7.16

phalgunlolur:
I see that @sjohannes  'sjohannes linked a pull request that will close this issue 12 hours ago'

Does that mean, this will happen in a future release? If so, is there something that I can do for a quick fix in the meantime?

genodeftest:
> Does that mean, this will happen in a future release?

Yes, it does. The next release (probably 4.1.1) should solve this issue.

> If so, is there something that I can do for a quick fix in the meantime?

On Windows, this is not very easy. If you are experienced, you can try building the latest git commit for Windows. If not, there is probably not much you can you.

rokm:
> > If so, is there something that I can do for a quick fix in the meantime?
> 
> On Windows, this is not very easy. If you are experienced, you can try building the latest git commit for Windows. If not, there is probably not much you can you.

You can also try using the artifact that was built by AppVeyor from the latest commit, e.g.,  [here](https://ci.appveyor.com/project/ExaileDevelopmentTeam/exaile/builds/38308847/artifacts).

genodeftest:
@rokm : Perfect, I forgot that.

PR #747, opened 2021-03-15
xlgui/widgets/dialogs: Remember last playlist export directory

sjohannes:
Fixes: https://github.com/exaile/exaile/issues/745


rokm (approved):
Looks good. 

Other dialogs seem to keep track of last location via `_last_location` class member, but this seems better as it should persist across program runs.

Issue #750, opened 2021-04-01
Tests fail with "AttributeError: 'NoneType' object has no attribute 'split'"


X0rg:
As reported by [lwilkinson on AUR](https://aur.archlinux.org/packages/exaile/#comment-799705):
> The test "is_valid_track" in exaile-4.1.1/xl/trax/util.py is failing because it uses:
>
> Gio.File.new_for_uri(location)
> 
> which returns 'None'.
> 
> If i change it to:
> 
> Gio.File.new_for_path(location)
>
> the tests work fine. I have not looked into why this is happening, as I am a novice in python, and don't know the reasons why 'new_for_uri' was chosen, however, I can report that the change above works for me.

### Steps to Reproduce (for bugs)

1. `make PYTEST=py.test test`

### Expected Behavior
All tests pass.

### Current Behavior
Some tests fail:
```
-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================================================================================================== short test summary info ===========================================================================================================
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track0] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track1] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track2] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track3] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track4] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track5] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track6] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track7] - AttributeError: 'NoneType' object has no attribute 'split'
FAILED tests/xl/trax/test_util.py::test_is_valid_track_valid[test_track8] - AttributeError: 'NoneType' object has no attribute 'split'
[...]

PR #751, opened 2021-04-02
trax/util.is_valid_track: Change new_for_uri to new_for_commandline_arg

sjohannes:
In Exaile core this is always called with a URI so it works, but the tests
call it with a local file path and we get test failures. (It must have
worked in earlier GIO versions, because our CI doesn't catch this.)

trax.Track uses new_for_commandline_arg for its location argument, so this
should be correct.

Fixes: https://github.com/exaile/exaile/issues/750


X0rg:
Tests pass with this fix, thanks.

PR #753, opened 2021-04-08
Negative start and stop offset per track

luzip665:
This PR implements the ability of negative start and stop offset per track.
So it would be possible to have a different gaps between the tracks and to have all tracks starting and ending with 3 seconds of silence.

Most work is in xl/player/player.py. The play function now adds a GLib-Timeout in case of negative start offset for delaying.

This PR is a draft and implements only the start offset. If you find it usefull and okay  I would implement the stop offset also.


virtuald:
This seems like a reasonable feature to add. If you rebase and add the stop offset I would be happy to merge it.

It would be good to add some unit tests for this, but I suppose that's not really doable at this time.

luzip665:
Rebasing did weird things. So this PR continues in #785

PR #759, opened 2021-07-09
Preserve timestamp upon file installation.

grahamwhiteuk:
Modifies the install command used in the Makefile to add a flag to preserve the timestamps from the released files.  This is a 
Fedora packaging guideline, see https://docs.fedoraproject.org/en-US/packaging-guidelines/#_timestamps


virtuald:
Thanks!

Issue #758, opened 2021-06-18
Unable to sort a playlist with rating value


phalgunlolur:
I just upgraded to the latest exaile and I am no longer able to sort playlists by their rating

### Steps to Reproduce (for bugs)

1. Click on Rating and the sorting no longer works. 
2. Before clicking on Rating, sorting works well with other fields. Sorting doesn't work on other fields after clicking on Rating

### Expected Behavior
Be able to sort according to rating

### Current Behavior
Sorting stops working after clicking on the Rating tab

### Environment

* Operating System and version: Windows 10

* Exaile Version: 4.1.1

GStreamer: 1.18.4
GTK+: 3.24.27
GTK+ theme: Adwaita
Locale: English_United States 1252
Mutagen: 1.40.0
PyGObject: 3.38.0
Python: 3.8.8

luzip665:
Hi,

this happens in case there is a track without rating in the list.

Possible bugfix: https://github.com/exaile/exaile/pull/762

rokm:
Fixed by #762.

PR #762, opened 2021-07-11
Bugfix Unable to sort a playlist with rating value #758

luzip665:
Hello Johannes,

this problem happens in case that there is a track with none rating.
In the case the get_tag_sort function returns None.
Instead the track function get_rating should be used, shouldn't it?

### DIFF
--- a/xl/trax/track.py
+++ b/xl/trax/track.py
@@ -654,6 +654,8 @@ class Track:
         elif tag == '__basename':
             # TODO: Check if unicode() is required
             value = self.get_basename()
+        elif tag == '__rating':
+            value = self.get_rating()
         else:
             value = self.__tags.get(tag)
 


rokm:
> this problem happens in case that there is a track with none rating.
> In the case the get_tag_sort function returns None.
> Instead the track function get_rating should be used, shouldn't it?

Indeed, the problem is that if no rating is set, `self.__tags.get(tag)` (with `tag='__rating') returns `None`, and [subsequent code block converts it to "unknown" string](https://github.com/exaile/exaile/blob/96e4db1095ca055245e02cfdbcf69dedf51270e8/xl/trax/track.py#L660-L661). This causes errors when valid ratings (floats) are compared to unknown ratings (strings).

So a possible fix would also be to catch `None` ratings and explicitly force them to `0`. 

Or, as the proposed fix does, use `get_rating()`, which already returns 0 for invalid ratings. (The slight difference is that `self.__tags.get('__rating')` returns floats in range 0-100, whereas `get_rating()` returns unscaled integers in range 0-5, but for purposes of sorting, that should not be important).

xl/trax/track.py
```diff
@@ -654,6 +654,9 @@ def get_tag_sort(
         elif tag == '__basename':
             # TODO: Check if unicode() is required
             value = self.get_basename()
+        elif tag == '__rating':
+            # TODO: Check if unicode() is required
```
rokm, line 658: Can you remove this comment? It was probably copied from above case, but it is not applicable in the case of ratings (as we're dealing with integers).

rokm (approved):

rokm:
Thanks @luzip665 !

sjohannes:
Interesting; I suppose this used to work because comparing None with a number doesn't raise an exception on Python 2.

This seems like a pretty significant bugfix that should be in a release version. I'll get to releasing 4.1.2 one of these days....

Issue #764, opened 2021-08-09
Traceback from on_location_changed when scanning files that have been removed


grahamwhiteuk:
This issue comes from [BZ #1990693](https://bugzilla.redhat.com/show_bug.cgi?id=1990693) on the Red Hat Bugzilla system having been raised for a Fedora system running Exaile 4.1.1.  I've seen the issue come up for several users via the Fedora automatic bug reporting tool (ABRT) at https://retrace.fedoraproject.org/faf/reports/220274/

Please note: I'm the Fedora packager and not the original source of the bug report.

### Steps to Reproduce (for bugs)

1. Install Exaile 4.1.1 on Fedora 33 or Fedora 34
2. Run Exaile
3. Scan for new files in the collection
4. Having something deleted during scanning
5. Traceback happens

User reports the following:

> It happened when trying to access a '.tmp' file that had already been deleted by another process.
> 
> This happened when working with Exaile 4.1.1 in a directory with music files, which were parallely edited in Ardour 6.8.0.
> 
> Eventually I have switched the directory (... location changed ... from the error message) and tried listing its files.
> 
> Apparently Ardour would create '.tmp' files next to its project files, and delete them when not needed anymore.
> 
> This may have happened after Exaile picked up the file through its INotify monitor, but after it was already deleted.

### Expected Behavior

Exaile should handle files going missing during scanning

### Current Behavior

Exaile bombs out with traceback:

```
collection.py:470:on_location_changed:gi.repository.GLib.GError: g-io-error-quark: Fehler beim Holen der Informationen für Datei »/home/yala/Musik/Ardour/Power Hour/Power Hour.tmp«: Datei oder Verzeichnis nicht gefunden (1)

Traceback (most recent call last):
  File "/usr/share/exaile/xl/collection.py", line 470, in on_location_changed
    fileinfo = gfile.query_info(
gi.repository.GLib.GError: g-io-error-quark: Fehler beim Holen der Informationen für Datei »/home/yala/Musik/Ardour/Power Hour/Power Hour.tmp«: Datei oder Verzeichnis nicht gefunden (1)

Local variables in innermost frame:
self: <collection.LibraryMonitor object at 0x7f30a19a12c0 (xl+collection+LibraryMonitor at 0x555a4d046840)>
monitor: <__gi__.GInotifyFileMonitor object at 0x7f309863e0c0 (GInotifyFileMonitor at 0x7f306002bd00)>
other_gfile: None
[...]

PR #765, opened 2021-09-06
xl/collection: mitigate errors when files are removed during scan

rokm:
Attempt to mitigate errors that arise when files in monitored collection are removed shortly after being created (#764).


sjohannes (approved):
Looks good to me.

PR #766, opened 2021-09-11
Bugfix  AutoScrollTreeView scrolls to the top when reordering or addi…

luzip665:
…ng songs to a playlist #685


virtuald:
Fixes the issue for me, thanks for looking into this! Sorry for the delayed review.

PR #768, opened 2021-09-22
inhibitsuspend: Support sway

ryneeverett:
I took the opportunity to pull out dbus-specific functionality into
DbusSuspendAdapter. Once Wayland's idle inhibition protocol is
stabilized it should be possible to adapt most wayland compositors with
a comparable WaylandAdapter. See
<https://wayland.app/protocols/idle-inhibit-unstable-v1>.


virtuald:
LGTM, thanks.

PR #772, opened 2021-12-14
Quickbuttons Plugin: Replace Gtk.Toolbar with Gtk.Box

luzip665:


luzip665:
As seen on XFCE with Bluebird Theme the GTK.Toolbar is rendered with a non transparent background.
For this it is replaced by Gtk.Box, which seems to more save in the future (as of GTK4)

virtuald:
Thanks!

Issue #775, opened 2021-12-15
Add source selection to quickbuttons plugin


virtuald:
Great. I was wondering if it possible to add a shortcut to select a source for the output audio and the preview device directly too. I'll check out the equalizer button later tonight! Thanks for adding it!

_Originally posted by @phalgunlolur in https://github.com/exaile/exaile/issues/746#issuecomment-994046065_

luzip665:
@phalgunlolur
Hi!
You mean the playback setting for Audio Sink or Audio Device?

phalgunlolur:
Audio Device. I usually leave Audio Sink to be automatic, without it being an issue for me.

Picking the main audio device and the preview device from the main window would be super useful. Especially, when I'm dealing with multiple sound cards

luzip665:
Can you check
[quickbuttons_0.4.98.tar.gz](https://github.com/exaile/exaile/files/7732632/quickbuttons_0.4.98.tar.gz) ?
 
Right now working for the main device.

phalgunlolur:
I'm unable to install it as I get the error - 'Plugin archive contains an unsafe path'
![image](https://user-images.githubusercontent.com/17207822/146640295-556fc424-4fef-4e3a-ad74-68a4bffef090.png)

luzip665:
Ah. You need to rename the archive to quickbuttons.tar.gz.

luzip665:
[quickbuttons.0.4.99.tar.gz](https://github.com/exaile/exaile/files/7750116/quickbuttons.0.4.99.tar.gz)

Now with both devices. The device list is a menu in order to update it if a device is added.

phalgunlolur:
I've renamed the file but I cannot install the update as I already have the plugin installed with the same name (older version). How do I remove it to do a fresh install?

![image](https://user-images.githubusercontent.com/17207822/147304673-ea923f9d-6637-4927-8a63-7617aa04f7ea.png)

luzip665:
Did a plugin update ever work?

For a fresh installation you need to find the exaile-plugin directory on your system. 
You use Windows, right? Could be something like `%USERPROFILE%\AppData\Local\exaile` .

If found in there is a folder named quickbuttons. This you need to remove.

phalgunlolur:
I've never installed a plugin update until now. I removed the plugin folder in the local folder. I was initially looking for it in the main plugins folder in Program Files. This works really well for me now.

[...]

PR #776, opened 2021-12-17
Quickbuttons: Add source selection

luzip665:
See https://github.com/exaile/exaile/issues/775

### DIFF
--- a/plugins/quickbuttons/__init__.py
+++ b/plugins/quickbuttons/__init__.py
@@ -74,7 +74,7 @@ class QuickButtons:
             "type": "equalizer",
             "label": _("EQ"),
             "tooltip": _("Equalizer"),
-            "depends_on": "equalizer"
+            "depends_on": "equalizer",
         },
         "player/audiosink_device": {
             "show_button": "quickbuttons/btn_audio_device",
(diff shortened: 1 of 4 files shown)


plugins/quickbuttons/__init__.py
```diff
@@ -177,7 +219,20 @@ def _add_button(self, setting: str) -> None:
...
+            tbs.connect("clicked", self._on_equalizer_press)
+
+        elif self._options[setting]["type"] == "audio_device_selection":
+            # @see plugins/previewdevice/previewprefs.py:65
+            from xl.player.gst.sink import get_devices
+
+            tbs = Gtk.ComboBoxText()
+            val = self._options[setting]["value"]
+            for device_name, device_id, _unused in list(get_devices()):
+                tbs.append(device_id, device_name)
```
virtuald, line 231: This isn't going to account for devices being added/removed. Maybe add a refresh option (right click menu?)

Also, the combo box doesn't render properly for me (I have to scroll to see the items), but I expect that's just a GTK quirk?

luzip665: You are right. I think best solution would be to connect to the combobox notify::popup-shown signal to reload the list of devices just in time.

luzip665: The rendering seems to relate to the position of the box. If it is not on the lower end of the screen it works better..

luzip665:
Now there are both devices listed
Both selections are updated in time

virtuald:
I think the translation strings would be best in a separate commit. Doesn't matter to me whether it's this PR or another.

luzip665:
I'll add the translations in a separate PR.

Issue #777, opened 2021-12-21
Plugin Previewdevice is not shown after being activated after deactivation


luzip665:
### Steps to Reproduce (for bugs)
1. Enable Plugin Previewdevice
2. Previewdevice is shown in the upper right corner
2. Deactivate the plugin
3. Previewdevice is not shown in the upper right corner
4. Activate the plugin
5. Previewdevice is not shown in the upper right corner

### Possible Solution
The option plugin/previewdevice/shown is set to false in case of deactivation but never set to true

### Environment
* Operating System and version: Ubuntu 20.04
* Exaile Version: Current master

PR #778, opened 2021-12-21
Plugin Previewdevice cannot be activated after deactivation #777

luzip665:
see https://github.com/exaile/exaile/issues/777

### DIFF
--- a/plugins/helloworld/__init__.py
+++ b/plugins/helloworld/__init__.py
@@ -76,12 +76,20 @@ from . import testlib
 
 
 class HelloWorld:
+    def on_plugin_installed(self):
+        '''Optional function'''
+        print('Install Hello World')
+
     def enable(self, exaile):
         print("Hello, world!")
         testlib.sucess()
 
     def disable(self, exaile):
         print("Goodbye. :(")
+
+    def on_plugin_uninstalled(self):
+        '''Optional function'''
+        print('Uninstall Hello World')
 
     def teardown(self, exaile):
         '''Optional function'''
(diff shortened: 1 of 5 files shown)


virtuald:
This doesn't seem quite right. Won't this always show the GUI now when Exaile restarts? I thought there was a preferences page option to show/hide it...

luzip665:
You are right. I'm able to hide the preview device in the view menu.
In this case I would suggest, that the option `plugin/previewdevice/shown` keeps unhandled in case of deactivating the plugin.

So the option is only set in _on_view

virtuald:
This PR doesn't fix #777 for me. 

I think the correct solution would be to introduce an `installed` callback for plugins that is called whenever a plugin goes from 'not enabled' to 'enabled'. Currently the 'enabled' callback is called every time that Exaile starts, so the semantics are slightly different than what is required for this.

If an 'installed' callback existed, then the 'shown' option could always be set when that function is called.

luzip665:
You are right, `enable` is called every time. Many plugins use this just for retrieving the exaile instance. For this may be it's not necessary to have a special `installed` hook, even it would be a clearer naming.
Mostly, I think, the content of `enabled` can be moved to `on_exaile_loaded` or `on_gui_loaded`. But this may be a huge work.

Imo, a possible solution for re-enabling could be to set `shown` to true in `disable`. Then it will be shown on the next real enabling via the plugin list.

luzip665:
Finally I added `install` and `uninstall` hooks and enhanced the hello world plugin.
But I would recommend to rename the called functions to get more clarity on it.

virtuald:
Works for me, thanks!

While `enable` is called whenever Exaile starts (and at install time), it seems that 'disable' is actually only called when the plugin is uninstalled. This is a bit confusing, but it's old enough that I don't think it would be wise to change it. I modified the commit to only keep the installation callback.

Issue #779, opened 2021-12-21
Window title does not change after resuming from pause


luzip665:
### Steps to Reproduce (for bugs)
1. Play a track
2. Pause via button
3. Seek or not to the right end of the track
4. Resume via button

### Current Behavior
After changing to the next track the window title remains on the old one.
May relate to #572 

### Expected Behavior
After changing to the next track the window title is updated to this track

### Environment
* Operating System and version: Ubuntu 20.04
* Exaile Version: 4.1.1 and current master

Issue #571, opened 2018-11-14
Playback buttons sometimes move around when hovered


sjohannes:
1. (Re)start Exaile.
2. Hover on Stop button.

The buttons will shake for a bit.

Need to check if putting the buttons in a Box with homogeneous=true fixes it.

Tested on Windows.

sjohannes:
Fixed by #780.

PR #780, opened 2021-12-21
Window title does not change after resuming from pause #779

luzip665:
see #779

### DIFF
--- a/plugins/previewdevice/__init__.py
+++ b/plugins/previewdevice/__init__.py
@@ -62,7 +62,6 @@ class SecondaryOutputPlugin:
     def on_gui_loaded(self):
 
         self.hooked = False
-        self.resuming = False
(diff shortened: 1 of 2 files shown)


luzip665:
As far as I can see, `on_playback_start` is never called after resume from pause, so it is not necessary to check resume.

sjohannes:
With this, can the `resuming` variable be removed?

Similar code is also present in the previewdevice plugin. Should this be done there as well?

luzip665:
I removed resume in main and previewdevice
Also I added lines to set icon size on playpause and stop initially to prevent later moving (may affect #571)

virtuald:
Works for me. Looks like all of that resuming stuff is pretty old, so probably just vestigial. Thanks!

Issue #662, opened 2019-10-28
Add a way to easily add current played song to playlist


brunetton:
Hi,
I'm using Exaile for a couple of hours and I'm creating playlists. First impression is very good ! Very pluggable, simple and clean. I love it ! 
But I've the global feeling that it's quite or impossible to "find" the track I'm listening to; to add it to playlist(s):

- as soon as a track is started to being played, it dissapears from Queue
- if I didn't started the play action from an existing playlist, `view -> show playing track` option is useless, nothing happens

### Current Behavior
- add some tracks in queue from collection
- double-click on a track to play it
- try to find this track to add to a playlist

### Possible Solution
I see two simple solutions for this problem:

1. keep playing track in the queue (like the most part of players I've tried) 
  - and remove it only when the track finished
  - or display it in gray at the top of queue
2. add a way to drag & drop current track to playlist from the top part of the app (where the album cover is displayed, and the track name is displayed in big)

I think the two could be implemented:
- keep playing track in the queue is a solution to always find the current track by invoking `view -> show playing track`: this could simply focus/show queue
- being able to drag & drop current track to playlist from the top part of the app could be a simple way to always find a way to drag & drop current track, always from the same location (whatever the state of playlists panels is)

### Environment
* Operating System and version:
Arch Linux

* Exaile Version: 
4.0.0

virtuald:
Hi @brunetton, we're happy you find Exaile useful!

There is currently already an option to ensure that tracks don't disappear from the queue:

* Edit -> Preferences
* Under "Playback", there is an option "Remove track from queue upon playback", make sure it's unchecked.

Hope that helps!

Issue #781, opened 2021-12-21
QueuePage improvements


luzip665:
I wonder if it is impossible to 

* see the play/pause icon on the queue tab
* find the currently played song in the queue tab

So I added both things to the QueuePage. I'm quite sure, this can be done better. The icons can be moved to the PlaylistPageBase or so.

Additionally I would like to change or extend the behaviour of the option queue/remove_item_when_played.
Imo it should be possible to leaf the currently playing track in the queue and remove just after playing.
Doing so it would also be possible to find the currently playing song also if it is in the queue. 

Does someone agree?

May also affect #662

PR #782, opened 2021-12-21
Queue improvements

luzip665:
see #781


luzip665:
I added following functions:
* Possibility to show current track even in QueuePage
* Moving icons and on_playback_state_change to PlaylistPageBase
* Set queue playlist as active playlist on autoadvance
* Added preference "queue/remove_item_after_played" to keep playing track in queue
* Fixes: #662

virtuald:
With "Added preference "queue/remove_item_after_played" to keep playing track in queue" ... it seems that the "Remove after playback" option always removes after playback (and never at playback), regardless if it's enabled?

The fix for #662 feels like it should be a separate PR? The fix seems to work, and I would merge it if it were separate... :)

luzip665:
I'm not sure if I understand correctly.

With preference "Remove track from queue upon playback" the track is removed at the beginning of playback, as before.
With preference "Remove track from queue upon playback" and "Remove after playback" the track is removed after forwarding to the next track in queue.

Do you see different behaviour?

Anyway, I will separate the fix into an own PR

virtuald:
I saw different behavior, yes.

PR #785, opened 2021-12-23
Start Stop offset

luzip665:
This PR is a replacement for #753

### DIFF
--- a/xl/player/player.py
+++ b/xl/player/player.py
@@ -60,6 +60,7 @@ class ExailePlayer:
         self._playtime_stamp = None
 
         self._delay_id = None
+        self._delayed_stop_id = None
         self._stop_id = None
         self._engine = None
 
@@ -67,6 +68,11 @@ class ExailePlayer:
         self._auto_advance = True
         self._gapless_enabled = True
         self.__volume = 1.0
+        self._delayed_until = 0
+        self._paused_until = 0
+        self._paused_start = 0
+        self._track = None
+        self._current_play_args = None
 
         options = {
             '%s/auto_advance_delay' % name: '_auto_advance_delay',
@@ -167,6 +173,8 @@ class ExailePlayer:
 
     @property
     def current(self):
+        if self._track:
+            return self._track
         return self._engine.get_current_track()
 
     def play(self, track, start_at=None, paused=False):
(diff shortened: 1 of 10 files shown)


luzip665:
@virtuald 
I would like to implement unit tests for this feature. But, I have no idea, how.
I think I can write a test for the tags, so that it's possible to check whether it's possible to use negative values for the start.

But how can the player be tested?

virtuald:
I'm not really sure how to write tests that use gstreamer, unfortunately. It's fine if we don't have tests, just would be nice if the functionality could be somehow tested.

virtuald:
Also, I can't try this at the moment, so I'll review in a few days when I can.

xl/player/player.py
```diff
@@ -504,8 +571,31 @@ def engine_autoadvance_get_next_track(self, gapless=False):
             if self._auto_advance_delay != 0 or not self._gapless_enabled:
```
virtuald, line 571: Looks like if gapless playback is enabled, then the delayed stop won't work?

xl/player/player.py
```diff
@@ -504,8 +571,31 @@ def engine_autoadvance_get_next_track(self, gapless=False):
...
+        track = self._track
+        stop_at = track.get_tag_raw('__stopoffset') or 0
+        track_len = track.get_tag_raw('__length') or 0
+
+        if stop_at > track_len:
+            logger.debug('entering delayed stop')
+            diff = stop_at - track_len
+            self._paused_start = time.time()
+            self._paused_until = diff + self._paused_start
+            self._delayed_stop_id = GLib.timeout_add_seconds(1, self._wait_for_end)
```
virtuald, line 583: Why is this a periodic callback? Why not just delay for the difference?

xl/player/player.py
```diff
@@ -504,8 +571,31 @@ def engine_autoadvance_get_next_track(self, gapless=False):
             if self._auto_advance_delay != 0 or not self._gapless_enabled:
                 return
 
+        track = self._track
+        stop_at = track.get_tag_raw('__stopoffset') or 0
+        track_len = track.get_tag_raw('__length') or 0
+
+        if stop_at > track_len:
```
virtuald, line 578: This doesn't seem like the right place to implement this. As the documentation for the function says, "the (next) track may or may not actually get played."

xl/player/player.py
```diff
@@ -187,15 +195,48 @@ def play(self, track, start_at=None, paused=False):
...
+            if start_at < 0:
+                logger.debug('entering delayed start')
+                self._delayed_until = -1 * start_at + time.time()
+                self.engine_notify_track_start(track)
+                self._delay_id = GLib.timeout_add_seconds(1, self._play_engine)
+            else:
+                self._play_engine()
+
+            if self._current_play_args[2]:
                 event.log_event('playback_player_pause', self, track)
```
virtuald, line 220: This feels like it should happen in _play_engine instead? I guess I don't really remember what this is supposed to do.

xl/player/player.py
```diff
@@ -187,15 +195,48 @@ def play(self, track, start_at=None, paused=False):
...
+            )
+
+            self._delayed_until = 0
+            self._cancel_delayed_stop()
+            self._track = track
+            start_at = self._current_play_args[1]
+
+            if start_at < 0:
+                logger.debug('entering delayed start')
+                self._delayed_until = -1 * start_at + time.time()
```
virtuald, line 213: If you just had the timeout call you at the correct time, this wouldn't be needed?

xl/player/player.py
```diff
@@ -187,15 +195,48 @@ def play(self, track, start_at=None, paused=False):
         if track is None:
             self.stop()
         else:
+            self.stop()
```
virtuald, line 198: Why is this here? I think this would break gapless playback, and it's going to emit a stop event too? The player isn't actually stopping after all.

virtuald (changes requested):

xl/player/player.py
luzip665: So it is. At least I didn't find a better place.
`engine_autoadvance_get_next_track` is called by `gst/engine.py::_autoadvance_track`, which is called when the current stream ends. In this moment the current track is still the last one (that with the stop offset in future).
I tried to move this to `engine_autoadvance_notify_next` but there the offset information is allready lost.

xl/player/player.py
luzip665: This is used in case that there is a global auto advance delay. Then the player is paused until the delay times out.
I'll try to move.

xl/player/player.py
luzip665: Was just for printing a pretty countdown in the log

luzip665:
Finally I end up in moving everything to AudioStream.
Starting there it works with gapless.

@virtuald Can you have a lookg for the AudioStream class?

Next I'm working on the properties window and rebasing.

virtuald:
AudioStream seems like a better place for it (ha, I forgot it even existed!), but I don't have time at the moment to properly review. Will try to do so in the near future.

Issue #788, opened 2021-12-24
Plugin update and remove is difficult


luzip665:
If I try to update or remove custom installed plugins, there are some difficulties

### Steps for reproduce
1. Install a custom plugin via Settings > Plugins > Install > Add Plugin File
2. Install the same plugin (in a newer version) again

### Current Behavior
A Message with 'A plugin with the name ... is already installed.' comes up

### Expected Behavior
The new plugin version is installed

### Steps for reproduce
1. Install a custom plugin via Settings > Plugins > Install > Add Plugin File

### Current Behavior
There is no button to remove the custom plugin

### Expected Behavior
There should be a button to remove the custom plugin. May be in the plugin list next to the refresh plugin.
The button is only needed for custom plugins. Shipped plugins must not be deletable.

phalgunlolur:
Great. I'm glad this got added!

Issue #591, opened 2019-02-04
Rename xdg.local_hack to something less misleading


sjohannes:
`xdg.local_hack` checks whether the `data` subdir exists in `$EXAILE_DIR`. There are three situations where this is the case:

* Running Exaile from source directory.
* Running Exaile from the official Windows package.
* Running Exaile from the official Mac package.

The name may lead people (such as myself) to think that it only checks for the first case, causing confusion like in #588.

I want to rename it to something else. The variable essentially checks whether Exaile is installed in an FHS-compliant location, so I'm thinking of changing it to `is_fhs_install` and flipping the logic and usage.

luzip665:
Done with #792

PR #792, opened 2021-12-30
Plugin Installation and Removing

luzip665:
Fixes: https://github.com/exaile/exaile/issues/788

### DIFF
--- a/xl/plugins.py
+++ b/xl/plugins.py
@@ -43,18 +43,17 @@ class InvalidPluginError(Exception):
         return str(self.args[0])
 
 
+class PluginExistsError(Exception):
+    def __str__(self):
+        return str(self.args[0])
+
+
 class PluginsManager:
     def __init__(self, exaile, load=True):
-        self.plugindirs = [os.path.join(p, 'plugins') for p in xdg.get_data_dirs()]
-        if xdg.local_hack:
-            self.plugindirs.insert(1, os.path.join(xdg.exaile_dir, 'plugins'))
 
-        try:
-            os.makedirs(self.plugindirs[0])
-        except Exception:
-            pass
+        self.user_installed_plugindir = xdg.get_user_plugin_dir()
+        """Dir for user installed plugins"""
 
-        self.plugindirs = [x for x in self.plugindirs if os.path.exists(x)]
         self.loaded_plugins = {}
 
         self.exaile = exaile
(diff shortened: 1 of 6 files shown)


luzip665:
With this it is possible to override plugins with new ones. Includes the override of built-in plugins.
Also, user installed plugins can be removed from the plugins list.

xl/plugins.py
```diff
@@ -55,6 +60,12 @@ def __init__(self, exaile, load=True):
             pass
 
         self.plugindirs = [x for x in self.plugindirs if os.path.exists(x)]
+
+        self.user_installed_plugindir = self.plugindirs[0]
```
virtuald, line 64: I don't think using list indices is a good idea here. Since `get_data_dirs` is defined in xdg.py, it seems like it would be better to have a way to return the user data directory and the system data directories instead?

xl/plugins.py
```diff
@@ -55,6 +60,12 @@ def __init__(self, exaile, load=True):
...
+        """Dir for user installed plugins"""
+        self.builtin_plugindir = self.plugindirs[1]
```
virtuald, line 66: Sometimes there are two or more system data directories.

virtuald (changes requested):

luzip665:
Okay. 
As far as I can see there are many directories in `get_data_dirs` but only 2 of them does really relate to plugins?
So I will try to reduce the directories and create methods to retrieve the plugin directories directly.

Also I will rename xdg.local_hack (as in #591)

luzip665:
Moved detection of plugindirs to xdg. 
user_plugin_dir is still taken from plugindirs[0], which worked for years, I think
system_plugin_dir ist taken from plugindirs[1]. With this I am quite unsure, but I think it is allways correct. 
In my tests it resolves to `/path/to/exaile/plugins` in case of running from source. This seems also to happen if running on windows or osx.
Running from linux installation system_plugin_dir is always `/usr/share/exaile/plugins`

xl/plugins.py
luzip665: I added a function

luzip665:
Now the user plugin dir is taken from xdg.get_user_plugin_dir.
There is no special builtin plugin dir. Instead as earlier all plugin data dirs are searched.

virtuald (approved):
LGTM :shipit:

Issue #799, opened 2022-01-13
Windows build broken due to feedparser & setuptools incompatibility


sjohannes:
[Build log.](https://ci.appveyor.com/project/ExaileDevelopmentTeam/exaile/builds/42190852)

The culprit is [feedparser 5.2.1 using setuptools' `use_2to3` feature](https://github.com/kurtmckee/feedparser/blob/5.2.1/setup.py#L6), which has been [removed in setuptools 58](https://setuptools.pypa.io/en/latest/history.html#v58-0-0).

Suggested solution:

* Test whether the *podcasts* plugin can work with feedparser 6.
* If it works, then just upgrade the feedparser version in `tools/installer/requirements.txt`.
* If it doesn't work, make it work with both feedparser 5 and 6. This is necessary because Ubuntu is still on feedparser 5 while Fedora is already on 6.

The quick workaround, if we just want the Windows build to work again, is to revert the [Windows SDK update](https://github.com/exaile/exaile/commit/904ea460361617fd48aa66c10b4bc7a513ad3b0f).

luzip665:
Test with feedparser 6.0.8 on Ubuntu 20.04 seem to work.

luzip665:
Tested on Win10 with feed 'https://www.bremeneins.de/feed/podcast/chronik-114.xml'

Works in general, but has an issue with the ssl. May relate to https://stackoverflow.com/questions/35569042/ssl-certificate-verify-failed-with-python3

Issue #51, opened 2015-03-02
"Change language" option in the menu


asereze:
I'm translating Exaile in Sardinian (at 70.1% now), a minority language spoken by more than 1.2 million people, but using the program I've not found a "change language" option, and it opens automatically in English+Italian on my PC. Could that be fixed?
Thanks for your time and work.

virtuald:
I believe that the program uses the current locale set by the OS to open the program.

l2212:
Yes, I think that too (I could not verify because I still have not finished the translation so it could not be in the program). The problem is that, for example, Windows 7 doesn't seem to have Sardinian in the possibilities for setting (while Windows 8 does, at least, I don't know about Osx or Linux-based OS).

virtuald:
Hm. Well, I suppose one possible solution is we could write the preferred locale to settings, and at startup when we do locale detection we could see if the user has a preferred locale, and use that if set.

Definitely would not be able to dynamically change the locale, however. Would require a restart.

l2212:
Yeah, that could work fine. After all, I don't think the need of a restart to apply the change would be a great problem for users, the only necessary thing would maybe be a disclaimer about that.

genodeftest:
I don't know wheter this option exists under Windows, but under Linux/Unix you could run
`LC_ALL=sc exaile` to run exaile with sardinian locale. The `LC_ALL=sc` statement sets the environment variable for libc's locale to the language code for sardinian.

I expected this python code in exaile.py to do the trick (it should afaik), but it didn't:
```
import locale

locale.setlocale(locale.LC_ALL, 'ru_RU')
```
Maybe you can set LANG or LC_ALL in exaile.bat.

sjohannes:
According to [1], setting locales in Windows is rather broken, and you 
have to actually set the LC_ALL environment variable.

[1] 
https://stackoverflow.com/questions/3678174/python-gettext-doesnt-load-translations-on-windows

I tried adding

   set LC_ALL=ru_RU

to exaile.bat, and it partially works. However, there is still a 
significant number of strings that don't get translated. (And it's not 
because of incomplete translation; Exaile's Russian translation almost 
100% complete.) I can't figure out what makes these strings different.

PR #800, opened 2022-01-18
Translation on windows fails

luzip665:
Bug: On german Windows exaile speaks english.

I added setting os.environ['LANG'] which helps mostly.

Only the main menu isn't translated yet.

Additionally, because it depends on windows, it solves https://github.com/exaile/exaile/issues/799

fixes: https://github.com/exaile/exaile/issues/799
fixes: https://github.com/exaile/exaile/issues/51

### DIFF
--- a/xlgui/guiutil.py
+++ b/xlgui/guiutil.py
@@ -27,6 +27,7 @@
 import contextlib
 import logging
 import os.path
+import sys
 
 from gi.repository import Gio
 from gi.repository import Gdk
@@ -36,6 +37,7 @@ from gi.repository import Gtk
 from gi.repository import Pango
 
 from xl import settings, xdg
+from xl.nls import gettext as _
 
 # Required for xlgui.get_controller()
 import xlgui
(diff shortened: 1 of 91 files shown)


luzip665:
On linux the manual setting of os.environ['LANGUAGE'] works.
On Win many strings are translated but some not. May depend if the string come from .ui-files?

luzip665:
Seems to be a more general problem
https://gitlab.gnome.org/GNOME/pygobject/-/issues/422

luzip665:
Still not translated:
* Developer tools
* Device Manager
* Equalizer
* Shortcuts
* Collection Manager
* Cover Manager

luzip665:
Everything gets translated now.

May be there are better ways to implement the translation of the ui-xml especially for GtkTemplate. 
Hints would be appreciated.

xlgui/guiutil.py
```diff
@@ -730,4 +732,39 @@ def without_model(tv):
             tv.scroll_to_cell(scroll_to, None, True, 0, 0)
 
 
+def get_builder(uifile: str) -> Gtk.Builder:
+    """
+    Loads Gtk.Builder with uifile.
+    In case of platform win32 each translatable node in the ui is translated here
```
sjohannes, line 738: This is due to the bug you mentioned (https://gitlab.gnome.org/GNOME/pygobject/-/issues/422), right? Could you add the link, so we remember why this hack is needed?

xl/common.py
```diff
@@ -299,6 +299,27 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def platform_is_win() -> bool:
```
sjohannes, line 302: I don't think we need these functions. On Windows `sys.platform` is always `win32` so we can just directly test for that. (And the other two functions are unused.)

xlgui/guiutil.py
```diff
@@ -730,4 +732,39 @@ def without_model(tv):
...
+
+    :param uifile: Path to uifile
+    """
+    builder = Gtk.Builder()
+
+    if True or common.platform_is_win():
```
sjohannes, line 744: This is always `True`.

xl/common.py
luzip665: I would like to replace occurences of `sys.platform == 'darwin'` and `sys.platform == 'linux'.
But may be it is off topic.

xlgui/guiutil.py
luzip665: oops

xlgui/guiutil.py
luzip665: Added

Issue #801, opened 2022-01-21
Volume scale broken after changing the output device


luzip665:
Right after starting exaile I can scale the volume with the volume scaler and see the change on pulseaudio.

After I change the output device via Settings > Playback > Audio Device it is broken. When I scale in exaile nothing happens to pulseaudio.

### Steps to Reproduce (for bugs)
1. Change output device in Settings > Playback > Audio Device
2. Scale volumne

### Environment
* Operating System and version: Ubuntu 21.10
* Exaile Version: Current head

virtuald:
This definitely didn't used to be the case. Can you check an older version of exaile (4?) to see if this behavior is present?

luzip665:
Latest version I got working is 4.1.0. In this version it is the same.

Volume control works after click on stop and start again. It fails only if I change the device during playback.

virtuald:
If a prior version wasn't broken, then it seems to me better to use `git bisect` to pinpoint what change broke it, and address the issue that way.

luzip665:
You are right, but actually I got only 4.1.0 running. There it is the same.
I hardly try to run a 4.0.x

luzip665:
Updated summary (I hope this is more clear):

It works as expected if I
* play a song
* stop it
* change the device
* start playback

It does not work if I
* play the song
* change the device
* In this case the playback continues directly using the new device. Only volume control doesn't work further. It works after stopping the track and play again.

Unfortunately I'm not able to run Exaile 4.0.x because of outdated packages. But I think there should be no difference, because the code is from 2015.

I could imagine, that this thing wasn't noticed before.

virtuald:
67fda42ce38fe5d8699eb54ab53e1732f78b2173 will need to be reverted, as it completely breaks switching audio streams for me.

If you use the gstreamer graphviz debugging stuff mentioned in the devel docs (you can use this in the simple console)

```python
from xl.player import PLAYER
from gi.repository import Gst
Gst.debug_bin_to_dot_file(PLAYER._engine.main_stream.playbin, Gst.DebugGraphDetails.ALL, "filename")
```

You can see from the output image that the playbin and the playsink both have the correct volume after the sink change, but the audiosink is stuck at whatever volume the parent had.

[...]

Issue #709, opened 2020-05-30
Volume Control Scales Unnaturally


aphirst:
### Steps to Reproduce (for bugs)

1. Use Exaile with Pulseaudio (the Exaile in-program volume control affects the volume of the source/sink that's assigned to Exaile by Pulseaudio directly)
2. Start a familiar music track playing, with reasonably consistent volume throughout
3. Lower Exaile's volume to 0 via the in-application dial
4. Raise Exaile's volume by the smallest possible notch (scroll wheel, one pixel, whatever)

### Expected Behavior

The playback volume should start very low (both in the stated value in a Pulseaudio volume control application and in terms of how it actually sounds), and increase in proportion to how far you increase the dial.

### Current Behavior

With only one "notch" increased in Exaile, the Pulseaudio volume level is already 37% (and it sounds pretty loud, of course). With a second notch it increases to 37%, then 46%, then 53% and it's diminishing returns from there.

### Possible Solution

Sound volume can be measured in a few different ways, linearly and logarithmically, to reflect how we actually perceive the sound. It seems as if this is being overcorrected by Exaile somewhere in the volume control code. The % of Exaile's volume meter should correspond to the % of Pulseaudio's volume source/sink because Pulseaudio is **already** correcting for the difference between decibels and amplitude/energy/whatever.

### Environment

* Operating System and version: Arch Linux (x86_64, rolling release)

* Exaile Version: 4.1.0-alpha1

sjohannes:
As far as I know, we're passing the volume number to GStreamer without doing any logarithmic scaling, but I'm not very familiar with the engine code. @virtuald, any idea?

virtuald:
I don't believe we're doing anything special. @aphirst maybe try disabling fade in the playback settings? That's the only place I'm aware of that anything is multiplied with volume.

There's a commented out debug statement in [engine.py](https://github.com/exaile/exaile/blob/ccde3c0f018d9d61dd47e9f04c9f818b908c8fd3/xl/player/gst/engine.py#L546), might try uncommenting it and see if the internal volume matches the pulse volume?

aphirst:
@virtuald I don't have fade enabled.
![ex](https://user-images.githubusercontent.com/1735556/83610569-8dfbc480-a577-11ea-8b12-40c5572a486b.png)

I think you meant L546:
```
self.logger.debug("Set playbin volume: %.2f", volume)
```
[...]

PR #805, opened 2022-01-31
Reset playback after reconfiguring output device

luzip665:
Fixes: https://github.com/exaile/exaile/issues/801


xl/player/gst/engine.py
```diff
@@ -413,10 +413,22 @@ def reconfigure_sink(self):
...
+        state = self.get_gst_state()
+        if state == Gst.State.PLAYING or state == Gst.State.PAUSED:
+            track = self.current_track
+            position = self.get_position() / 1e9
+            if state == Gst.State.PLAYING:
+                position = position - 1
+            volume = self.get_volume()
+            self.set_volume(0)
+            GLib.timeout_add_seconds(
+                1, self._restart, track, position, state == Gst.State.PAUSED, volume
```
virtuald, line 426: This is weird, why is there a whole second delay here? This seems ripe for causing race conditions.

luzip665: The problem is to deal with the timeout. 
In this way I can seek to one second before.

virtuald: I don't claim to understand the problem yet, but this doesn't feel like a good way to solve it.

luzip665: Problem is described here: https://github.com/exaile/exaile/issues/801#issuecomment-1026314414

May be this should be solved in the dynamic_sink?

luzip665:
@virtuald 
Please, can you have a look for it?
Now it uses the playbin directly and does not deal with volume.

I tried to do exactly this in dynamic_sync::reconfigure with the audio_sink, but then the volume control is lost again.

virtuald:
I will take a look tonight.

virtuald:
You're correct that the volume has likely always been broken. However, this is not the correct fix, and 67fda42ce38fe5d8699eb54ab53e1732f78b2173 should be reverted. I'll post more on the other bug.

xl/player/gst/engine.py
```diff
@@ -358,6 +358,7 @@ def __init__(self, engine):
         # again. Instead, wait for the user to try to play a track,
         # and maybe the issue has resolved itself (plugged device in?)
         self.needs_sink = True
+        self.current_sink = None
```
virtuald, line 361: I'm uncertain about adding another variable here. Why not just add a get_volume to the DynamicAudioSink?

luzip665: This would be the easiest thing, I guess. This is audio_sink, right?
But I am uncertain about your favorite way of solving.

virtuald:
I will try to review tomorrow night, I keep forgetting about this.

xl/player/gst/engine.py
```diff
@@ -542,11 +544,14 @@ def seek(self, value):
...
-    def set_volume(self, volume):
+    def set_volume(self, volume: float) -> None:
         # self.logger.debug("Set playbin volume: %.2f", volume)
         # TODO: strange issue where pulse sets the system audio volume
         #       when exaile starts up...
-        self.playbin.props.volume = volume
+        if not self.needs_sink:
+            self.audio_sink.set_volume(volume)
+        else:
+            self.playbin.props.volume = volume
```
virtuald, line 554: This is a bit unsettling to me. I feel like it should either *all* be managed by the playbin volume, or never managed by the playbin volume.

I'm not sure if this could ever happen, but if someone set the volume to 50% when the audio sink wasn't present, once you add the audio sink the playbin volume can never be adjusted.

luzip665: This happens only right after starting exaile before playing the first track. To me it seems that if I set the playbin volume before configuring the sink, the sink volume is set to this volume after configuring. But may be I am wrong because of https://github.com/exaile/exaile/issues/709.
We can keep the volume in the engine and set volume of the sink right when it is present.

I saw there is `on_volume_change` that is also reading the volume from the playbin.

virtuald: I think if you never touch volume in playbin that would be best.

luzip665:
Now it never touch the volume of the playbin.

luzip665:
Correcting a crash if last selected device is not present at first play after start. FakeSink doesn't have a volume.

xl/player/gst/dynamic_sink.py
```diff
@@ -177,5 +179,19 @@ def _add_audiosink(self, audio_sink, buffer_position):
 
         self.audio_sink = audio_sink
 
+    def set_volume(self, volume: float) -> None:
+        if not hasattr(self.audio_sink.props, 'volume'):
+            self._volume = volume
```
virtuald, line 184: This should always be stored, that way `get_volume` can never return a stale volume?

luzip665:
Finally still need the playbin, because the auto device uses GstAutoAudioSink, that doesn't implement volume.
So playbin is needed to control volume for "auto" audio device.

So, currently it is like
* start playback with "auto" device, volume control works
* then switch to another device, volume control works
* switching back to auto, volume control is broken
* then switching back to another device, volume control works

So, volume control works in more cases than before but not all.

luzip665:
@virtuald 
What do you think about?

virtuald:
Ugh. Stupid gstreamer.

It does seem that it would be better to have playbin manage the volume all the time. I wonder if a bin-in-bin approach referenced in https://github.com/exaile/exaile/issues/801#issuecomment-1030526161 would work all the time. Or, could we just have the bin implement the necessary volume controls? Or a bin that isn't a bin.

Issue #322, opened 2017-02-02
Playlist columns always added on the right end of treeview where users can't see them


grubersjoe:
**Exaile 4.0.0dev**
Arch Linux (AUR package exaile-git)

When right-clicking on a column header in the playlist to choose shown track attributes the view does not update. If you switch afterwards from "Resizable" to "Autoresize" or vice-versa the view updates correctly and shows the newly chosen attribute as well. Sometimes it works instantly, but most of the time not.

**Steps to reproduce**
1. Right click on a column header in the playlist and choose an arbitrary unselected track attribute
2. The view does not update

**Expected result**
The view should update immediately and show all selected track attributes.

genodeftest:
The issue is a bit different, actually: Playlist columns are added, and the view is updated. The issue here is that the column is added on the right, outside of the visible area, so you can't see it. Changing the column resize mode changes column widths, so it gets visible.

PR #807, opened 2022-02-07
Add playlist columns next to the clicked column

luzip665:
Sort playlist columns menu alphabetical
Fixes: https://github.com/exaile/exaile/issues/322


virtuald (changes requested):
Works for me, but when I add a colume via the `View` menu I get an exception:

```
  File "xlgui/widgets/playlist_columns.py", line 751, in on_item_activate
    elif context['trigger_column']:
KeyError: 'trigger_column'
```

luzip665:
Corrected

PR #813, opened 2022-07-06
Debian packaging

luzip665:
Changes necessary to create a debian package according to mentors.debian.org

The package in Debian Mentors: https://mentors.debian.net/package/exaile/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ make-install-dirs:
 	install -d -m 755 $(EXAILESHAREDIR)/data/images/24x24
 	install -d -m 755 $(EXAILESHAREDIR)/data/images/32x32
 	install -d -m 755 $(EXAILESHAREDIR)/data/images/48x48
+	install -d -m 755 $(EXAILESHAREDIR)/data/images/128x128
 	install -d -m 755 $(EXAILESHAREDIR)/data/images/scalable
 	install -d -m 755 $(EXAILESHAREDIR)/data/ui
 	install -d -m 755 $(EXAILESHAREDIR)/data/ui/panel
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -154,7 +154,7 @@ install-target: make-install-dirs
 	install -p -m 644 data/exaile.desktop \
 		$(DESTDIR)$(DATADIR)/applications/
 	install -p -m 644 data/exaile.appdata.xml \
```
virtuald, line 156: Not super familiar with the new specification (... why is there a new specification >_>), but looks like this should have a different filename too? metainfo.xml instead of appdata.xml

luzip665: Seems not to be necessary according to https://wiki.debian.org/AppStream/Guidelines
The debian lintian message is https://lintian.debian.org/tags/appstream-metadata-in-legacy-location

Makefile
```diff
@@ -207,6 +207,7 @@ completion: builddir
 clean:
 	-find . -name "*.~[0-9]~" -exec rm -f {} \;
 	-find . -name "*.py[co]" -exec rm -f {} \;
+	-find . -name "__pycache__" -exec rm -rf {} \;
```
sjohannes, line 210: If I remember correctly, Python 3 puts the `.py[co]` files in `__pycache__`, so I think the previous line is not needed anymore? (Same with `plugins/Makefile`.)

data/exaile.desktop
```diff
@@ -42,3 +42,4 @@ Terminal=false
 StartupWMClass=exaile.py
 MimeType=audio/musepack;application/musepack;application/x-ape;audio/ape;audio/x-ape;audio/x-musepack;application/x-musepack;audio/x-mp3;application/x-id3;audio/mpeg;audio/x-mpeg;audio/x-mpeg-3;audio/mpeg3;audio/mp3;audio/x-m4a;audio/mpc;audio/x-mpc;audio/mp;audio/x-mp;application/ogg;application/x-ogg;audio/vorbis;audio/x-vorbis;audio/ogg;audio/x-ogg;audio/x-flac;application/x-flac;audio/flac;
 Categories=AudioVideo;Audio;Player;GTK;
+Keywords=player,mp3,ogg
```
luzip665, line 45: Does someone know better keywords?
It is obligatory to add.
See https://lintian.debian.org/tags/desktop-entry-lacks-keywords-entry and https://wiki.gnome.org/Initiatives/GnomeGoals/DesktopFileKeywords

Issue #770, opened 2021-12-06
Migrate from bsddb3 to berkeleydb


limburgher:
bsddb3 module is deprecated and replaced with berkely3 module: https://pypi.org/project/bsddb3/

Migration is not drop-in but is uncomplicated.

sjohannes:
For reference:

> The main change you need to be aware is that in *berkeleydb* keys and values are now **bytes**, while in *bsddb3* they were **strings**. You would only need to change your code to add or remove type encoding/decoding, depending of how you use the library. The process should be simple. 

Our database read/writes are very localised (`xl/trax/trackdb.py` and `xl/lyrics.py`) so this should be pretty easy to do. We do still need to support bsddb3, so we have to keep track of which module we are using.

The thing we need to find out is how we can read existing database files. The [berkeleydb website](https://www.jcea.es/programacion/pybsddb.htm) doesn't mention what encoding bsddb3 uses.

sjohannes:
> The thing we need to find out is how we can read existing database files. The [berkeleydb website](https://www.jcea.es/programacion/pybsddb.htm) doesn't mention what encoding bsddb3 uses.

Looks like bsddb3 [uses iso8859-1](https://hg.jcea.es/pybsddb/rev/9ed00d529319), but because we're using Shelf I think it uses utf-8 (Shelf's default keyencoding arg) instead. We only use ASCII as trackdb keys so either one should work as is. For lyrics, it's only a cache so in the worst case we'll get a cache miss.

mexicarne:
Sadly, FreeBSD port was removed from the repository because of this stake dependency.

mexicarne:
Thank you! I'll test it in next couple of days. But I don't have old database, so can only test if works from scratch.

mexicarne:
I'd say it is working fine here with this patch. I run exaile from scratch (no old database or profile files existed) and everything was created without problem while playing some media. Also added some tracks with unicode filenames and unicode mp3 tags (Russian) - all good. So it looks like it atleast fine with new installations, but MAY break for users with pre-existed data directory.

Tested with berkeleydb-18.1.5 with python 3.9.13 and bsddb not-installed.

PR #814, opened 2022-07-25
Try import berkeleydb if bsddb3 fails

sjohannes:
See https://github.com/exaile/exaile/issues/770 for context.

I've barely tested this, so would love to get feedback from people who have berkeleydb installed but not bsddb3.

### DIFF
--- a/xl/common.py
+++ b/xl/common.py
@@ -44,7 +44,12 @@ import urllib.parse
 import urllib.request
 import weakref
 
-import bsddb3 as bsddb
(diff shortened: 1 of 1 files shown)


luzip665:
Works here on Debian 11.
Which points are necessary to check?

sjohannes:
Just that it doesn't break anything in your existing DB (especially on non-Latin tags).

luzip665:
Tested with german umlauts and some chinese chars from google in my existing DB.
No problems found.

PR #818, opened 2022-08-02
tests: remove mox3 dependency

### LINKED ISSUE #817: mox3 is no longer maintained
The [mox3](https://opendev.org/openstack/mox3/) package is no longer maintained and is used by Exaile for testing.  This is now starting to cause some issues on modern Linux distributions ([such as Fedora 37](https://bugzilla.redhat.com/show_bug.cgi?id=2113217)) where the mox3 package is failing to build and thus Exaile is also failing to build.

virtuald:
- Fixes #817

Note that the tests in test_util.py still use mox, but... they're all skipped because I never fixed them previously, so I guess it's ok to keep them skipped? Not 100% sure what the tests were supposed to do.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -112,6 +112,5 @@ BPM Counter plugin:
 
 Test dependencies:
 
-* mox3 (python3)
 * pytest (python3)
 
--- a/tests/xl/trax/test_search.py
+++ b/tests/xl/trax/test_search.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8
 
-from mox3 import mox
+from unittest.mock import call, patch
 
 from xl.trax import search
 from xl.trax import track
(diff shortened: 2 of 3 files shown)


rokm:
> Note that the tests in test_util.py still use mox, but... they're all skipped because I never fixed them previously, so I guess it's ok to keep them skipped? Not 100% sure what the tests were supposed to do.

Should we get rid of `TestGetTracksFromUri` class in `test_util`, along with the import of `mox3`, then? 

I see no point in keeping those skipped tests around if they are broken and are likely to remain such? Plus, is #817 really fixed as long as that `mox3` import remains in there?

virtuald:
There is no mox3 import, merely access to self.mox, so 817 should be fixed.

I don't have strong opinions either way for keeping/removing the skipped test. If someone ever wants to fix it, it might be useful? Of course, it's been there for 6 years now...

rokm:
Ah, right, sorry... I was looking at the original file instead of modified one.

rokm (approved):

virtuald:
:shipit:

sjohannes:
TestGetTracksFromUri is fixed now.

PR #820, opened 2022-08-03
Playlist simplification

ryneeverett:
These changes shouldn't affect behavior but just aim at improving readability. See commit messages for details.

PR #821, opened 2022-08-03
SPAT: leave next track ready to play. Fix #767.

### LINKED ISSUE #767: SPAT should leave the subsequent track next in queue
Stop Playback After This Track should leave the next track stopped and selected after "this track" completes.

1. Toggle on "Stop Playback After This Track"
2. Play the given track to completion.

### Expected Behavior

The subsequent track should be next in queue.

### Current Behavior

The same track is next in queue.

### Possible Solution

I think this should be the default behavior. Why would one want to play the same track again (next time they hit play) when it has already played to completion?

I would be probably be willing to implement it myself.

### Environment

* Operating System and version: Nixos 21.05
* Exaile Version: 4.1.1

ryneeverett:
This branch is based on #820.

<s>With these changes the behavior is as I would expect, except the play/pause button gets left with the pause symbol showing when it is in a stopped state and the button actually behaves as a play button.</s> (fixed)

I do not believe there is any way to accomplish proper SPAT behavior
from within the playlist/queue, so we need to add the ability to
propagate a command to stop up to the engine.

### DIFF
--- a/xl/player/gst/engine.py
+++ b/xl/player/gst/engine.py
@@ -246,8 +246,8 @@ class ExaileGstEngine(ExaileEngine):
         if self.other_stream is not None:
             self.other_stream.stop()
 
-    def play(self, track, start_at, paused):
-        self._next_track(track, start_at, paused, False, False)
+    def play(self, track, start_at, paused, stopped):
+        self._next_track(track, start_at, paused, stopped, False, False)
 
     def seek(self, value):
         return self.main_stream.seek(value)
@@ -299,8 +299,9 @@ class ExaileGstEngine(ExaileEngine):
         self.destroy(permanent=False)
         self.initialize()
 
-    def _next_track(self, track, start_at, paused, already_queued, autoadvance):
-
+    def _next_track(
+        self, track, start_at, paused, stopped, already_queued, autoadvance
+    ):
         prior_track = self.main_stream.current_track
 
         # Notify that the track is done
@@ -313,6 +314,7 @@ class ExaileGstEngine(ExaileEngine):
                 track,
                 start_at,
                 paused,
+                stopped,
                 already_queued,
                 self.crossfade_duration / 1000.0,
                 self.crossfade_duration / 1000.0,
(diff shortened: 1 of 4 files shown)


ryneeverett:
Another issue I'm noticing with this implementation is that whatever hook triggers the inhibit suspend plugin to unsuspend idle is not getting called after playback is stopped by SPAT.

luzip665:
Not sure if I understand correct.

The signal is connected in `xl/players/adapters.py` line 38 and triggered by stopping a track.

ryneeverett:
Thanks! That got me to the point of manually triggering the `on_playback_player_end` event which is a step in the right direction. Unfortunately it doesn't fix the problem because there is still another `on_playback_player_start` event triggered by a `Gst.MessageType.STREAM_START` message after I've stopped playback which undoes everything done by my `on_playback_player_end` event.

luzip665:
May come from `xl/player/gst/engine.py:335`. This is because playlist::__get_next now returns a track instead of None.

The local trace at this moment is
`
on_playback_track_start, __init__.py:169
_emit, event.py:359
emit, event.py:300
log_event, event.py:78
engine_notify_track_start, player.py:446
_next_track, engine.py:335
_autoadvance_track, engine.py:283
_eos_func, engine.py:293
_idle_callback, common.py:185
main, Gtk.py:1632
__init__, main.py:593
main, exaile.py:60
<module>, exaile.py:64
`

ryneeverett:
> May come from `xl/player/gst/engine.py:335`. This is because playlist::__get_next now returns a track instead of None.

Correct. By skipping this in the `stopped` case all the side effects I complained about disappear. However, one of the side effects of the "start" event which is missed is moving the playlist position to the next track so that it is highlighted and the subsequent press of the play button plays it. There does not appear to be any way to do this at present so I added an event for that purpose.

The implementation probably needs some work but I think I finally have the behavior right.

ryneeverett:
To summarize the status: There are no known problems with the behavior but I'm open to suggestions if there are better ways to implement.

luzip665:
Looks good to me.
@virtuald What do you think?

virtuald:
I can't test it at the moment, but this seems fine to me.

PR #824, opened 2022-10-04
Replace cddb by libdiscid

ryneeverett:
This just a rebase and python syntax and formatting update of #652.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -54,7 +54,8 @@ Device detection:
 
 CD info: (TODO: This is currently broken on python3, see #608 and #652)
 
-* cddb (python2), from http://cddb-py.sourceforge.net/
+* python-libdiscid or python-discid (optional on linux, required to use musicbrainz)
+* python-musicbrainzngs (optional)
 
 DAAP plugins (daapserver and daapclient):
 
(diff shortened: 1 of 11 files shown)


luzip665:
I tested with a couple of cds. Fetching the track information seems to work for me.

But there are some problems in the plugin preferences that prevent importing a cd. Can you fix them in this branch?

luzip665 (changes requested):

rokm:
> @sjohannes The package discid should be added to `tools/installer/requirements.txt`, right? What about its dependency to libdiscid (https://python-discid.readthedocs.io/en/latest/install/#libdiscid) ?

`discid` goes to `tools/installer/requirements.txt`, and `mingw-w64-i686-libdiscid` goes [here](https://github.com/exaile/python-gtk3-gst-sdk/blob/master/win_installer/data/sdk.config). We'll need a PyInstaller hook for `discid` to collect that shared lib (I'll probably add one upstream, so by the time this is sorted out and merged it will work out of the box).

luzip665:
Thank you for fixing!
I'll check at weekend

luzip665 (commented):
Works for me.
Please, can you rebase? Then I will merge.

luzip665:
Thanks for rebasing.
But I mean it a way that results in one commit with a well readable commit message.

@rokm It's not working on windows because of missing dbus, right?

rokm:
> @rokm It's not working on windows because of missing dbus, right?

Yeah, looks like the `cd` plugin requires udisks2 (accessible via dbus) for listing devices. 

Which makes my earlier comment about installing `discid` on Windows moot, since the plugin does not work on that platform in the first place...

PR #825, opened 2022-10-04
Select correct playlist and track if starting with option resume_playback and gui

luzip665:
I have the option resume_playback enabled and use queue and playlist

If I play a track in the playlist and close Exaile, after a restart the track in the playlist is neither selected nor played.
Instead the last selected track in queue is played.


virtuald (approved):
Seems fine to me.

PR #826, opened 2022-10-06
Automatically generate .desktop and AppData files

luzip665:
Based on PR #559. Thanks to @Carmina16 !

* po files updated
* merge conflict solved
* Makefile updated

exaile.desktop and exaile.appdata.xml are now build during `make all`


luzip665:
There is one question.
In https://github.com/exaile/exaile/blob/66cff823c84266ec3523677d62faaccd69694ee0/Makefile#L254 `test_compile` is calling `all`, which includes `locale`.
But `locale` does nothing because all the po files were not copied before.
So, should the `all` in  https://github.com/exaile/exaile/blob/66cff823c84266ec3523677d62faaccd69694ee0/Makefile#L257 should be a `compile`?

sjohannes:
`test_compile` doesn't seem used anywhere, so it's probably just added for dev purpose, to test that a 'fresh' build works. I'd say it should probably also copy everything that `make all` needs, including locale files.

luzip665:
Works.
But xgettext is missing for the ci tests.

sjohannes (approved):

sjohannes:
Apparently `test_compile` is called by the CI tests; oops. I've updated the CI Docker images to include xgettext (also had to fix other stuff due to changes in latest Fedora).

PR #835, opened 2022-12-05
Remove lyricwiki plugin.

ryneeverett:
Confusingly, all the documentation says this plugin fetches lyrics from lyrics.wikia.com which seems to be a mirror of the new endpoint which is lyrics.fandom.com.

Regardless, this wiki was removed over 2 years ago for purported licensing infringement. An archive mirror was opened on fandom a few weeks ago and it has already been closed, presumably for the same reason. See <https://community.fandom.com/f/p/3897532950249474242>.

The plugin returns no lyrics and requests to the endpoint return "Error
410: Gone".


virtuald:
Seems reasonable.

PR #836, opened 2022-12-05
Write popm rating to mp3

### LINKED ISSUE #218: Please add an option to store track ratings in ID3 tags.
If I move some files to another folder, I loose all my ratings.

It would be nice to have an optional setting to store the rating in the ID3 tag (clementine has such an option).

Alternatively Exaile could store a checksum of each file and if it rescans the collection it can recognize the file on its checksum even if it has a new path. That way it can keep the rating, play count, date added, ...

luzip665:
Fixes #218


virtuald:
Hm, should there be an action that runs once that applies existing ratings to files?

xl/metadata/_id3.py
```diff
@@ -105,6 +107,11 @@ def _get_tag(self, raw, t):
         elif t == 'COMM':  # Newlines within comments are allowed, keep them
             for item in field:
                 ret.extend([value for value in item.text])
+        elif t == 'POPM':
+            # Rating Stars
+            data = field[0].rating
+            rating = self._rating_to_stars(data)
```
virtuald, line 113: What about tags with multiple POPM fields (each with a unique email?) I think that's how it's supposed to work.

xl/metadata/_id3.py
```diff
@@ -143,6 +150,11 @@ def _set_tag(self, raw, tag, data):
             ]
         elif tag == 'WOAR':
             frames = [id3.WOAR(encoding=3, url=d) for d in data]
+        elif tag == 'POPM':
+            #Rating Stars
+            data = int(data[0])
+            rating = self._stars_to_rating(data)
+            frames = [id3.POPM(email='', rating=rating)]
```
virtuald, line 157: Other players provide a way to configure the rating email.

xl/metadata/_id3.py
luzip665: I think this would work on mp3 only. flac/vorbis and mp4 only use one rating for all.

xl/metadata/_id3.py
luzip665: Here it's the same. Only mp3 supports email in rating.

luzip665:
@virtuald Which tool did you use to check?
I ended up with kid3 and hope it is reliable.

virtuald:
I don't recall tbh. Historically I've used quod libet and beets as guides.

data/ui/preferences/collection.ui
```diff
@@ -62,5 +62,71 @@
         <property name="top_attach">3</property>
       </packing>
     </child>
+    <child>
+      <object class="GtkCheckButton" id="collection/write_rating_to_disk">
+        <property name="label" translatable="yes">Save ratings in files</property>
```
ryneeverett, line 67: Could we be a bit more clear with our wording here? If I'm not mistaken, ratings are already written to a file on disk in `music.db`. This feature would write ratings to the metadata of the audio files themselves, right? Maybe semantics like `collection/write_rating_to_audio_file` and `Save ratings in audio file metadata` would be more clear?

Pkanjai:
> Fixes #218 

> Fixes #218

luzip665:write_popm_rating_to_mp3

luzip665:
@ryneeverett Thanks for your hints.
Do you like to have a look again?

Now it should work for mp3, mp4, flac and ogg.
The migration can be triggered in the collection preferences.

xl/metadata/mp4.py
```diff
@@ -73,6 +74,12 @@ def _get_tag(self, f, name):
             for value in f[name]:
                 ret.append("%d/%d" % (value[0], value[1]))
             return ret
+        elif name == 'rate':
+            # Rating Stars
+            data = int(f['rate'][0])
+            rating = self._rating_to_stars(data)
+            # rating = data
```
ryneeverett, line 81: This comment looks like the debris of scaffolding you could discard.

xl/trax/track.py
```diff
@@ -891,11 +891,21 @@ def get_rating(self):
...
+            f = self._get_format_obj()
+            rating = None
+            if f and '__rating' in f.tag_mapping and settings.get_option('collection/write_rating_to_audio_file_metadata', False):
+                rating = self.get_tag_disk('__rating')
+            if rating:
+                rating = rating[0]
+            elif rating is None:
+                rating = self.get_tag_raw('__rating')
+            if rating is None:
+                return 0
```
ryneeverett, line 903: I would delete the `if rating is None` block and move `rating = float(rating)` to the end here. If `rating is None` then the result will be the same (`float(None) -> TypeError -> return 0`) and it also handles any other case where `rating` cannot be coerced into a `float`.

xl/trax/track.py
```diff
@@ -891,11 +891,21 @@ def get_rating(self):
...
-            rating = float(self.get_tag_raw('__rating'))
-        except (TypeError, KeyError, ValueError):
+            f = self._get_format_obj()
+            rating = None
+            if f and '__rating' in f.tag_mapping and settings.get_option('collection/write_rating_to_audio_file_metadata', False):
+                rating = self.get_tag_disk('__rating')
+            if rating:
+                rating = rating[0]
+            elif rating is None:
+                rating = self.get_tag_raw('__rating')
```
ryneeverett, line 901: The `if rating` / `elif rating is None` clauses do not handle the case where `rating` is falsey but not `None`. This doesn't appear to be a logical possibility, so why not reformulate this so that it is inherently inclusive with either `if rating` / `else` or `if rating is None` / `else`?

ryneeverett (commented):
All in all far easier to follow than before, thanks!

luzip665:
Finally I switched to try...except

xl/trax/track.py
ryneeverett: i still think this is applicable. It seems like the `except TypeError` is better designed to handle this. At this point maybe just replace this with:

```py
try:
    rating = float(rating)
except TypeError:
    return 0
```

Probably don't need to catch the `TypeError`s in the above blocks either as I'm pretty sure that was only to handle the `float` conversion.

ryneeverett: I'm sorry if I haven't communicated this clearly but I don't think I'm getting through.

The original was:

```py
        try:
            rating = float(self.get_tag_raw('__rating'))
        except (TypeError, KeyError, ValueError):
            return 0
```

The `TypeError` was to catch the `float()` part of the statement. I think you still ought to be wrapping that type conversion in a `try` / `except TypeError` and you still ought to be wrapping the rest of the statement (`self.get_tag_raw('__rating')`) in `try` / `except (KeyError, ValueError)`.

luzip665: I think your suggestion is good.
But right know I would say, the approach is odd. Now it would read the rating from file each time the track is shown, which seems to much.
I would say, rating should only be read during scanning the library. Then this part is not necessary anymore.

ryneeverett: If I'm understanding you correctly, you're saying this logic should be reverted to the original and the `get_tag_disk` call should be moved to where the scan occurs and there insert the metadata into the exaile database. That makes good sense to me.

luzip665:
So, reading the ratings from file is now done during scanning the library.
But actually it's necessary to have the library monitored.

@virtuald 
In https://github.com/exaile/exaile/blob/master/xl/collection.py#L250-L253 the scanning of libraries is handled.
What I wonder is the intention of startup_scan and monitored.
```
for library in self.libraries.values():
            if (
                not force_update
                and startup_only
                and not (library.monitored and library.startup_scan)
            ):
                continue
```
So, if the library has startup_scan but not monitored, on startup nothing happens, right?
Is this the intented behaviour?

virtuald:
I'm not sure tbh. I don't have mine monitored, but it does the startup scan... and it seems to do things? Though actually now that I'm thinking about it, maybe it doesn't. 

Usually I just manually do a quick scan when I need to import things?

luzip665:
Based on gui I need both to get a startup scan.

I have `monitored` true and `startup_scan` false. 
In this case changes of files are recognized during Exaile is running.
Changes of files between last time Exaile was running and the next start are not recognized.
This was confusing.

@ryneeverett 
Can you have a look if it works as expected?

ryneeverett (approved):
Code and behavior seems reasonable enough that additional refinement could probably happen in another pull request if preferred.

I did notice one edge case issue with the current behavior, which I presume was introduced in the change from reading the rating metadata each time the file is shown to only reading it at scan:

1. Rate track.
2. Move track to another directory.
3. Re-rate track.
4. "Write existing ratings to track files now"
5. Move track back to original directory.

Because the track is cached at that path in the database, the original rating will be shown, not the one written to the file.

I only tested this with "Files" and not "Collection". I assume this would be resolved on re-scan in the collection.

I would think the solution to this would be storing the timestamp in the database and checking it to decide whether to use cached metadata or to re-read it.

luzip665:
Tracks in Files panel are only read once. After that the metadata is cached.
I think too this is another issue.

Issue #841, opened 2022-12-13
Track start/end events not triggered for not local streams on track change


eri-trabiccolo:
### Steps to Reproduce (for bugs)

1. Enable the OSD plugin (to better appreciate the issue)
2. Add a radio playlist, e.g. any SomaFM playlist
3. Play it
4. Wait for a new track to be parsed

### Expected Behavior
I'd expect to be notified of the track change via OSD

### Current Behavior
OSD doesn't notify the track change

### Possible Solution
I'm proposing [this PR](https://github.com/exaile/exaile/issues/841).
I'm not sure that's the best way to "save" the current track prior to tags changing.

### Environment

* Operating System and version:
Slackware Linux 15.0
* Exaile Version: 
Exaile 4.1.2-beta1+70fa4d1

PR #842, opened 2022-12-13
Fix remote streams not triggering track events

eri-trabiccolo:
Fixes https://github.com/exaile/exaile/issues/841

### DIFF
--- a/xl/player/gst/engine.py
+++ b/xl/player/gst/engine.py
@@ -634,9 +634,13 @@ class AudioStream:
             Useful for streams and files mutagen doesn't understand."""
 
             current = self.current_track
+            remote_newsong = False
             if not current.is_local():
                 prior_track = copy.deepcopy(current)
-                newsong = gst_utils.parse_stream_tags(current, message.parse_tag())
+                remote_newsong = gst_utils.parse_stream_tags(
+                    current,
+                    message.parse_tag()
+                )
 
             if current and not current.get_tag_raw('__length'):
                 res, raw_duration = self.playbin.query_duration(Gst.Format.TIME)
@@ -647,7 +651,7 @@ class AudioStream:
                 if duration > 0:
                     current.set_tag_raw('__length', duration)
 
-            if newsong:
+            if remote_newsong:
                 self.engine.player.engine_notify_track_end(prior_track, False)
                 self.engine.player.engine_notify_track_start(current)
 
--- a/xl/player/gst/engine.py
+++ b/xl/player/gst/engine.py
@@ -31,6 +31,7 @@ from gi.repository import Gst
 import logging
 import os
 import urllib.parse
+import copy
 
 from xl import common
 from xl import event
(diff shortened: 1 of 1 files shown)


luzip665:
Hi.
Thanks for your commits.

There was #823 some weeks ago, where the track tag parsing changed. Did you test with master?
So, I would say, it would be better if the osd plugin is just listening to `track_tags_changed` event.

eri-trabiccolo:
Hi @luzip665 
yeah I tested it with: Exaile 4.1.2-beta1+70fa4d1 (as wrote in the related issue).
I see what you mean, and it's something that I considered, although the OSD plugin wouldn't be the only impacted by that: audioscrobbler and streamripper would be as well - they're currently not working fine with streams because of other issues so I didn't cite them here, but still they rely on those events to be triggered as they were before.
Also listening to `track_tags_changed` implies that every plugin (or any other part of the code) then implements the code to detect a _newsong_ is actually started?!

luzip665:
Ah, sorry, 4.1.2-beta1 sounded for me like https://github.com/exaile/exaile/releases/tag/4.1.2-beta1

IMO it's not a track end. But we should ask @virtuald .

If streamripper plugin is not working well it seems to be a different issue. The streamripper plugin just start streamripper in a shell. So the correct cutting is on behalve of streamripper.

eri-trabiccolo:
@luzip665 
thanks.

Oh yeah about the streamripper you're right, thanks!
(The real issue with the streamripper is that it looks for a player's _pipe (it's uri) which doesn't exist anymore, I replaced that part by getting the uri from the current track, I'm going to open another issue).

virtuald:
I agree that this doesn't feel like the right fix. Perhaps there's another Gst.MessageType that could indicate a new track? So like.. SEGMENT_DONE? or INFO.

On the other hand, since this only affects remote tracks, maybe this isn't so bad. Though, maybe OSD should be listening for a 'tags change' event instead of 'new track'?

luzip665:
Finally I agree. 
But if the change of tag emits a `playback_track_start` the streamripper plugin needs to be changed. Otherwise it will start a new streamripper process each tag change.

@eri-trabiccolo 
Btw. what about streamripper? 
Did you get the stream uri? 
Additionally it seems necessary to add a custom user agent because `streamripper` may be blocked by some stations.

eri-trabiccolo:
@luzip665 
Yeah I got the stream uri from the being played track, replacing this:
https://github.com/exaile/exaile/blob/master/plugins/streamripper/__init__.py#L55
with this:
`options.append(common.sanitize_url(current_track.get_loc_for_io()))`

But I didn't investigate that much on that, there might be a better way(?!)

And yes the user agent must be changed, at least for somafm streams, I tried some combinations and looks like that agents which contain the word "ripper" are forbidden.
I was thinking that maybe an option to customize the ripper user agent should be added.

> But if the change of tag emits a playback_track_start the streamripper plugin needs to be changed. Otherwise it will start a new streamripper process each tag change.

Well only when a "new song" is detected, so not at each tag change. 
The streamripper plugin could keep track of the current being ripped stream URI, and when reacting to `playback_track_start`, do nothing it's the same URI.

PR #843, opened 2022-12-16
Upgrade Icecast plugin to use https://dir.xiph.org/yp.xml

### LINKED ISSUE #829: Icecast Radio not loading streams

1. Open Exaile (fresh install)
2. Choose Radio 
3. Open the Icecast Radio folder

### Expected Behavior

The Icecast Radio folder should show some stations

### Current Behavior

Instead it just says "Loading streams..." (and then never loads). At the bottom it briefly displays the message "Error connecting to Icecast server."

### Environment

* Operating System and version: Windows 10 (21H2)

* Exaile Version: 4.1.3

luzip665:
Upgrading Icecast plugin to use https://dir.xiph.org/yp.xml for retrieving stations.
Fixes #829 

For now it is quite slow. And some optimizations and cleaning needs to be done.

But can someone check if it works correct and as before?


rokm:
> But can someone check if it works correct and as before?

It lists the stations and plays the streams, so at least it seems to work now. 

No idea if it behaves the same as before, though, as I haven't really used it in the past.

luzip665:
One thing with the genres.

There are genres like "Classic Rock". In the xml all genres are mentioned like "Soft Rock Classic Rock Hard Rock Rock".  
For example see station "Uturn Radio: Classic Rock Music".
Actually it's splitting on the whitespace. But so it's impossible to get a correct genre list.
Do you know an url with all genres?

rokm:
Based on the contents, I'd say the `<genre>` tag is a "free-form" one; i.e., the stream owner can enter anything they want. It doesn't even have to be in English (e.g., "Variété Française"). 

So if we have to assign genres, one approach would be to take all values of `<genre>` tag from and cluster them into groups of unique genre values. Another one would be to take a set of standard genre names (e.g., https://www.esm.rochester.edu/ears/docs/lame/id3.html), and looks for those phrases within the `<genre>` strings (and put strings with no matches into special "Unknown" genre).

luzip665:
Finally now the genre list from http://dir.xiph.org/genres is parsed.

PR #844, opened 2022-12-18
xl/trax: search_tracks: remove time.sleep(0)

### LINKED ISSUE #832: Sluggish collection pane UI on Fedora 37 (python 3.11)
With the latest Fedora release and its python 3.11, I'm observing noticeable sluggishness when it comes to the collection UI and its operations:
 - expanding for example an album artist group in the collection pane has a noticeable delay
 - dragging tracks from collection pane into playlist/queue also seems to take forever and freezes the UI for the duration
 
This applies both to the packaged 4.1.2 and git `master`; it likely also requires a large music collection to trigger.

I did some instrumentation for the group expansion codepath, and compared the results in a minimal python 3.10 virtual environment and minimal python 3.11 virtual environment (to eliminate any issues specific to Fedora python packages, and to match both environments in term of `mutagen`, `pygobject`, `dbus-python`, and `bsddb3` versions). 

The initial finding was that this `trax.sort_result_tracks(tags, srtrs)` call 
https://github.com/exaile/exaile/blob/4d72f7ed4b03a6c28d0f0c005fc024aa63cbcf38/xlgui/panel/collection.py#L620
takes 0.12 second in python 3.10, but 1.5 second in python 3.11 - hence the noticeable delay when expanding the group.

Further investigation with many dead ends finally revealed that it is not the `sort_result_tracks` per se that has slowed down; it is retrieval of tracks from the `srtrs˙ generator where most time is spent (confirmed by adding `srtrs = list(srtrs)` before that call, and timing both calls).

This finally brings us to the culprit:

https://github.com/exaile/exaile/blob/4d72f7ed4b03a6c28d0f0c005fc024aa63cbcf38/xl/trax/search.py#L540-L546

Removing that `sleep(0)` call seems to fix the problem (will patch the daily-use exaile to see if there's any noticeable side effects).

rokm:
~When processing the tracks in `search_tracks` generator, yield the current thread's timeslice only when running in the secondary (non-main) thread. This prevents unnecessary slow-downs in the main thread, where we want to finish the operation as fast as possible (especially under python 3.11 on linux, where `time.sleep(0)` introduced additional overhead).~

When processing the tracks in `search_tracks` generator, do not explicitly yield the current thread's time-slice via `time.sleep(0)`.
    
Removal of sleep call prevents unnecessary slow-downs in the main thread, where we want to finish the operation as fast as possible (especially under python 3.11 on linux, where `time.sleep(0)` introduced additional overhead).
    
The comment accompanying the sleep call refers to GIL behavior, which might have been applicable in python2-era, but does not seem to be applicable to python3 anymore.

Fixes #832.

### DIFF
--- a/xl/trax/search.py
+++ b/xl/trax/search.py
@@ -24,7 +24,6 @@
 # do so. If you do not wish to do so, delete this exception statement
 # from your version.
 
-import time
 import re
 from typing import Collection
 
@@ -536,14 +535,6 @@ def search_tracks(trackiter, trackmatchers: Collection[TracksMatcher]):
             srtr = SearchResultTrack(srtr)
         if all(tma.match(srtr) for tma in trackmatchers):
             yield srtr
-
-        # On large collections, searching can take a while. Due to
-        # peculiarities in python's GIL that means the now-cpu-bound
-        # thread running the search can end up blocking other threads.
-        # Calling out to time.sleep forces a release of the GIL and
-        # allows other threads to run. Benchmarks show this has no
-        # noticeable effect on search speed.
-        time.sleep(0)
 
 
 def search_tracks_from_string(


rokm:
I'm kind of torn between completely removing that `time.sleep(0)` and let Python's scheduler do its thing in multi-threaded environment. But I have no idea of how much of an impact that would actually have on UI and main thread for search operations that are running in the secondary thread. On the other hand, I am also unsure how many of such threaded search operations we even have...

So I'm limiting the removal of that `time.sleep(0)` only to cases when we are in the main thread, which should have no reason to yield its time slices to other threads.

virtuald:
It's worth noting that the `time.sleep(0)` was introduced in 2010 (in https://github.com/exaile/exaile/commit/e2174a690d438853de44b2eacd4278b8a3f5d246)... so the behavior that was characterized would have been Python 2. Python 3 is likely to have very different characteristics.

At some point I had written a tool to generate really large collections (100k+) of random files to see what performance impacts occurred. Might be worth removing the sleep(0) and see what happens on a large collection on something that isn't 3.11?

Found the tool: https://github.com/exaile/exaile-test-files/blob/master/collection-generator/generate.py

rokm:
That's a good point. 

I'm not familiar with Python 2 and its GIL behavior; so like the comment implies, perhaps it really had to do something with GIL, even in the main thread. From my understanding of Python 3 behavior, though, that `time.sleep(0)` *might* seems to make sense if we iterated over the search results in a secondary thread, and wanted to yield its time slices to main thread as often as possible (to make the UI, which is running in the main thread, as responsive as possible). In my experience, though, python's thread switching seems to do its job fairly well, so this sort of timeslice yielding seems rather unnecessary even for multi-threaded operation. In the main thread, it makes no sense to me, though. 

But I agree; instead of this partial work-around, it makes more sense to test the behavior with `time.sleep(0)` completely removed under various OSes and python 3 versions, and do away with it completely if possible.

That said, do you happen to know if we have any operations that might be using `search_tracks` from a secondary (worker) thread? All invocations I came across seem to come from the main thread...

rokm:
I generated a large test collection (50k entries) and tested the behavior with python 3.7~3.11 on my Fedora linux system, and with mingw/msys2 python 3.10 under Windows. Except with 3.11 on linux, where the UI slow-downs are fixed, I cannot tell the difference before and after the `time.sleep(0)` call is (completely) removed. 

That said, I think most of the standard UI operations (expanding entries in collection panel, searching incollection panel, drag&drop-ing tracks from collection panel) use the search function in the main thread anyway.

So I've revised the PR to completely remove that `sleep` instead of conditioning it on the non-main thread.

PR #850, opened 2023-01-03
Fix python 3.7 compatibility

rokm:
The `typing` package from python 3.7 stdlib does not provide `Literal` type, so external `typing-extensions` package must be used instead.

Alternatively, we should drop official support for python <= 3.7.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -12,6 +12,7 @@ Core dependencies
 Essential:
 
 * python3 >= 3.6
+* python3-typing-extensions (python3 <= 3.7 only)
 * python3-bsddb3
 * gtk+ >= 3.22
 * gstreamer (>= 1.14)
--- a/xlgui/widgets/info.py
+++ b/xlgui/widgets/info.py
@@ -25,7 +25,16 @@
 # do so. If you do not wish to do so, delete this exception statement
 # from your version.
 
-from typing import Literal
+try:
+    from typing import Literal  # python >= 3.8
+except ImportError:
+    # python <= 3.7 requires typing_extensions package
+    try:
+        from typing_extensions import Literal
+    except ModuleNotFoundError as e:
+        raise RuntimeError(
+            "Please install typing-extensions package to run exaile under python <= 3.7."
+        ) from e
 
 from gi.repository import GLib
 from gi.repository import Gtk


sjohannes:
I think we can drop Python 3.7 support. Debian Stable and latest RHEL are on 3.9.

Corin-EU:
I have been trying to get the latest exaile running on openSUSE Leap 15.6 which is still on python3.6

After installing the needed python3-bsddb3 package, I was getting

          File "/usr/local/stow/Audio/exaile-r.89363501/lib/exaile/xlgui/widgets/info.py", line 28, in <module>
          from typing import Literal
          ImportError: cannot import name 'Literal'

After a fruitless search for the openSUSE package that was "literal" for python3, and finding that it came from the module "typing" but openSUSE has python3-typing-extensions,, a further web search revealed the explanation of the problem and solution proposed three and a half years ago by kaustubhgupta in https://github.com/adamerose/PandasGUI/issues/159

Simply replace "import literal from typing" with

        try:
        from typing import Literal
        except ImportError:
        from typing_extensions import Literal

and exaile 4.1.4-beta1 (latest r.8936350) is up and running (even though the welcome screen says 4.1.3) with python3.6 on openSUSE Leap 15.6.

sjohannes:
That's great info, thanks. (I'm surprised that we haven't added anything else that broke Python 3.6/3.7 support.) An alternative is to just remove all the `Literal` type hints.

In case you run into other issues that can't be easily worked around: I found [this announcement](https://news.opensuse.org/2023/06/06/leap-modern-python-stack/) saying that openSUSE Leap has Python 3.11 in its repository. But I think it doesn't have all the required modules and you may have to supplement that with the [devel:languages:python:backport](https://build.opensuse.org/project/show/devel:languages:python:backports) repository.

Corin-EU:
"openSUSE Leap has Python 3.11 in its repository"

Yes that is the case and I do have python3.11 installed as well.  But when I started looking for the versions of the 3.11 modules necessary, I was having difficulty finding them and for some I was confused by the impression that they were built for 3.12 rather than 3.11.

Python being rather out of date on openSUSE has been an issue raised quite some time ago (2022) and my recollection was that they were holding back because official SUSE (commercial) was not ready to move forward from 3.6 (presumably because of customers still recovering from the shock of having to migrate from 2.7).

Anyways, openSUSE 16 will arrive sometime in the Fall 2025.

A quick web search results in no clues as to which python version it will have, but I guess it should be 3.10 at least, and hopefully with a newer version than the current exaile 4.1.2. so all these old python version  compatibility  issues should then fade away and the "if version less than) tests can be expunged.

PR #851, opened 2023-01-03
tools/installer: switch to upstream PyInstaller

rokm:
Exclude GStreamer plugins via corresponding hook configuration option that was introduced in PyInstaller 5.4. Switch to upstream PyInstaller release instead of using private fork.


tools/installer/exaile.spec
```diff
@@ -27,87 +27,85 @@ binaries = []
...
-  ('_inst/usr/share/exaile/data', 'data'),
-  ('_inst/usr/share/exaile/plugins', 'plugins'),
-  ('_inst/usr/share/locale', 'share/locale')
+datas = [
+    ('_inst/usr/share/exaile/data', 'data'),
+    ('_inst/usr/share/exaile/plugins', 'plugins'),
+    ('_inst/usr/share/locale', 'share/locale'),
+]
+
+# requires https://github.com/pyinstaller/pyinstaller/pull/7040 (PyInstaller >= 5.4)
```
sjohannes, line 36: I think just remove this comment.

tools/installer/exaile.spec
```diff
@@ -27,87 +27,85 @@ binaries = []
...
-      'gstxvimage',
-      'gstzbar',
-    ]
-    
-    def _exclude(b):
-        for r in to_remove:
-            if r in b and 'libgstvideo-1' not in b:
-                print("Excluding", b)
-                return True
-        return False
```
sjohannes, line 72: I don't remember what this is for; is it not needed anymore?

sjohannes (commented):
Off-topic, but I wonder if it's possible to add this .spec file to our Black run.

tools/installer/exaile.spec
rokm: Done.

tools/installer/exaile.spec
rokm: This is how we were removing the unwanted gstreamer plugins up until now; it is part of the `assemble_hook` function, which was executed by our modified PyInstaller before performing binary analysis.

The same can now be achieved by explicitly filtering gstreamer plugins via PyInstaller's hook configuration.

rokm:
This should now collect (or rather, exclude) the same plugins as the old approach.

PR #852, opened 2023-01-04
Add metadata handler for aac files

luzip665:
According to https://mutagen.readthedocs.io/en/latest/api/aac.html ID3 tags can be used with aac.

Checked with kid3.


luzip665:
Seems more complex.
aac files can at least be 
``` Audio file with ID3 version 2.3.0, contains:\012- MPEG ADTS, AAC, v2 LC, 22.05 kHz, stereo```
like tracks recorded from somafm or
``` ISO Media, MP4 Base Media v1 [ISO 14496-12:2003]```
like the test aac file

PR #861, opened 2023-03-06
Lyricsmania doesn't display first line of lyrics (fixes #860)

### LINKED ISSUE #860: LyricsMania 4.1.3 plugin doesn't display the first string
This plugin is working fine and can fetch lyrics from lyricsmania.com when it's available for the playing song but it's never displaying the first line from the lyrics.

Steps to Reproduce

1. Enable plugin in Preferences. 
2. Start to play any track for what the lyrics available at lyricsmania.com website.

Expected Behavior

All lines of lyrics should be displayed by plugin and not skipping the first one.

Current Behavior

The first line of lyrics is not displayed.

Environment

Operating System and version: Ubuntu 20.04.5 LTS
Exaile Version: 4.1.3-beta2

luzip665:
For me this seems to be a bug in lxml. See https://bugs.launchpad.net/lxml/+bug/2009356

### DIFF
--- a/plugins/lyricsmania/__init__.py
+++ b/plugins/lyricsmania/__init__.py
@@ -77,14 +77,14 @@ class LyricsMania(LyricSearchMethod):
         try:
             lyrics_body = lyrics_html.find_class('lyrics-body')[0]
             video_elem = lyrics_body.get_element_by_id('video-musictory')
-
+            first_line = video_elem.tail
+            lyrics_body.remove(lyrics_body.get_element_by_id('video-musictory'))
             lyrics = lyrics_body.text_content()
-            lyrics = lyrics.replace(video_elem.text_content(), '', 1)
             lyrics = re.sub(r'^\s+Lyrics to .+', '', lyrics)
+            lyrics = first_line + lyrics
             lyrics = lyrics.replace('\t', '')
             lyrics = self.remove_script(lyrics)
             lyrics = self.remove_html_tags(lyrics)
-            lyrics = lyrics.strip()
         except Exception:
             raise LyricsNotFoundException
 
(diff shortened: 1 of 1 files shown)


luzip665:
It's intended to work like this. The first line needs to be readded after removing the first element.

plugins/lyricsmania/__init__.py
```diff
@@ -76,8 +76,12 @@ def find_lyrics(self, track):
 
         try:
             lyrics_body = lyrics_html.find_class('lyrics-body')[0]
+            video_elem = lyrics_body.get_element_by_id('video-musictory')
+            first_line = video_elem.tail
             lyrics_body.remove(lyrics_body.get_element_by_id('video-musictory'))
-            lyrics = re.sub(r'^\s+Lyrics to .+', '', lyrics_body.text_content())
+            lyrics = lyrics_body.text_content()
+            lyrics = re.sub(r'^\s+Lyrics to .+', '', lyrics)
```
rokm, line 83: FWIW, I think this line is a no-op (nowadays), because the *Lyrics to ...* element seems to be contained in the parent `<div>` of the `lyrics-body` `<div>` that we process, so we don't actually ever see that string.

rokm (approved):
Neat; this does indeed bring back the missing first line in the lyrics (at least for the couple of songs that I've tested with).

plugins/lyricsmania/__init__.py
```diff
@@ -76,8 +76,12 @@ def find_lyrics(self, track):
 
         try:
             lyrics_body = lyrics_html.find_class('lyrics-body')[0]
+            video_elem = lyrics_body.get_element_by_id('video-musictory')
+            first_line = video_elem.tail
             lyrics_body.remove(lyrics_body.get_element_by_id('video-musictory'))
```
sjohannes, line 81: Does this work?

```suggestion
            lyrics_body.remove(video_elem)
```

luzip665: You are right

PR #864, opened 2023-03-07
Set LANGUAGE=C for 'make test' to get rid of localisation errors

luzip665:
Without this env variable it's impossible to run tests successfully in a non-english environment.


luzip665:
@sjohannes 
Hi,
could you check this at yours please?

sjohannes (approved):
Eep, that sound like a Python bug. Oh well....

PR #870, opened 2023-03-26
win installer: manually collect libsoup-2.4-1.dll and its dependencies

### LINKED ISSUE #867: Not playing any radio stream on Win 11
Current head of Exaile is not playing any radio streams.

The last working build I found is 1106 from https://github.com/exaile/exaile/commit/938edd7464655aded30eec29f7c554d0737ce05c.
It stops playing starting with build 1107 from https://github.com/exaile/exaile/commit/863673cedb839925dbfe303755fc02cc6b6fd282.

rokm:
In GStreamer 1.22.1 used by our SDK 39, the `soup` GStreamer plugin, which is used for radio streams, is not linked against the `soup` shared library (DLL) anymore, so PyInstaller fails to pick up the DLL and some of its dependencies. Failing to load the `soup` plugin causes fall-back to `curl` one, which seems to fail to load any radio stream.

Therefore, ensure that `libsoup-2.4-1.dll` is collected, by passing it to PyInstaller via `binaries` in the spec file.

Fixes #867.

### DIFF
--- a/tools/installer/exaile.spec
+++ b/tools/installer/exaile.spec
@@ -1,6 +1,7 @@
 # -*- mode: python -*-
 
 import sys
+import pathlib
 from PyInstaller.utils.hooks import collect_submodules
 
 if sys.platform == 'win32':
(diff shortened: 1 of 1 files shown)


luzip665:
Works for me.

Issue #682, opened 2020-01-24
FLAC BPMs are written to non-standard TEMPO tag


dangmai:
Exaile's [mapping](https://github.com/exaile/exaile/blob/8fb0ebb4ea108792d38e2b8ad355b27d70e1d246/xl/metadata/flac.py#L36) for FLAC files use the `tempo` tag to store bpm information, but in all the documents that I can find, it should be stored in a `BPM` tag instead. For example, [this mapping](https://picard.musicbrainz.org/docs/mappings/) from Musicbrainz, or [this mapping](http://wiki.hydrogenaud.io/index.php?title=Tag_Mapping) from HydrogenAudio. This makes it difficult to use the BPM information in other tools, for example I use [Beets](https://beets.io/) to manage smart playlists, which does not take into account the `tempo` tag at all.
### Steps to Reproduce (for bugs)

1. Enable the BPM plugin
2. Tap the BPM to a random FLAC song
3. Use any tag reading tool (e.g. `ffprobe`) to read the tag content
4. See that the BPM is stored in the `tempo` tag

### Possible Solution

IMO just changing the tag mapping only is not enough, since it will confuse users when they re-scan their collection and see their BPM information disappear. We can either do a one time migration from the `tempo` tag to the `BPM` tag, or we can offer the option in the setting for the user to choose which tag they'd rather use, and let them handle the migration outside of Exaile if they choose to do so.

### Environment

* Operating System and version: Windows 10

* Exaile Version: Exaile 4

sjohannes:
AFAIK our goal is to follow the MusicBrainz mapping. So yes, it's currently wrong.

Looking at the code just makes me confused.

* If FLAC uses the Vorbis tag mapping, can't we merge it into ogg.py as a subclass of OggFormat (just like OggOpusFormat)?
* The `bpm` tag mapping in ogg.py is also wrong....
* That `comment -> description` mapping also looks wrong.

We'll have to do a one-time migration, but I'm not sure what the best way to do it is. Handling conflicts will be an issue as well (if the file has both tags). And then there's also the problem that non-library files won't get fixed.

genodeftest:
How about:
1. Reading both the standard (`BPM`) and the non-standard (`tempo`) tag, with `BPM` always winnig
2. Writing only the standard (`BPM`) tag from now on.

For when the migration happens::
* As a user, I would be surprised if my files changed on disk without me knowing, I wouldn't like a "silent" migration.
[...]

PR #871, opened 2023-03-26
Migrate metadata mapping

luzip665:
Some metadata mappings are not like the standard.
Sources:
https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html
https://wiki.hydrogenaud.io/index.php?title=Tag_Mapping
https://docs.kde.org/trunk5/en/kid3/kid3/commands.html#frame-list

See also discussion at #682 

For flac and ogg there is new mapping and migration for bpm and comment.

### DIFF
--- a/xl/main.py
+++ b/xl/main.py
@@ -646,6 +646,7 @@ class Exaile:
 
         # Activate new flac/ogg metadata handling for new installations
         from xl.migrations.settings import flac_tempo
+
         flac_tempo.migrate()
 
         # TODO: enable audio plugins separately from normal
--- a/xl/main.py
+++ b/xl/main.py
@@ -644,6 +644,11 @@ class Exaile:
 
         engine.migrate()
 
+        # Migrate flac tempo tag
+        from xl.migrations.settings import flac_tempo
+
+        flac_tempo.migrate()
+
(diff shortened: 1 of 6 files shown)


xl/main.py
```diff
@@ -644,6 +644,11 @@ def __init(self):
 
         engine.migrate()
 
+        # Migrate flac tempo tag
```
sjohannes, line 647: Is this just for FLAC or also Ogg (/Vorbis) files?

sjohannes (commented):
I suppose the problem here is that the migration only works for files in the collection. But I'm not sure I have a better alternative.

xl/main.py
luzip665: It's for both. 
But the migration only sets the setting to true in case of a first run and false otherwise.

luzip665:
@sjohannes 
The migration works only with files in the collection.
But how could we change files outside of it? I mean, we could offer a context menu entry to migrate each track individually, but I'm not sure if this is necessary.

sjohannes:
I was thinking that we'd notify the user when we see an external file with the 'wrong' tags. But I don't know if it's worth the effort.

PR #872, opened 2023-04-29
Add plugin path to sys.path to allow dynamic loading of modules

luzip665:
The cd-plugin loads the parser modules for discid and musicbrainzng dynamically to catch missing dependencies.

This works in case Exaile is run from source and plugins are in `{$exaile_dir}/plugins` because the `$exaile_dir` is in sys.path

If Exaile is installed to the system the exaile directory is /usr/lib/exaile and plugins are in /usr/share/exaile.
In this case the plugins directory is not in sys.path

### DIFF
--- a/plugins/cd/__init__.py
+++ b/plugins/cd/__init__.py
@@ -56,8 +56,11 @@ class CdPlugin:
 
     def __import_dependency(self, module_name):
         try:
-            full_name = 'plugins.cd.' + module_name + '_parser'
-            return importlib.import_module(full_name)
+            sys.path.append(os.path.dirname(__file__))
+            full_name = module_name + '_parser'
+            imported = importlib.import_module(full_name)
+            sys.path.pop()
+            return imported
         except ImportError:
             logger.warn(
                 'Cannot import optional dependency "%s" for plugin cd.', module_name


sjohannes:
I'm confused, why does the code do dynamic import like that instead of using a plain `try: import ...; except: ...`?

Temporarily changing `sys.path` feels a bit weird. I wonder if [this](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) can be used instead.

luzip665:
@ryneeverett Why are this imports done in this way?

@sjohannes Your example is used in https://github.com/exaile/exaile/blob/master/xl/plugins.py#L78-L89 for the same reason. IMO it's the same as manipulating sys.path.
But we could also offer a lazy import function in general?

ryneeverett:
> @ryneeverett Why are this imports done in this way?

I don't know, that's how it was implemented in #652 and I didn't have any reason to change it. I have no objection to moving to a more idiomatic import mechanism.

ryneeverett:
Would it work to simply change it to a relative import?

```py
            full_name = '.' + module_name + '_parser'
            return importlib.import_module(full_name, package=__name__)

```

luzip665:
I don't think so.

The problem is sys.path.
In case Exaile is running from source in https://github.com/exaile/exaile/blob/master/exaile.py#L52 the base directory is added to sys.path. For this it works to import the modules relative to this directory, because the plugins directory is also in this path. Current working directory in this case is also the base directory of Exaile.
In case Exaile is really installed (after `make install`) the base path is `/usr/lib/exaile`. This is also added to sys.path. Plugins are placed in `/usr/share/exaile/plugins`, which is now not part of sys.path. In this case even the current working directory is users home directory.
So, I think it is not possible to get the plugins directory relative to the exaile base directory.

You can check this using the IPython console from plugins.


For me the way of importing the modules in cd plugin is okay, because it makes it possible to use the plugin even without musicbrainzng or discid installed. Even I would like to better inform users about missing modules. But it's another issue.

ryneeverett:
I've been using the cd plugin in an installed exaile and haven't had a problem.

If plugins aren't in sys.path, how do any of the plugins (including the cd plugin itself) get imported?

luzip665:
The plugins get imported using https://github.com/exaile/exaile/blob/master/xl/plugins.py#L78-L89

CD plugin is working but not connecting to musicbrainz, this is my point.
If I run `exaile --debug` there is following in log
```
00:47:12 WARNING : Cannot import optional dependency "discid" for plugin cd. (cd)
00:47:12 DEBUG   : Traceback for importing dependency "discid" for plugin cd. (cd)
  File "/usr/share/exaile/plugins/cd/__init__.py", line 62, in __import_dependency
    return importlib.import_module(full_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'plugins'

Locals at innermost frame:

{ 'import_': <function _gcd_import at 0x7f6339363400>,
  'name': 'plugins',
  'parent': '',
  'path': None,
  'spec': None}

00:47:12 WARNING : Cannot import optional dependency "musicbrainzngs" for plugin cd. (cd)
00:47:12 DEBUG   : Traceback for importing dependency "musicbrainzngs" for plugin cd. (cd)
  File "/usr/share/exaile/plugins/cd/__init__.py", line 62, in __import_dependency
    return importlib.import_module(full_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'plugins'

Locals at innermost frame:

{ 'import_': <function _gcd_import at 0x7f6339363400>,
  'name': 'plugins',
  'parent': '',
  'path': None,
  'spec': None}

00:47:12 DEBUG   : Loaded plugin cd (xl.plugins)
```

ryneeverett:
I see. Yeah, given the hacky plugin system I can't see a better workaround than you have here.

ryneeverett (approved):

PR #874, opened 2023-05-07
Small trifles with cd plugin

luzip665:
Some small fixes:
* Missing gettext
* Retrieving default format from formats list
* encode as mp3 needs to use lamemp3enc now

FWIW reading the disc index (https://github.com/exaile/exaile/blob/master/plugins/cd/__init__.py#L174) is seems not to be running on its own thread.
At least here Exaile gui is blocked during reading the index.
@ryneeverett Do you know of this?

### DIFF
--- a/plugins/cd/_cdguipanel.py
+++ b/plugins/cd/_cdguipanel.py
@@ -138,7 +138,7 @@ class CDImporter:
 
     def __prepare_transcoder(self):
         formats = transcoder.get_formats()
-        default_format = next(formats.keys())
+        default_format = next(iter(formats))
         self.format = settings.get_option("cd_import/format", default_format)
         default_quality = formats[default_format]['default']
         self.quality = settings.get_option("cd_import/quality", default_quality)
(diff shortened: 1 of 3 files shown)


ryneeverett:
> FWIW reading the disc index (https://github.com/exaile/exaile/blob/master/plugins/cd/__init__.py#L174) is seems not to be running on its own thread. At least here Exaile gui is blocked during reading the index. @ryneeverett Do you know of this?

I don't know any more than you but it doesn't surprise me that the implementation of async seems to have been overlooked. The only computer I have ancient enough to have a cd player is so slow it is entirely unresponsive while reading a cd, so I wouldn't have noticed.

plugins/cd/_cdguipanel.py
```diff
@@ -138,7 +138,7 @@ def __init__(self, tracks):
 
     def __prepare_transcoder(self):
         formats = transcoder.get_formats()
-        default_format = next(formats.keys())
```
ryneeverett, line 141: I don't really get this one. I thought these were identical in python3.

luzip665: Me too. I am quite sure that it worked with python 3.9 when I tested the plugin last time.

PR #875, opened 2023-05-19
wikipedia: use WebKit2 4.1(and libsoup 3)

yselkowitz:
The only difference between 4.0 and 4.1 is the libsoup version (2.4 vs 3.0) used by each.  libsoup 2.4, and hence WebKit2 4.0, are unmaintained and are being dropped by distributions.  Nothing else here uses libsoup directly, so no further porting is required.

### DIFF
--- a/plugins/wikipedia/__init__.py
+++ b/plugins/wikipedia/__init__.py
@@ -30,7 +30,7 @@ from . import preferences
 
 import gi
 
-gi.require_version('WebKit2', '4.0')
+gi.require_version('WebKit2', '4.1')
 from gi.repository import WebKit2
 
 


luzip665:
Hi @yselkowitz 

thank you for you contribution.
Fwiw on Ubuntu 22.04 there are two packages for webkit2gtk. There's version 2.38.6 with API 4.0 and with API 4.1.
Version 2.38.6 was released in april.
At least on Ubuntu 20.04 there is only 2.38.6 with API 4.0.

Please can you add a switch as in https://github.com/exaile/exaile/blob/master/xl/common.py#L48-L51 ?
So we would support API Version 4.1 without breaking support for Ubuntu 20.04.

luzip665:
@yselkowitz 
Btw. is the plugin working for you?
Here (Ubuntu 22.04, Webkit2gtk 4.1) in https://github.com/exaile/exaile/blob/master/plugins/wikipedia/__init__.py#L127 bytes are returned and not accepted by https://github.com/exaile/exaile/blob/master/plugins/wikipedia/__init__.py#L141

So, here it doesn't show anything.

luzip665:
Solved in #888

PR #878, opened 2023-06-16
chore: Deduplicate code from Python standard lib

genodeftest:
As we are supporting Python >= 3.6, we can rely on this function being present in Python's standard library.

### DIFF
--- a/plugins/equalizer/__init__.py
+++ b/plugins/equalizer/__init__.py
@@ -26,8 +26,8 @@
 
 
 import logging
-import os
 import math
+import os
 
 from gi.repository import Gst
 from gi.repository import Gtk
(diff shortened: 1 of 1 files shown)


plugins/equalizer/__init__.py
```diff
@@ -27,6 +27,7 @@
 
 import logging
 import os
+import math
```
sjohannes, line 30: Sorting alphabetically :-)

```suggestion
import math
import os
```

sjohannes (approved):

plugins/equalizer/__init__.py
genodeftest: Nice catch, thanks!

PR #879, opened 2023-06-16
chore: Remove outdated comment

genodeftest:
Python 2 is long dead and unsupported


sjohannes (commented):
The comment is saying that in Python 3 we can use the ["raise ... from e"](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) syntax (and omit printing out `e` in the string).

(Actually in both cases we don't need to print `e`, Python will do it for us. But using `from e` we get a slightly clearer error message, as the examples show in the documentation.)

PR #882, opened 2023-06-17
chore: drop dysfunctional jamendo plugin

genodeftest:
Jamendo v2 API has been deprecated since 2014¹ and removed some time ago². As I have not seen any issue (bug report) related to this plugin, I assume it is no longer important to any user.

If anyone objects, that's fine for me. In this case I ask you to open an issue (bug report) instead or fix the plugin. One would need an API key for that which I don't have.

_____
¹ [old documentation](https://web.archive.org/web/20200916080854/https://developer.jamendo.com/v3.0/docs) mentions:
> Old versions of the Jamendo Api, the 'get' and 'get2' versions, are *deprecated*, but for the moment still available through the _api.jamendo/get2/*_ and _api.jamendo/get/*_ urls. Those Api will still be working *until mid-2015*, but the returned data hasn't been updated since Jan 2014. 

² it vanished from Jamendo's API documentation some time between [2020-09-16](https://web.archive.org/web/20200916080854/https://developer.jamendo.com/v3.0/docs) and [2020-09-27](https://web.archive.org/web/20200927093939/https://developer.jamendo.com/v3.0/docs), so I guess the old API was shut down around that time or before.


sjohannes (approved):
Oh, that's too bad. We refer to the Jamendo plugin [in our documentation](https://exaile.readthedocs.io/en/stable/search.html?q=jamendo&check_keywords=yes&area=default); we should probably do something about that as well.

PR #884, opened 2023-06-26
Disable fuzzy matching for msgmerge

luzip665:
In this PR I would like to disable fuzzy matching for msgmerge.

While generating po files for #877 I saw that `msgmerge` fuzzy matching wants to use the translation of `Disc metadata` as the translation of `Migrating metadata`, which would be wrong. This happens even with the current master

If using the `--no-fuzzy-matching` switch it won't happen.

Example without `--no-fuzzy-matching`
```
#: ../xlgui/preferences/collection.py:195
#, fuzzy
#| msgid "Disc metadata"
msgid "Migrating metadata"
msgstr "CD-Metadaten"
```

Example with `--no-fuzzy-matching`
```
#: ../xlgui/preferences/collection.py:195
msgid "Migrating metadata"
msgstr ""
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ pot:
 	  xgettext -j -o messages.pot --keyword=X-GNOME-FullName '../data/exaile.desktop.in' && \
 	  xgettext -j -o messages.pot '../data/exaile.appdata.xml.in' )
 	find po -name '*.po' -exec \
-	  msgmerge --previous --update {} po/messages.pot \;
+	  msgmerge --previous --no-fuzzy-matching --update {} po/messages.pot \;
 	echo $(LINGUAS) > po/LINGUAS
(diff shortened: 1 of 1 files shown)


sjohannes:
I think using `--no-fuzzy-matching` is the right move, but it's not very convenient if a small change results in translators having to start from scratch translating that string. Is there a way to use fuzzy matching but only as comment (and have that somehow be displayed by Weblate)?

luzip665:
It's a pitty.

Sometimes it is very helpful (as in https://github.com/exaile/exaile/blob/master/po/en_GB.po#L4556-L4560)
Sometimes it is just wrong (as in https://github.com/exaile/exaile/blob/master/po/en_GB.po#L4625-L4629)

PR #889, opened 2023-07-30
ci: re-enable IRC messages

genodeftest:
The referenced issue in pydle has been closed a long time ago.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,26 +76,26 @@ jobs:
             --title "Exaile ${version}" \
             --notes "${notes}"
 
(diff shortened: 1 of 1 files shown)


sjohannes:
Since Freenode imploded, we don't have an IRC channel anymore. (Even when we had one it was basically empty anyway.)

genodeftest:
> Since Freenode imploded, we don't have an IRC channel anymore. (Even when we had one it was basically empty anyway.)

Thanks for the hint.

sjohannes:
We should probably just remove that code. I don't think anyone is too keen on using IRC these days.

PR #892, opened 2023-07-30
chore: Drop some outdated backward compatibility code

genodeftest:
This code is no longer needed because support for the software versions it refers to has been removed.


genodeftest:
Marking as draft because I prefer to wait for #891 to be merged as it is based on that change.

sjohannes (approved):

PR #893, opened 2023-07-30
chore: fix some pytest warnings

genodeftest:
See also: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose


genodeftest:
Except for the deprecation warnings on `sunau` in `xl/metadata/wav.py`, the pytest target could now be made pedantic, i.e. fail on warnings (`py.test -x`).

sjohannes (approved):

PR #896, opened 2023-09-15
Add support for autodetecting pipewire sink devices

virtuald:
pipewire is default on fedora now, and of course it's audio sink is slightly different...

Interestingly enough, on my desktop this isn't needed, but on my laptop audio device selection doesn't work unless this patch is applied.


luzip665:
Here (Ubuntu 22.04) it works in general.

I need to install gstreamer1.0-pipewire package to see the sink.
Additionally I need to cast device id to string in https://github.com/exaile/exaile/blob/e58736e34349604d2e2867b0d37661f86cb55353/xl/player/gst/sink.py#L137 because with gstreamer1.0-pipewire package installed the device id is (int)54, without this package it is "alsa.something"

virtuald:
I had some issues with devices being restored after rebooting my laptop, so I'll have to look at the properties a bit harder to see if there's one that persists across reboots.

virtuald:
It appears that there are no stable properties in `device.props`, but `node.name` in `device.props.properties` seems to be intended to be stable (see https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/525)

virtuald:
On my upgraded Fedora 40 laptop it freezes when I am playing audio and try to switch the sink.

virtuald:
Upgraded to Fedora 41 with pipewire 1.2.7 and it also freezes.

PR #898, opened 2023-09-20
Extend multitextentrydialog with values and hints

luzip665:
With this the MultiTextEntryDialog widget is able to show text entries with preset value and an optional hint.

### DIFF
--- a/xlgui/widgets/dialogs.py
+++ b/xlgui/widgets/dialogs.py
@@ -203,7 +203,7 @@ class MultiTextEntryDialog(Gtk.Dialog):
 
         self.fields = []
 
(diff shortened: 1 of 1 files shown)


xlgui/widgets/dialogs.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, parent, title):
 
         self.fields = []
 
-    def add_field(self, label):
+    def add_field(self, label, value=None, tooltip=None):
```
virtuald, line 206: probably should document the parameters in the docstring?

luzip665: Done

PR #911, opened 2023-12-16
Shutdown plugin - Add preferences for activation by default and timeout

luzip665:
This adds a plugin preference page to set the shutdown timeout and to activate the "Close Exaile after Playback" checkbox automatically each start of Exaile.


KarlG1965:
Hi

This preference needs to be set EACH time Exaile starts. Could you please make this a preference in Exaile itsself and NOT a plugin?

PR #917, opened 2024-03-27
Don't remove tracks twice during scan.

ryneeverett:
A bug was introduced in #885 by potentially adding tracks to `removals` twice if both conditions are true. This results in a KeyError when removal of the second track is attempted because the `location` has already been deleted from `self.tracks`:

    File "./xl/trax/trackdb.py", line 369, in remove_tracks
      self._deleted_keys.append(self.tracks[location]._key)
    KeyError: 'file:///...'

### DIFF
--- a/xl/collection.py
+++ b/xl/collection.py
@@ -840,10 +840,7 @@ class Library:
                 logger.exception("Error decoding file location")
                 continue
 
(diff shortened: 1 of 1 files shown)


virtuald:
LGTM

PR #940, opened 2024-11-29
improv: added support for webm videos in files panel

nuxero:


luzip665:
Hello.

Thank you for your patch.

But, maybe, it's not that simple.

With your patch the webm-File is shown in the files panel. But it's not playable here.  It's imported with length 0.

Can you play it?

Local setup: Python 3.12, Mutagen 1.46
Tested with https://dl11.webmfiles.org/big-buck-bunny_trailer.webm and https://github.com/michaelmob/WebMCam/blob/master/Preview/example-audio.webm

luzip665:
Seems to be related with fade on user action. Without it works.

luzip665:
@nuxero Please, can you check also with fading enabled and #942 ?

nuxero:
Hey there! Thanks for reviewing this.

I have this running and working on my local fork.  You mention that this not working might be related with "fade on user action". I'm not familiar with that, so probably i don't have it enabled. How do I enable it?

My local setup:
Python 3.13 (python3)
mutagen 1.47.0 (python3-mutagen)
(both from Fedora repos)

nuxero:
I found it! Indeed, enabling "fade on user actions" did broke the ability to play webm files, but it also broke radio streams (tested http://listen.livestreamingservice.com/181-xkkountry_128k.mp3) which also has length 0 when adding to the queue, so it's likely not related to this particular change.

luzip665:
You are right.

PR #944, opened 2024-12-02
Provide mechanism to display hidden devices

virtuald:
This is mostly a response to https://github.com/exaile/exaile/pull/896... in that pipewire is awful and exaile keeps crashing when I use it. If I switch to the other GST sinks then I think it's stable.

... need others to take a look at this, but maybe I should be using a custom sink instead -- though one of the problems with that is even I would have a challenging time figuring out how to set up the custom sink, lol

### DIFF
--- a/data/ui/preferences/playback.ui
+++ b/data/ui/preferences/playback.ui
@@ -156,6 +156,22 @@
       </packing>
     </child>
     <child>
+      <object class="GtkCheckButton" id="player/show_hidden_devices">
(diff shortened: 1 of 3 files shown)


luzip665:
Here (Ubuntu 24.04) I don't see any different behaviour between current master and your patch.

virtuald:
I think if this works for you then let's merge it. On my computer if it showed all the devices it was really obnoxious and had duplicate devices, so that's why I added the checkbox.

Issue #945, opened 2024-12-02
queue tests needed


virtuald:
The various queue behaviors (remove before/after playback, never remove) have a bunch of edge cases and we should write some proper tests to ensure they all work and someone doesn't accidentally break the existing behavior.

luzip665:
@virtuald 

Would be great to have tests for the queue. I tried something in https://github.com/exaile/exaile/pull/904/commits/dd330d31ec01836bc89d0d85b46193f75ac5e553, https://github.com/exaile/exaile/pull/904/commits/5c53f294eece9878fdceb5e09ff6763bd7153359 and https://github.com/exaile/exaile/pull/904/commits/0c5c12dcd15c9e0652a6b49aaa95d0a345f379fd but finally I gave up.

PR #904, opened 2023-10-03
Playback stops at end of queue

luzip665:
Fixes #903 

All tests done with a playlist and the queue
Scene 1
* Without preferences "Remove upon playback" and "Remove after playback"
* Start in playlist
* It switches to queue and plays first track of it
* After ending the queue it switches back to next track in playlist
This one is not working right now

Scene 2
* With preference "Remove upon playback" and without "Remove after playback"
* Start in playlist
* It switches to queue and immediately plays and removes the first track 
* After ending the queue it switches back to next track in playlist

Scene 3
* With preferences "Remove upon playback" and "Remove after playback"
* Start in playlist
* It switches to queue and plays the first track 
* After finishing the first track it is removed from queue
* After ending the queue it switches back to next track in playlist

Are the more scenarios to consider?

PR #947, opened 2024-12-02
Fix and clarify queue behaviors

### LINKED ISSUE #946: Queue repeats first two tracks over and over again
If `remove_item_when_played` is False and `remove_item_after_played` is True (or maybe if both are false, not sure, don't care), the queue repeats the first two tracks instead of behaving like a normal playlist.

virtuald:
@luzip665 I think this retains your desired queue behaviors and also retains mine as well. In particular, if there is no removal from the queue during playback, I expect the queue to behave like a normal playlist

- Clearly separate remove before/after playback flags
- If not removing from queue, restore normal playlist behavior

I should probably write some queue tests (#945).

Fixes #946

### DIFF
--- a/data/ui/preferences/playback.ui
+++ b/data/ui/preferences/playback.ui
@@ -218,7 +218,7 @@
     </child>
     <child>
       <object class="GtkCheckButton" id="queue/remove_item_after_played">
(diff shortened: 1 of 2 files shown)


luzip665:
@virtuald Thank you for fixing. It was broken in #904

PR #956, opened 2025-01-23
Fix race condition for exaile.desktop build

bmwiedemann:
Without this patch, `make -j4` would create the `exaile.desktop` file at the very beginning of the build and complain about missing `po/LINGUAS` input. That means it missed translations when compared to the -j1 build.

This patch was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,7 +301,7 @@ format:
 check_format:
 	$(BLACK) --check --diff -S *.py plugins/ xl/ xlgui/ tests/ tools/installer/exaile.spec
 
-desktop_files: builddir
+desktop_files: builddir pot
 	msgfmt --desktop --template=data/exaile.desktop.in -d po -o build/exaile.desktop
 	msgfmt --xml --template=data/exaile.appdata.xml.in -d po -o build/exaile.appdata.xml
 


sjohannes:
If the only thing it needs is the `po/LINGUAS` file then `make pot` is probably overkill. I think it should be something along the lines of this:

```make
.PHONY: ... linguas ...

linguas:
	echo $(LINGUAS) > po/LINGUAS

pot:
	...
	# Remove the last line that creates po/LINGUAS

desktop_files: linguas
	...

# Remove builddir (unrelated, but it's not needed)
# Add linguas (not needed because it's a dependency of desktop_files, but just for clarity)
locale: $(LOCALE_OBJS) desktop_files linguas
```

bmwiedemann:
I had tried a patch like that, but it made the build fail for some reason.

```diff
--- a/Makefile
+++ b/Makefile
@@ -240,11 +240,13 @@ clean:
        if [ -f po/LINGUAS ]; then rm po/LINGUAS; fi
 
 po/messages.pot: pot
+po/LINGUAS:
+       echo $(LINGUAS) > po/LINGUAS
 
 # The "set -o pipefail" makes the whole thing die if any of the find fails.
 #   dash (Debian's /bin/sh) doesn't support it and exits immediately, so we test it in a subshell.
 # The "export LC_ALL=C" disables any locale-dependent sort behavior.
-pot:
+pot: po/LINGUAS
        ( ( set -o pipefail 2> /dev/null ) && set -o pipefail ; \
          export LC_ALL=C && cd po && \
          { find ../xl ../xlgui -name "*.py" | sort && \
@@ -258,7 +260,6 @@ pot:
          xgettext -j -o messages.pot '../data/exaile.appdata.xml.in' )
        find po -name '*.po' -exec \
          msgmerge --previous --no-fuzzy-matching --update {} po/messages.pot \;
-       echo $(LINGUAS) > po/LINGUAS
 
 potball: builddir
        tar --bzip2 --format=posix -cf build/exaile-po.tar.bz2 po/ \
@@ -301,7 +302,7 @@ format:
 check_format:
        $(BLACK) --check --diff -S *.py plugins/ xl/ xlgui/ tests/ tools/installer/exaile.spec
 
-desktop_files: builddir
+desktop_files: builddir po/LINGUAS
        msgfmt --desktop --template=data/exaile.desktop.in -d po -o build/exaile.desktop
        msgfmt --xml --template=data/exaile.appdata.xml.in -d po -o build/exaile.appdata.xml
```

bmwiedemann:
I think, it needs more than just `LINGUAS`. I got a variant to build and still had this variation:
```diff
--- old//usr/share/metainfo/exaile.appdata.xml  
+++ new//usr/share/metainfo/exaile.appdata.xml 
@@ -9,7 +9,6 @@
   <name xml:lang="zh">Exaile</name>
   <name xml:lang="sv">Exaile</name>
   <name xml:lang="sk">Exaile</name>
-  <name xml:lang="ru">Exaile</name>
   <name xml:lang="pt_BR">Exaile</name>
   <name xml:lang="pt">Exaile</name>
   <name xml:lang="pl">Exaile</name>
...
```

sjohannes:
@bmwiedemann Could you try <https://github.com/exaile/exaile/pull/957>? I think the LINGUAS file is only required by the .desktop and appdata build steps, so what I've done in that PR is to always create the file just before it's needed (and remove it afterwards).

bmwiedemann:
Closed in favor of #957

PR #957, opened 2025-01-24
Fix .desktop & appdata generation

sjohannes:
Previously this depended on the po/LINGUAS file created by the `pot` target, potentially causing error on `make -j`.

This commit also changes the `locale` target to not depend on `desktop_files` because `locale` is only meant to build the .mo files.

As a bonus, this also results in the .desktop and appdata files now not being pointlessly created in the Windows CI build. We make the installation of these files optional in the `make install` step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,10 +181,8 @@ install-target: make-install-dirs
 	install -p -m 644 data/ui/preferences/*.ui $(EXAILESHAREDIR)/data/ui/preferences
 	install -p -m 644 data/ui/preferences/widgets/*.ui $(EXAILESHAREDIR)/data/ui/preferences/widgets
 	install -p -m 644 data/ui/widgets/*.ui $(EXAILESHAREDIR)/data/ui/widgets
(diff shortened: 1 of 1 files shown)


bmwiedemann:
This was deterministic in my tests.
However, I am a bit worried why you delete `po/LINGUAS` there - if anything else in the build reacts to its content, there is a chance, it could be seen in a parallel build. Why not move that into `make clean` ?

sjohannes:
The deletion is to emphasise that `po/LINGUAS` is only a temporary file meant as input for those two `msgfmt` calls. If anything else tries to read the file then that's a bug and I want it to fail (at least most of the time — you're right that it's not a reliable detection method because the file does exist for a short period, but that's okay).

...I'm not sure I explained that well; hopefully it makes sense.

bmwiedemann:
OK. Then it is probably fine.

sjohannes:
Thanks for your initial PR and for testing this one.

PR #965, opened 2025-03-30
readme: Fix homepage link

vdawg-git:
`https://www.exaile.org/` is not reachable.
But `https://exaile.org/` is


luzip665:
Hi @vdawg-git 

thank you for the merge request.

But I think we better check the website.

Please see https://github.com/exaile/exaile/discussions/961

PR #970, opened 2025-06-14
playlist export

### LINKED ISSUE #932: Exporting playlists as M3U using an "absolute path" doesn't work

1. Export a playlist using absolute paths (not relative)
2. Open it in a text editor
3. See that the files aren't listed

### Expected Behavior

The text file should show files in it (M3U)

### Current Behavior

It instead shows this.
```
#EXTM3U
#PLAYLIST: Playlist 1
```

### Environment

* Operating System and version:

* Exaile Version: 

- Windows 11 22H2
- Exaile version 4.1.3

luzip665:
* use paths instead of uris for local files
* fixes #932

### DIFF
--- a/xl/playlist.py
+++ b/xl/playlist.py
@@ -287,58 +287,63 @@ class FormatConverter:
         return track_uri
 
     def get_track_export_path(
-        self, playlist_path: str, track_path: str, options: PlaylistExportOptions
+        self, playlist_uri: str, track_uri: str, options: PlaylistExportOptions
     ):
         """
         Retrieves the export path of a track,
         possibly influenced by options
 
-        :param playlist_path: the export path of the playlist
-        :param track_path: the path of the track
+        :param playlist_uri: the export path of the playlist
+        :param track_uri: the path of the track
         :param options: options
         """
 
-        track_path_components = urllib.parse.urlparse(track_path)
-        export_path_components = urllib.parse.urlparse(playlist_path)
+        track_uri_components = urllib.parse.urlparse(track_uri)
+        export_uri_components = urllib.parse.urlparse(playlist_uri)
 
         if (
-            export_path_components.scheme != track_path_components.scheme
-            or export_path_components.netloc != track_path_components.netloc
+            export_uri_components.scheme != track_uri_components.scheme
+            or export_uri_components.netloc != track_uri_components.netloc
         ):
             # save files to playlist on different location
             # return track path as is
-            if track_path_components.scheme == 'file':
+            if track_uri_components.scheme == 'file':
                 # as path if local file
-                return Gio.File.new_for_uri(track_path).get_path()
-            return track_path
+                gio = Gio.File.new_for_uri(track_uri)
+                path = gio.get_path()
+                return Gio.File.new_for_uri(track_uri).get_path()
+            return track_uri
 
         if options is None or not options.relative:
             # return absolute uri
-            if track_path_components.scheme == 'file':
+            if track_uri_components.scheme == 'file':
                 # as path if local file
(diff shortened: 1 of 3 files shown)


xl/playlist.py
```diff
@@ -297,34 +297,25 @@ def get_track_export_path(
...
-                ):
-                    # Gio.File.get_relative_path does not generate relative paths
-                    # for tracks located above the playlist in the path hierarchy,
-                    # thus process both paths as done here
-                    track_path = os.path.relpath(
-                        track_path_components.path, export_path_components.path
-                    )
+        track_path_components = urllib.parse.urlparse(track_path)
+
+        if not track_path_components.scheme == 'file':
```
rokm, line 303: `if track_path_components.scheme != 'file':` would be a more canonical I think.

xl/playlist.py
```diff
@@ -297,34 +297,25 @@ def get_track_export_path(
...
+        track_path = Gio.File.new_for_uri(track_path)
+        if options is None or not options.relative:
+            # return local files as is but as path instead of uri
+            return track_path.get_path()
+
+        # calculate relative path
+        playlist_file = Gio.File.new_for_uri(playlist_path)
+        export_path = playlist_file.get_parent().get_path()
+
+        track_path = os.path.relpath(track_path.get_path(), export_path)
```
rokm, line 316: Hmm, what should happen if a relative path cannot be computed?

For example, on Windows, music files might be on one drive, and playlist is being stored on another drive. In such case, `os.path.relpath()` will raise an exception, and the file will end up being skipped again. Should such files be returned with absolute path instead?

The changes here remove support for using relative paths with non-local schema in case when both media files and the playlist use the same schema and location (for example, remote gvfs mounts under linux). I have no idea if this worked as it should have and if it is being used, though.

luzip665: I cannot confirm on Win11 with Exaile built in https://ci.appveyor.com/project/ExaileDevelopmentTeam/exaile/builds/52247370

With relative and with absolut paths I get `\` as separator.

@desbest Can you check also?

rokm: Hmm, you're right. I guess this is because I was running from source in the msys environment.

The msys-installed python's `ntpath.py` has this block:

```python
if sys.platform == "win32" and os.environ.get("MSYSTEM", ""):
    sep = '/'
    altsep = '\\'
else:
    sep = '\\'
    altsep = '/'
```

So it ends up with `/` as separator when ran from source using msys python in msys environment, and `\` when ran as PyInstaller-frozen, because `MSYSTEM` is not set.

xl/playlist.py
```diff
@@ -287,44 +287,63 @@ def _iter_uris(pp, tp):
...
-                ):
-                    # Gio.File.get_relative_path does not generate relative paths
-                    # for tracks located above the playlist in the path hierarchy,
-                    # thus process both paths as done here
-                    track_path = os.path.relpath(
-                        track_path_components.path, export_path_components.path
-                    )
+                track_uri = os.path.relpath(
+                    track_uri.get_path(), playlist_file_folder.get_path()
+                )
```
rokm, line 334: The separators are still inconsistent in relative-path variant when running on Windows from source. 

It is not really an issue for most end-users, but it might be confusing when developing/debugging on Windows to see different behavior between source version and PyInstaller-frozen build.

Adding a
```python
                # Ensure '\' is always used as separator on Windows.
                if sys.platform == 'win32' and os.sep == '/':
                    track_uri = track_uri.replace(os.sep, '\\')
```
here should do the trick (requires an `import sys` at the top, though).

rokm (approved):

PR #974, opened 2025-07-16
xl/plugins: fix plugin enumeration with PyGObject >= 3.52.0

rokm:
The old GIRepository API (exposed by PyGObject < 3.52.0) has a `Repository.get_default()` method to obtain global singleton object; this method was removed in the new API (exposed by PyGObject >= 3.52.0), which requires creation of separate `Repository` instances.

Properly handle both APIs, in order to fix plugin enumeration when PyGObject >= 3.52.0 is used; otherwise, trying to list plugins in Preferences dialog fails with "Could not load plugin info! Failed plugins: <list of all plugins>" message, which is caused by an `AttributeError` in `PluginsManager.is_potentially_broken()`.

### DIFF
--- a/xl/plugins.py
+++ b/xl/plugins.py
@@ -264,11 +264,19 @@ class PluginsManager:
         # https://github.com/exaile/exaile/issues/962
         try:
             gi.require_version('GIRepository', '3.0')
+            new_api = True
         except ValueError:
             gi.require_version('GIRepository', '2.0')
+            new_api = False
         from gi.repository import GIRepository
 
-        gir = GIRepository.Repository.get_default()
+        # The old API had `get_default` method to obtain global singleton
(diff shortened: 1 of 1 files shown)


rokm:
This is a follow-up to https://github.com/exaile/exaile/pull/963, which did not completely sort out the compatibility issues with PyGObject >= 3.52.0; while the application starts, none of the plugins are available.

sjohannes:
Sorry, yeah, I've just run into this as well. I don't really understand why `get_default` is removed; what's even more confusing is that the current GLib unstable version added [`dup_default`](https://docs.gtk.org/girepository/type_func.Repository.dup_default.html), which may or may not be the same thing.

sjohannes (approved):

Issue #344, opened 2017-04-28
cover blinks when tags are changed


virtuald:
This happens on OSX when adding tags with grouptagger.

genodeftest:
I also see this on Linux when just pressing the "next" button. This issue is more visible when using the newly updated `osd` plugin which just uses a simple CoverWidget.

sbrubes:
Too many calls to _TrackInfoPane.set_track_, specially at [_on_track_tags_changed_](https://github.com/exaile/exaile/blob/master/xlgui/widgets/info.py#L311).

Checking if tag is cover will minimize it.

PR #1009, opened 2026-04-07
fix: call to on_cover_chosen method

grahamwhiteuk:
Fix an issue around updating the cover image when tags are changed.

Possible fix to: #344


sjohannes:
Thanks. `on_cover_chosen` was modified in 4cfcd7cd4080da59ea83e4eb6c7c3abce3c1420a but this was missed.

I don't know what calls this `on_track_tags_changed` function though. Editing tags doesn't seem to call it at all….

PR #1016, opened 2026-04-23
Inhibitsuspend: systemd-inhibit, elogind-inhibit, and logind dbus support and fix xfce/gnome

ryneeverett:
I don't know if there are actually any systems which support systemd-inhibit but not the logind bus. I also don't understand the ecosystem enough to know why all these desktop environments have their own dbus interfaces to handle suspend inhibitors and wonder if the logind bus would also work for them. I'm happy to rework or break this into separate PR's if somebody knows better.

### DIFF
--- a/plugins/inhibitsuspend/__init__.py
+++ b/plugins/inhibitsuspend/__init__.py
@@ -68,6 +68,13 @@ class SuspendInhibit:
         # see https://askubuntu.com/questions/72549/how-to-determine-which-window-manager-is-running
         xdg_session = os.getenv('XDG_CURRENT_DESKTOP', '').lower()
 
+        # Attempt to get logind bus
+        try:
+            bus = dbus.SystemBus()
+            logind = bus.get_object('org.freedesktop.login1', '/org/freedesktop/login1')
+        except dbus.exceptions.DBusException:
+            logind = None
+
         # Attempt to find an adaptor that works
         if 'gnome' in session or 'gnome' in xdg_session:
             self.adapter = GnomeAdapter()
@@ -79,6 +86,8 @@ class SuspendInhibit:
                 self.adapter = KdeAdapter()
         elif 'xfce' in session or 'xfce' in xdg_session:
             self.adapter = XfceAdapter()
+        elif logind:
+            self.adapter = LogindAdapter(logind)
         elif shutil.which(SystemdAdapter.cmd) is not None:
             self.adapter = SystemdAdapter()
         elif shutil.which(ElogindAdapter.cmd) is not None:
@@ -95,7 +104,7 @@ class SuspendInhibit:
                 self.adapter = GnomeAdapter()
         else:
             raise NotImplementedError(xdg_session)
-        logger.debug(f"Inhibiting suspend with {self.adapter.__qualname__}")
+        logger.debug(f"Inhibiting suspend with {self.adapter.__class__.__name__}")
 
     def destroy(self):
         self.adapter.destroy()
@@ -193,6 +202,31 @@ class SuspendAdapter(adapters.PlaybackAdapter):
         Must not block
         """
         raise NotImplementedError('Method not Overridden')
+
+
+class LogindAdapter(SuspendAdapter):
+    """
+    Adapter for logind interface.
+
+    See https://systemd.io/INHIBITOR_LOCKS/
+    """
+
+    def __init__(self, logind_obj):
+        interface = 'org.freedesktop.login1.Manager'
+        self.iface = dbus.Interface(logind_obj, interface)
+        super().__init__()
+
+    def _inhibit_call(self):
+        lock = self.iface.Inhibit(
+            'shutdown:sleep:idle',  # what
+            'Exaile',  # who
+            'Playing Music',  # why
+            'block',  # mode
+        )
+        self.lock_file_descriptor = lock.take()
+
+    def _uninhibit_call(self):
+        os.close(self.lock_file_descriptor)
 
 
 class DbusSuspendAdapter(SuspendAdapter):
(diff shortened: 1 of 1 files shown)


Saijin-Naib:
There is also the xdg desktop portal inhibit suspend method, which should in theory, be DE-agnostic (ish):  
https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Inhibit.html

ryneeverett:
Nice. I don't have that portal on my sway systems though so I'm not inclined to implement without testing. (I didn't really set out to support everything in the world here. I started with the lazy systemd-inhibit solution suggested by the TODO comment and then pursued the more elegant dbus solution.)

ryneeverett:
I did some testing on my Guix system, which is "interesting" because it does not use systemd, it does have elogind, and I run Gnome on it. (This is basically a default Guix desktop installation.)

On 4.2.1, the plugin crashes every time I play music. I didn't dig into it but I presume that the Gnome adapter is what is crashing and is doing so because the dbus interface doesn't actually exist.

Running with this PR, the plugin does not crash -- surprisingly with the Logind adapter -- and the inhibitor shows up in `elogind-inhibit --list`. However, the system still goes to sleep which doesn't surprise me because I haven't found `elogind-inhibit` to actually work on my system.

Saijin-Naib:
```
  File "/usr/lib/exaile/xl/plugins.py", line 169, in enable_plugin
    plugin.enable(self.exaile)
    ~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 33, in enable
    SUSPEND_PLUGIN = SuspendInhibit()
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 88, in __init__
    self.adapter = XfceAdapter()
                   ~~~~~~~~~~~^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 341, in __init__
    super().__init__()
    ~~~~~~~~~~~~~~~~^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 279, in __init__
    super().__init__(bus_name, object_name, interface_name)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 242, in __init__
    super().__init__()
    ~~~~~~~~~~~~~~~~^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 130, in __init__
    super().__init__(PLAYER)
    ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib/exaile/xl/player/adapters.py", line 49, in __init__
    self.on_playback_track_start('playback_track_start', player, player.current)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 181, in on_playback_track_start
    self.inhibit()
    ~~~~~~~~~~~~^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 148, in inhibit
    self._inhibit_call()
    ~~~~~~~~~~~~~~~~~~^^
  File "/usr/share/exaile/plugins/inhibitsuspend/__init__.py", line 197, in _inhibit_call
    raise NotImplementedError('Method not Overridden')
NotImplementedError: Method not Overridden
```

ryneeverett:
@Saijin-Naib Are you sure you can't reproduce without this patch? I don't see as I've touched anything that could cause this. It seems like the dbus adapters for xfce and gnome implement `_dbus` methods which are never called anywhere and fail to implement the required `_inhibit` methods.

Saijin-Naib:
> @Saijin-Naib Are you sure you can't reproduce without this patch? I don't see as I've touched anything that could cause this. It seems like the dbus adapters for xfce and gnome implement `_dbus` methods which are never called anywhere and fail to implement the required `_inhibit` methods.

I am not entirely sure.

What is the proper method to test? Roll back this patch and try the inhibit suspend plugin again and collect the error trace (if any)?

So, GNOME and XFCE can't be supported by this plugin without changes in the DEs themselves?

ryneeverett:
> What is the proper method to test? Roll back this patch and try the inhibit suspend plugin again and collect the error trace (if any)?

Yes, I suspect you'd get the same result with just different line numbers.

> So, GNOME and XFCE can't be supported by this plugin without changes in the DEs themselves?

No, it looks like a preexisting bug in this plugin that nobody has noticed before. I'd guess the method names just need to be corrected.

Saijin-Naib:
Ah, aweosme thanks. Fixing that is likely out of scope for this PR right?

ryneeverett:
Eh, a little scope creep wont hurt. I pushed a commit addressing it but did not test it out because i don't use these desktop environments. Please let us know if it works.

Saijin-Naib:
XFCE seems to pass with flying colors now:  
```
INFO    : Suspend Bus Acquired
INFO    : Suspend Inhibitor Enabled
```
The PowerManager applet for XFCE panel shows "exaile is currently inhibiting power management". Whether or not it does in practice I will test as well later.

It will take me a bit to get GNOME setup on here to test, but I'll report in when I do.

Thanks for the fix!

Saijin-Naib:
GNOME 51 seems to be happy with it as well:  
```
INFO    : Connected with notify server gnome-shell (version 51.0) by GNOME
INFO    : Supported spec version: 1.2
INFO    : Suspend Bus Acquired
INFO    : Suspend Inhibitor Enabled
```

Saijin-Naib:
> Whether or not it does in practice I will test as well later.

Yes, it properly inhibits all power management behavior in XFCE.

Seriously awesome stuff! Thank you!

Issue #922, opened 2024-05-09
Insert method in class MetadataList results in an extra None Value in its metadata list


Ozy-Viking:
## Insert method in class MetadataList results in an extra None Value in its metadata list

The `insert` method of MetadataList Class inserts an extra `None` value into its metadatalists, when `e == i`. This results in unequal length of metadata and list. Note, when the insert occurs at the end of the iterable, no extra `None` is inserted. 

Bug found while writing 1,000 test cases for a Masters course I'm undertaking, let me know if you want all the tests. :) 

### Steps to Reproduce (for bugs)

1. Use the insert method of class MetadataList with a non-zero length MetadataList and insert the item not with the length of the iterable.

### Expected Behavior

- Only add the metadata for the item being inserted and both lists (`metadata` and `__list`) maintain equal length.

### Current Behavior

- File location is in the root of the project, name doesn't matter for testing purposes.

#### Initialization

This section is setting up testing that shows the bug.

```python
# test_commom.py
from xl.common import MetadataList

def mock_metadata(title, **kwargs):
    return {"title": title } | kwargs

class MockSong:
    def __init__(self, title, **kwargs) -> None:
        self.__dict__ = mock_metadata(title, **kwargs)

    @property
    def metadata(self):
        return self.__dict__
        
    def __eq__(self, other):
        if isinstance(other, MockSong):
            return self.metadata == other.metadata
        else:
            return NotImplemented
    
    def __str__(self) -> str:
        return f'{self.title} by {self.artist}'
    
    def __repr__(self) -> str:
        return f'MockSong("{self.title}")'

def mock_songs_with_metadata(num_songs=10):
    song_list = [MockSong(f"Song {x}") for x in range(num_songs)]
    meta_list = [x.metadata for x in song_list]
    return (song_list, meta_list)

def mock_song():
    return MockSong("Insert Song")

NUM_OF_SONGS = 5
song_list, meta_list = mock_songs_with_metadata(NUM_OF_SONGS)
mdl = MetadataList(song_list, meta_list)

def reset_mdl():
    song_list, meta_list = mock_songs_with_metadata(NUM_OF_SONGS)
    global mdl
    mdl = MetadataList(song_list, meta_list)  
```

#### Insert at Beginning

When the index is 0 (`mdl.insert(0, insert_song, insert_song.metadata)`), the None Value is at index 1. The inserted object and metadata is in the correct location but the metadata after index 0 is 1 position out.

```python
metadata = [
[...]

PR #923, opened 2024-05-09
 #922 Insert method in class MetadataList results in an extra None Value in its metadata list

Ozy-Viking:
#922

PR #1026, opened 2026-08-02
Fix MetadataList insert metadata alignment

virtuald:
Note: AI generated, human reviewed

* Fixes issue #922 (alternative to https://github.com/exaile/exaile/pull/923)

The original implementation inserted placeholder metadata through `__setitem__`, then accidentally inserted the supplied metadata a second time. The suggested fix fails for negative indices because the list length changes before self.metadata[i] is evaluated

### DIFF
--- a/xl/common.py
+++ b/xl/common.py
@@ -701,13 +701,7 @@ class MetadataList:
         self[len(self) : len(self)] = other
 
     def insert(self, i, item, metadata=None):
-        if i >= len(self):
-            i = len(self)
-            e = len(self) + 1
-        else:
-            e = i
-        self[i:e] = [item]
-        self.metadata[i:e] = [metadata]
+        self[i:i] = MetadataList([item], [metadata])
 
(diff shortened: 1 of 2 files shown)


virtuald:
I'm not really confident this is the Right Fix. It's weird that this hasn't exhibited itself elsewhere.

xl/common.py
```diff
@@ -701,13 +701,7 @@ def extend(self, other):
...
 
     def insert(self, i, item, metadata=None):
-        if i >= len(self):
-            i = len(self)
-            e = len(self) + 1
-        else:
-            e = i
-        self[i:e] = [item]
-        self.metadata[i:e] = [metadata]
+        self[i:i] = MetadataList([item], [metadata])
```
sjohannes, line 704: I think this is equivalent and easier to understand for my brain:

```py
self.__list.insert(i, item)
self.metadata.insert(i, metadata)
```

but it's fine either way.

